PR #89, opened 2022-02-24
Add intellij file settings instructions on Linux

wipfli:
On my ubuntu machine, the path to the settings is `file -> settings` and not `preferences`. This pull request adds a note about the path to the settings in linux.﻿


msbarry:
Thanks @wipfli ! Congrats on being the first non-me, non-@dependabot contributor! 😄

PR #94, opened 2022-02-25
Update adoptium links

wipfli:
Expands adoptium java install links to directly point to win, mac, and linux installers.﻿

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 Pull requests are welcome! To set up your development environment:
 
 - Fork the repo
-- [Install Java 16 or later](https://adoptium.net/installation.html)
+- Install Java 16 or later. You can download Java manually from [Adoptium](https://adoptium.net/installation.html) or use their installers for [Windows](https://adoptium.net/installation.html#windows-msi), [macOS](https://adoptium.net/installation.html#macos-pkg), and [Linux](https://github.com/adoptium/website-v2/blob/main/src/asciidoc-pages/installation/linux.adoc).
 - Build and run the tests ([mvnw](https://github.com/takari/maven-wrapper) automatically downloads maven the first time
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 Pull requests are welcome! To set up your development environment:
 
 - Fork the repo
-- [Install Java 16 or later](https://adoptium.net/installation.html)
+- Install Java 16 or later. You can download Java manually from [Adoptium](https://adoptium.net/installation.html) or use their installers for [Windows](https://adoptium.net/installation.html#windows-msi), [macOS](https://adoptium.net/installation.html#macos-pkg), and [Linux](https://github.com/adoptium/website-v2/blob/main/src/asciidoc-pages/installation/linux.adoc).
```
msbarry, line 6: This install url is also referenced from planetiler-examples/README.md and README.md https://github.com/onthegomap/planetiler/search?q=https%3A%2F%2Fadoptium.net%2Finstallation.html - think we should change it there as well?

wipfli: Good catch. I suggest cross-referencing the CONTRIBUTING.md file, see commit

msbarry: Thanks! that looks good. What do you think about inlining the command for common installers?
- `brew install --cask temurin`
- `apt-get install openjdk-17-jdk` (probably better to prefer the recommended distro's openjdk build)

wipfli: Sounds like a good idea

msbarry: Cool, do you want to add those or want me to?

msbarry: Looks like I can update your branch - how does this look?

wipfli (commented):
Looks good to me!

PR #109, opened 2022-03-04
Add newlines to OpenMapTiles table

wipfli:
Adds newlines to the OpenMapTiles table file. My motivation for this is to simplify customization of the basemap profile. For example, with the hot-air balloon I often land in places where it matters if the `highway=track` has `tracktype=grade1` or `tracktype=grade5`. The `tracktype` tag is not part of OMT, so I added it manually to the basemap profile. It is super fun to customize the tiles like this. After this pull request, adding new tags will only insert a new line and result in a simpler git diff. Merging upstream changes will also be easier.


wipfli:
This pull request should only insert new lines, i.e., it should only have whitespace changes.

wipfli:
To checkt that this pull request indeed only edited whitespaces, I did the following:

```
git checkout main
cat planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/generated/Tables.java | tr -d "\t\r\n[:space:]" > old.txt
git checkout newlines-omt-table
cat planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/generated/Tables.java | tr -d "\t\r\n[:space:]" > new.txt
diff old.txt new.txt
```

msbarry:
Thanks! All of the code in `planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/generated` is generated automatically by `scripts/regenerate-openmaptiles.sh` which invokes `planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/Generate.java`.  It downloads the yaml files that describe the schema from openmaptiles github repo and generates these java files, then prints instructions for how to autoformat in intellij.

For now I want to make sure that regenerating and auto-formatting generates a consistent output, so can you update the code in Generate.java to insert the newlines in the generated output? That would make it easier for people to fork/customize, and see diffs from importing new openmaptiles schema versions, but I don't think we'd want the version on main to diverge from openmaptiles.

In the medium term, we're going to move the openmaptiles profile to the openmaptiles repo, and use regular yaml files to control the tag selection.

wipfli:
Thanks. I was a bit suspicious of the generated folder :S In MapLibre GL JS we removed generated files from source control because (a) they are not required and (b) cause sometimes confusion with new community members... Should I submit a pull request to remove generated files from source control?

wipfli:
Oh and I thought, uh poor Mike had to copy all those tags by hand from OMT...

msbarry:
What were the generated files you removed in maplibre? Did you add a build step to generate them for new users?

Removing them (and generated protobuf files) would make sense, but we'd need to replace that with a maven build step that generates them instead. I haven't looked into that but if you were able to that would be great!

The generated code contains:

1. allowed layers, fields, and values with javadoc containing their descriptions so you can get nice tooltips in the IDE when working with the data
2. intermediate records (in that `Tables.java` file) that OSM elements get parsed into so you get autocomplete and compile warnings on the 
3. matching expressions that can filter OSM elements

I think we want to keep 1 and 2, but 3 will eventually go away and be replaced by dynamically parsing the yaml at runtime.

wipfli:
> What were the generated files you removed in maplibre? Did you add a build step to generate them for new users?

The style spec is a json document from which we generate `.ts`. We removed the generated files from git and renamed them to `.g.ts` to highlight that they are generated. The script to generate these files is run by `npm postinstall`.

wipfli:
#111 and #110

wipfli:
Closing this for now, since I should better not edit a generated file.

PR #110, opened 2022-03-04
Add intellij format workflow

wipfli:


msbarry:
I think it would be good to get automate formatting and linting... just wondering if it might be better to use a tool for both of those that doesn't force people to use intellij?  Ideally you could run the formatter/linter from the command line, github action, or use it in a plugin in your editor of choice. Intellij's formatter seems to be tightly coupled to the full intellij installation.

[spotless](https://github.com/diffplug/spotless) seems to be the most popular formatter framework for java/maven. It supports a few formatter implementations, not sure which one would be the best yet:

- [google java formatter](https://github.com/google/google-java-format) which also an [intellij plugin](https://plugins.jetbrains.com/plugin/8527-google-java-format/) - I think people like it, but it enforces google's coding style and is not really configurable
- [palantir java formatter](https://github.com/palantir/palantir-java-format) which is a fork of google's (not sure what they changed?) and also has an [intellij plugin](https://plugins.jetbrains.com/plugin/13180-palantir-java-format)
- eclipse jdt formatter - probably the most configurable, but doesn't have best practices out of the box, there's also an [intellij plugin](https://plugins.jetbrains.com/plugin/6546-adapter-for-eclipse-code-formatter)
- looks like [prettier](https://github.com/jhipster/prettier-java) also works with java ... that's interesting but would probably rather have one running inside the JVM to avoid issues installing node.

Any thoughts here @wipfli ?

msbarry:
Also on the github action, was the idea that it would format code on push, and update their fork? Or just fail if they haven't formatter?

wipfli:
Thanks for sharing other options. I was hoping that you would do this. I am happy to have a closer look at these. Regarding CI formatting, I think it should just fail...

wipfli:
Java friend @zelonewolf what is your preferred formatter?

wipfli:
By the way, what is the difference between java and maven?

ZeLonewolf:
@wipfli my typical practice is to use the Eclipse formatter, customized "the way I like it" so it's sane with Java 8 streams and then commit the settings file to the repository.  But there are probably more modern approaches today.

ZeLonewolf:
Maven is a dependency management system.  Gradle uses maven behind the scenes also.  Basically, you can declare your dependencies and maven pulls down the transitive dependencies.  It's similar to what npm does.

wipfli:
#117 shows what google-java-format would look like.

msbarry:
Dug into this a bit more...

Google java format intellij plugin works until you run it on a file that uses newer java features, then it fails silently... there are some workarounds but the setup is too complicated to expect any newcomer to do.

Eclipse java formatter plugin seems to work OK in intellij, and from the command-line with spotless. The current [intellij formatter settings](https://github.com/onthegomap/planetiler/blob/main/.idea/codeStyles/Project.xml) are customized from [Google's style guide](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) - they have matching [eclipse configuration](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml) so it wouldn't be too hard to get to identical config for that.

There's also a spotless issue for supporting the intellij formatter https://github.com/diffplug/spotless/issues/200 - but I don't see any path toward a resolution until intellij makes their formatter available outside the IDE installation itself.

So I think our options here are:

1. easiest: just continue requiring people to use intellij, no changes from what we've got now
2. probably best: switch to eclipse java formatter (intellij instructions for importing, local script to run `mvn spotless:apply`, and github action for `mvn spotless:check`)
3. craziest: like keep using intellij, and if you use a different IDE provide a script that runs intellij in docker to format the code?

What do you think?

ZeLonewolf:
> probably best: switch to eclipse java formatter (intellij instructions for importing, local script to run mvn spotless:apply, and github action for mvn spotless:check)

This seems reasonable to me.  The goal here is to avoid the creation of unnecessary diffs when contributing PRs, and this accomplishes it with what sounds like reasonably minimal pain.

msbarry:
Another plus for eclipse formatter config  (and checkstyle for linting) is that it works out of the box for vscode as well: https://code.visualstudio.com/docs/java/java-linting#_applying-formatter-settings

msbarry:
@wipfli I can help set this up, will aim for a diff that migrates to eclipse formatter and adds instructions for vscode, IntelliJ, and eclipse and a GitHub action to check.

wipfli:
Beautiful, thanks @msbarry. I use vscode.

wipfli:
And thanks for helping out with this @msbarry. I played around with maven and gradle and spotless this morning but was completely lost...

msbarry:
Took a stab at migrating to eclipse formatter to support vscode, intellij, and eclipse here: #122

PR #122, opened 2022-03-08
Migrate to eclipse formatter to support multiple IDEs

msbarry:
Migrate to eclipse formatter which seems to be the standard that is supported by many IDEs and add insructions for VSCode, IntelliJ and Eclipse.

- [x] Migrate from IntelliJ IDEA formatter to Eclipse formatter, started from https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml and tweaked until it produced the smallest diff to what's currently on main
- [x] Formatted all of the code according to those settings
- [x] Allow running formatter from command line through [spotless](https://github.com/diffplug/spotless) `mvn spotless:apply` / `mvn spotless:check` (bound to `mvn verify`)
- [x] Add instructions for using the [Adapter for Eclipse Code Formatter for IntelliJ](https://plugins.jetbrains.com/plugin/6546-adapter-for-eclipse-code-formatter)
- [x] Add instructions for VSCode and `.vscode/settings.json` with defaults
- [x] Add instructions for Eclipse
- [ ] Switch to ratchetFrom

See prior discussion on #110


planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/layers/Transportation.java
```diff
@@ -434,7 +435,7 @@ int getMinzoom(Tables.OsmHighwayLinestring element, String highwayClass) {
   private boolean isPierPolygon(Tables.OsmHighwayLinestring element) {
     if ("pier".equals(element.manMade())) {
       try {
-        if (element.source().worldGeometry() instanceof LineString lineString && lineString.isClosed()) {
+        if (element.source().worldGeometry()instanceof LineString lineString && lineString.isClosed()) {
```
msbarry, line 438: There's a bug in the current version of the eclipse formatter that removes the space before `instanceof` - should be fixed when they upgrade to 4.22 https://github.com/diffplug/spotless/issues/1059

msbarry:
@wipfli here's a stab at migrating to the eclipse formatter, want to try out the [VSCode instructions](https://github.com/onthegomap/planetiler/blob/eclipse-formatter/CONTRIBUTING.md#visual-studio-code)

msbarry:
Also I can add `<ratchetFrom>origin/main</ratchetFrom>` to the config for spotless which limits it to only format (or check formatting of) files that have been changed from main.  Do we think that's worth it to add? Or if we're starting off with a completely formatted repo then maybe it's not necessary.

wipfli:
I can run the tests from vscode. Cool!

wipfli:
I can check formatting from cli with `./mvnw spotless:check`

scripts/regenerate-openmaptiles.sh
```diff
@@ -12,3 +12,6 @@ echo "Building..."
 
 echo "Running..."
 java -cp planetiler-dist/target/*-with-deps.jar com.onthegomap.planetiler.basemap.Generate -tag="${TAG}"
+
+echo "Formatting..."
+./scripts/format.sh
```
wipfli, line 17: Nice :)

Issue #114, opened 2022-03-05
Generate code during maven build (protobuf/omt schema)


msbarry:
Ideally we wouldn't check generated code (protobuf and OMT schema) into git and instead generate it on the fly as part of the maven build.

Related to #109

wipfli:
With `sudo apt install -y protobuf-compiler` on ubuntu, I can also run the protobuf generation script.

PR #129, opened 2022-03-08
Remove generated code from source control

wipfli:
Remove generated code from source control. I don't know if the probuf generation works on windows. 

- [x] remove generated code
- [x] update CONTRIBUTING.md
- [ ] update ci workflows


wipfli:
#114

msbarry:
Oops sorry @wipfli! I deleted `onthegomap:eclipse-formatter` branch after merging that change which seems to have closed this PR...

Overall my thought with the generated code was to get maven to generate it automatically as part of the build without needing to install any external dependencies.  I think this should be possible for the protobuf (something like https://vlkan.com/blog/post/2015/11/27/maven-protobuf/) and probably also for the openmaptiles schema (something like https://stackoverflow.com/questions/2147115/generating-sources-by-running-a-projects-java-class-in-maven). 

If we need a separate tool or build dependency installed, I'd probably rather leave the generated code checked in until we can make it part of the maven build.

wipfli:
Sounds good to me.

PR #137, opened 2022-03-16
Read shapefiles in UTF-8 mode

wipfli:
Currently, if a shapefile has attributes with letters like ä, ö, ü, etc, the resulting tiles show some strange letters. This pull requests fixes this issue by hard-coding the shapefile mode to UTF-8.

﻿


wipfli:
Current (bad):

image

With this patch (good):

image

This example rendered the following shapefile: https://data.geo.admin.ch/ch.bafu.schutzgebiete-aulav_jagdbanngebiete/

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/ShapefileReader.java
```diff
@@ -125,7 +126,9 @@ private ShapefileDataStore open(Path path) {
       } else {
         throw new IllegalArgumentException("Invalid shapefile input: " + path + " must be zip or shp");
       }
-      return new ShapefileDataStore(uri.toURL());
+      var store = new ShapefileDataStore(uri.toURL());
+      store.setCharset(Charset.forName("UTF8"));
```
msbarry, line 130: Thanks for bringing this up! I noticed the same issue when trying to read the natural earth shapefile.  One concern here is that standard encoding for shapefiles is technically [ISO8859-1](http://en.wikipedia.org/wiki/ISO/IEC_8859-1) so I'd hesitate to always use UTF-8.

It looks like shapefiles that have a different encoding actually contain a ".cpg" file next to the .shp file with the text of the shapefile though, so maybe we could use that if it exists?

msbarry:
Thanks for making the change! Would it be possible to add a minimal test case that verifies a UTF-8 string from a shapefiles is decoded properly?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/ShapefileReader.java
wipfli: Interesting point with the `.cpg`. QGIS seems to get the encoding right without further configuration: 
image

msbarry: Hmm you're right.. the shapefile reader should be doing this automatically.  From the [geotools release page](https://docs.geotools.org/stable/userguide/welcome/upgrade.html) it looks like starting with version 26 (we're on 26.3) they attempt to use cpg file to determine the encoding. I wonder why that's not working here?

wipfli: https://docs.geotools.org/stable/javadocs/org/geotools/data/shapefile/ShapefileDataStore.html#isTryCPGFile-- returns `false`, i.e., by default it does not access the `.cpg` file. Let's check if I can change the behavior with `setTryCPGFile()`...

wipfli: [2d29b3e](https://github.com/onthegomap/planetiler/pull/137/commits/2d29b3e94154695d6c04be67ec992e7d6fbb6a7f)

wipfli:
Works without hardcoding utf-8 now, by using the `.cpg` file. Thanks for the hint. Should I still look into a unit test or do we trust the library?

msbarry:
> Works without hardcoding utf-8 now, by using the `.cpg` file. Thanks for the hint. Should I still look into a unit test or do we trust the library?

The unit test would be helpful! I'm hoping to switch to [Apache SIS](https://sis.apache.org/) at some point so it would be good to make sure it has the same handling. There is a tiny `shapefile.zip` in here. I haven't worked much with shapefiles, do you have an easy way of modifying the shapefile to have a cpg and adding a test UTF-8 attribute to one of the points?

Also, it might be useful to have an extra option on `ShapefileReader` to override the encoding just like you can override the projection.  What do you think?

wipfli:
I edited `planetiler-core/src/test/resources/shapefile.zip` in QGIS:

1. Changed `Van Dorn Street` to `Van Dörn Street` and saved in the non-default encoding UTF-8.
2. Added a `shapefile.cpg` file which is a text file containing the string `UTF8`

I updated the test file `ShapeFileReaderTest.java` to check if it finds the string `Van Dörn Street`. With the patch in this pull request, the test passes. Without, it fails.

wipfli:
Are you suggesting to overload `.addShapefileSource("EPSG:2056", ...)` to something like `.addShapefileSource("UTF8", "EPSG:2056", ...)`?

msbarry:
> I edited `planetiler-core/src/test/resources/shapefile.zip` in QGIS:
> 
> 1. Changed `Van Dorn Street` to `Van Dörn Street` and saved in the non-default encoding UTF-8.
> 2. Added a `shapefile.cpg` file which is a text file containing the string `UTF8`
> 
> I updated the test file `ShapeFileReaderTest.java` to check if it finds the string `Van Dörn Street`. With the patch in this pull request, the test passes. Without, it fails.

Awesome, looks great! Thank you!

> Are you suggesting to overload .addShapefileSource("EPSG:2056", ...) to something like .addShapefileSource("UTF8", "EPSG:2056", ...)?

Yeah I could go either way on this, since there are now 2 optional parameters (source encoding and projection) there could be 4 signatures based on which ones you want to set and 8 if we add a third...I should refactor this to allow come kind of parameter object to make it these kinds of settings easier to add.  Maybe a compromise would be to just add a character encoding argument to the existing method that includes a projection, and if it's null then ignore? Also I'd be fine doing nothing if adding the cpg file is an OK workaround.

wipfli:
I think the `.cpg` file is the simplest solution. It is a text file which contains the encoding, e.g., `UTF8`, as plain text. This is also already documented in the shapefile standard...

msbarry:
Sounds good to me! Thanks for making this change.

PR #138, opened 2022-03-16
Check disk space and memory only with OSM input file

wipfli:
`checkDiskSpace()` and `checkMemory()` estimate resouce requirements based on an OSM input file.
Currently, if no OSM input file is present, e.g., because only a shapefile input is used, Planetiler crashes complaining about the absence of a OSM input file for the resource checks.

With this pull request, the resource checks only run if an OSM file is present.﻿


wipfli:
I am happy to share an example how to render a custom shapefile with planetiler once this issue and #137 are solved.

msbarry:
Oops, good catch! This sounds good - for most other non-OSM sources, there should be very little memory/disk requirements besides the intermediate features and output size.

PR #140, opened 2022-03-21
Update README.md

SomeoneElseOSM:
Fix "planetilerler" typo on "java" line.
Nudge non-Java people into adding methods inside their new class, not outside it.

### DIFF
--- a/planetiler-examples/README.md
+++ b/planetiler-examples/README.md
@@ -56,7 +56,7 @@ public class MyProfile implements Profile {
 }
 ```
 
-Then, implement the `processFeature()` method in your class (add the code before the last closing curly bracket) 
+Then, implement the `processFeature()` method in your class (add the code before the last closing curly bracket)
 that determines what vector tile features to emit for each source feature.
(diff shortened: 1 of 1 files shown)


SomeoneElseOSM:
If someone can translate "The following files had format violations:" into English I'd be happy to resolve...

msbarry:
Thanks for helping improve the docs! Yikes it looks like we had a couple of CI flake-outs here. I re-ran some so they should be fixed now.

> If someone can translate "The following files had format violations:" into English I'd be happy to resolve...

We just added auto-formatting for java and markdown files, I think it's just complaining about the trailing whitespace at the end of line 59 - if you remove that it should be fine, I could also disable the check on markdown files as it doesn't help as much as with Java.

Also for what it's worth, we're going to be exploring some options for config or script-driven style definitions in the near future (see #81) so while you need to write Java code right now to add new styles, you eventually won't need to.

msbarry:
Thank you @SomeoneElseOSM !

PR #142, opened 2022-03-22
Remove unneeded hasTag(...) variant

ZeLonewolf:
This PR removes a variant of the hasTag(...) method that takes exactly two possible tag values as an argument.  This functionality is covered by the varargs version of hasTag(...) that follows it.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
@@ -1,8 +1,9 @@
 package com.onthegomap.planetiler.reader;
 
+import java.util.Map;
+
 import com.onthegomap.planetiler.util.Imposm3Parsers;
 import com.onthegomap.planetiler.util.Parse;
-import java.util.Map;
 
 /** An input element with a set of string key/object value pairs. */
 public interface WithTags {
@@ -28,12 +29,6 @@ public interface WithTags {
 
   default boolean hasTag(String key, Object value) {
     return value.equals(getTag(key));
-  }
-
-  /** Returns true if the value for {@code key} is {@code value1} or {@code value2}. */
-  default boolean hasTag(String key, Object value1, Object value2) {
-    Object actual = getTag(key);
-    return value1.equals(actual) || value2.equals(actual);
   }
 
   /** Returns true if the value for {@code key} is {@code value1} or {@code value2}. */
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
@@ -1,9 +1,8 @@
 package com.onthegomap.planetiler.reader;
 
-import java.util.Map;
-
 import com.onthegomap.planetiler.util.Imposm3Parsers;
 import com.onthegomap.planetiler.util.Parse;
+import java.util.Map;
 
 /** An input element with a set of string key/object value pairs. */
 public interface WithTags {


planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
```diff
@@ -1,8 +1,9 @@
 package com.onthegomap.planetiler.reader;
 
+import java.util.Map;
```
ZeLonewolf, line 3: I'm not sure why this moved, it's what the formatter did on its own 🤔

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
```diff
@@ -30,12 +30,6 @@ default boolean hasTag(String key, Object value) {
     return value.equals(getTag(key));
   }
 
-  /** Returns true if the value for {@code key} is {@code value1} or {@code value2}. */
-  default boolean hasTag(String key, Object value1, Object value2) {
-    Object actual = getTag(key);
-    return value1.equals(actual) || value2.equals(actual);
-  }
-
   /** Returns true if the value for {@code key} is {@code value1} or {@code value2}. */
```
wipfli, line 33: Might want to update the docstring?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
```diff
@@ -30,12 +30,6 @@ default boolean hasTag(String key, Object value) {
     return value.equals(getTag(key));
   }
 
-  /** Returns true if the value for {@code key} is {@code value1} or {@code value2}. */
-  default boolean hasTag(String key, Object value1, Object value2) {
```
msbarry, line 34: This was here to avoid allocating an array when called for the most common cases of 1 or 2 arguments since it's one of the few methods invoked on every single OSM element.  To make this less ambiguous (and to follow Joshua Bloch's Effective Java advice more closely) we should probably change:

```java
default boolean hasTag(String key, Object... values) {
```

to:

```java
default boolean hasTag(String key, Object value1, Object value2, Object... values) {
```

and add a comment to the 0, 1, and 2 arg versions that they are an optimization for small numbers of arguments.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
msbarry: as long as `spotless:check` is happy then 🤷

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/WithTags.java
ZeLonewolf: I was surprised the compiler doesn't cache varag arguments when you're calling it with the same value each time but TIL.
If I'm writing `hasTag("waterway", "stream", "canal", "river")` for example, it probably makes sense to to declare a primitive string array of ["stream", "canal", "river"] that's passed into the vararg argument each time rather then having vararg unpacking occur for each invocation.

msbarry: Yeah looks like it does not cache them:

```java
  public static void test(int... args) {
    System.err.println(args);
  }

  public static void main(String[] args) {
    for (int i = 0; i < 100; i++) {
      test(1, 2, 3);
    }
  }
```

prints:
```
[I@3b79fd76
[I@48c76607
[I@43599640
[I@1f81aa00
[I@6b6776cb
[I@1863d2fe
...
```

So you could do that, or we can add another `hasTag(key, Collection<value>)` option? In practice it probably doesn't have much of an impact - when the number of tags that you're comparing against starts to get large we probably want to start using `MultiExpression` anyway which evaluates a minimal set of the matching expressions for each.

Also I realized that these hasTag methods should probably check that the value is not null before comparing against the values!

msbarry: Another thing you could try is the MultiExpression class I made to optimize running a bunch of these lookups on each element, the test cases should show how to use it:

https://github.com/onthegomap/planetiler/blob/e20b41c88dc397c4fcb6c6e5124cd973ec830b53/planetiler-core/src/test/java/com/onthegomap/planetiler/expression/MultiExpressionTest.java#L161-L179

I ran some tests on my laptop, `hasTag(key, val1)` 12 billion times runs in 32 seconds, `hasTag(key, val1, val2)` runs in 49 seconds, and `hasTag(key, val1, val2, val3)` (vararg) runs in 57 - so it's not a huge penalty, and in the noise relative to geometry processing.  I think it's fine to keep these methods and even use the vararg one for simple profiles that don't do many lookups for each element, but for complex schemas that run many checks against each element you probably want to use MultiExpression.

Issue #86, opened 2022-02-24
Name translation improvements


msbarry:
[LanguageUtils](https://github.com/onthegomap/planetiler/blob/main/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/LanguageUtils.java) is a straight port from openmaptiles logic, but there are a few issues with it.  Please add a comment with any suggestions for improving the logic to assign element names, picking latin/nonlatin names, transliterating, etc. along with some example test cases to illustrate the desired behavior.

1ec5:
When the name contains no “Latin” characters, the code transliterates the text into Latin text using ICU’s `Any-Latin` transliterator:

https://github.com/onthegomap/planetiler/blob/0d727c4302aa0674df9ca1fa49cb988f84c0f789/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Translations.java#L122-L133

This transliterator can be rough coming from some languages, because different schemes are often used depending on the source language, region, and use case. For example:

[<img src="https://user-images.githubusercontent.com/1231218/155564286-63eeb805-03f9-421f-a87e-18444814f8d9.png" width="400" alt="Kiev">](https://zelonewolf.github.io/openstreetmap-americana/#14.16/50.46092/30.51636)<br>_[Cyrillic text](https://unicode-org.github.io/icu/userguide/transforms/general/#cyrillic) is transliterated to Latin according to the ISO&nbsp;9 standard, which is less biased toward a particular language but still differs from the more common transliteration schemes used in each language or country, as seen here in these [Ukrainian](https://en.wikipedia.org/wiki/Romanization_of_Ukrainian#Conventional_romanization_of_proper_names) park names._

[<img src="https://user-images.githubusercontent.com/1231218/155560968-8d8ed30d-b832-4415-be82-699750b490cf.png" width="400" alt="Shei-Pa">](https://zelonewolf.github.io/openstreetmap-americana/#9.71/24.3934/121.2146)<br>_Chinese text appears to be transliterated to Latin using Hanyu Pinyin. A map intended for lay readers would remove diacritics and spaces between syllables of compound words from these pinyin transliterations. Additionally, in Taiwan, the use of Hanyu Pinyin versus Tongyong Pinyin is [a partisan and regional matter](https://en.wikipedia.org/wiki/Pinyin#In_Taiwan). (These names all end in words like “Ecological Protection Area” that would ideally be translated rather than transliterated.)_

[...]

PR #147, opened 2022-03-25
Improve name:latin logic

### LINKED ISSUE #146: [BUG] Limit latin name extraction to valid language codes with no non-latin characters
The most robust fix would be to limit the subkeys to those that would be valid BCP 47 codes, like xx and xx-YY and xx-YY-ZZZZ. But limiting it to the languages in the profile should work too.

I think the revised non-Latin detection code would still be worth pursuing regardless. With just the smaller fix you’re suggesting, there will be cases where an alternative language’s name is arbitrarily chosen just because of a “non-Latin” character.

_Originally posted by @1ec5 in https://github.com/onthegomap/planetiler/issues/86#issuecomment-1076573531_

msbarry:
Simplify/improve latin character regular expressions and limit `name:latin` lookup to just language-specific `name:*` tags and not tags like `name:prefix` or `name:etymology:wikidata`.  Does not attempt to improve transliteration, that will be a future change.

Fixes #146 and partially improves #86.

### DIFF
--- a/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/LanguageUtils.java
+++ b/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/LanguageUtils.java
@@ -54,21 +54,22 @@ import java.util.stream.Stream;
  * <a href="https://github.com/openmaptiles/openmaptiles-tools/blob/master/sql/zzz_language.sql">openmaptiles-tools</a>.
  */
 public class LanguageUtils {
+  // See https://github.com/onthegomap/planetiler/issues/86
 
   // Name tags that should be eligible for finding a latin name.
   // See https://wiki.openstreetmap.org/wiki/Multilingual_names
-  // and https://github.com/onthegomap/planetiler/issues/86
   private static final Predicate<String> VALID_NAME_TAGS =
     Pattern.compile("^name:[a-z]{2,3}(-[a-z]{4})?([-_][a-z]{2,})?(-[a-z]{2})?$", Pattern.CASE_INSENSITIVE)
       .asMatchPredicate();
 
   // Match strings that only contain latin characters.
-  // and https://github.com/onthegomap/planetiler/issues/86
   private static final Predicate<String> ONLY_LATIN = Pattern
     .compile("^[\\P{IsLetter}[\\p{IsLetter}&&\\p{IsLatin}]]+$")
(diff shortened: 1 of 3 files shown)


planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/LanguageUtils.java
```diff
@@ -53,10 +54,22 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * <a href="https://github.com/openmaptiles/openmaptiles-tools/blob/master/sql/zzz_language.sql">openmaptiles-tools</a>.
  */
 public class LanguageUtils {
+  // See https://github.com/onthegomap/planetiler/issues/86
+
+  // Name tags that should be eligible for finding a latin name.
+  // See https://wiki.openstreetmap.org/wiki/Multilingual_names
+  private static final Predicate<String> VALID_NAME_TAGS =
+    Pattern.compile("^name:[a-z]{2,3}(-[a-z]{4})?([-_][a-z]{2,})?(-[a-z]{2})?$", Pattern.CASE_INSENSITIVE)
```
1ec5, line 62: Technically, BCP&nbsp;47 allows a UN&nbsp;M49 region code in place of an ISO&nbsp;3166-1 country code, but [the only extant occurrences](https://overpass-turbo.eu/s/1h7S) are mere coincidences, all of them instances of mistagging.

The `abc-x-extension` syntax is also permitted for unregistered extensions, but [`name:fr-x-gallo`](https://taginfo.openstreetmap.org/keys/name%3Afr-x-gallo) is the only key with non-negligible use, for [Gallo](https://en.wikipedia.org/wiki/Gallo_language).

1ec5 (approved):
Looks good to me at a glance. The notes below are only for completeness, but I’m glad my penchant for proper typography won’t cause OpenStreetMap Americana’s labels to be translated into QIDs going forward. 😅

planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/LanguageUtils.java
msbarry: Thanks! Added those cases and they pass now.  The transliteration improvements will be a larger project, will need to tackle that in a separate change later.

PR #150, opened 2022-03-26
Linting - spelling, tiny cleanups

nyurik:
Grammar fixes, marked one place as `final`, etc.


msbarry:
Thanks for these fixes! Looks like spotless is unhappy with the formatting somewhere. Could you run `mvn spotless:apply`?

nyurik:
Thanks @msbarry , I was a bit lost from the error message trying to figure out what was missing. Fixed.

msbarry:
> Thanks @msbarry , I was a bit lost from the error message trying to figure out what was missing. Fixed.

No worries, I should rearrange those actions to make it more obvious when spotless fails...

Thanks!

Issue #124, opened 2022-03-08
[FEATURE] Add linting


msbarry:
To improve consistency between people who may use different editors, we should adopt a java linter like [checkstyle](https://checkstyle.sourceforge.io/) or [sonarlint](https://www.sonarlint.org/).  You should be able to run it from the command line, through plugins to popular editors, and a github action should highlight any issues as comments on a PR.

msbarry:
Looked into some options:

**[Checkstyle](https://checkstyle.sourceforge.io/) (NO)**
It looks like checkstyle just finds easy syntactical bugs (indentation, spacing, etc.) that should be resolved automatically spotless.

**[Error prone](https://github.com/google/error-prone)**
This is a compiler "plugin" from Google that detects potential issues from analyzing the AST.
Pros: fast, warnings look sane and actionable
Cons: zero support for eclipse/vscode, clunky intellij plugin, doesn't run as you edit, introduces flakiness into compile step
Example PR: #173 

**[Spotbugs](https://github.com/spotbugs/spotbugs)**
This is the successor to [findbugs](http://findbugs.sourceforge.net/) to detect possible errors.
Pros: widely-used, many checks available, good IDE support
Cons: slower and noisier than error prone, will need some tuning to get useful output

**[PMD](https://github.com/pmd/pmd)**
Multi-language code analyzer with java/maven support
Pros: Configurable and output looks reasonable, good IDE support
Cons: Intellij plugin does not run as you type
Example PR:  #174

**[Qodana](https://www.jetbrains.com/qodana/jvm/)**
This is a tool for CI built on the warnings generated by IDEA.
Pros: High quality output, same that I'm used to from IDEA
Cons: Closed-source, needs to run in a docker container, _VERY_ slow
Example PR: #175  

Doesn't look like a clear winner like eslint or golint. Anyone have any suggestions ?

msbarry:
Closed by #177

PR #160, opened 2022-04-08
Declarative schema from configuration file

### LINKED ISSUE #18: [FEATURE] config-based profiles
Right now you need to write Java code to build a custom profile. It might be nice if you could also write a json or yaml file that defines the logic and config for the profile.

ZeLonewolf:
Fixes #18

This PR provides the ability for a user to specify an arbitrary vector schema via a configuration file.  Configuration is via a YML file which specifies the contents of the configured layers.  Documentation for the schema definition format is provided in planetiler-custommap/README.md.

The schema language allows for the following features:
* Ability to define layers and features within layers based on tag filters
* Ability to specify attributes as a function of tags
* Ability to specify feature zoom levels as a function of tagging or feature area
* Ability to specify data sources in a configuration file

A collection of unit tests are provided, as well as a sample schema intended to support OWG vector tile exploration under openstreetmap/operations#565

A typical invocation looks something like:
```
(java/docker launch) generate-custom --schema=schema_file.yml --download --area=rhode-island
```

This PR does not include handling for line merging, which needs a deeper investigation to deal with attributes that have minimum zooms greater than the overall feature minimum zoom, which would allow certain linestring features to merge only at certain lower zooms.


wipfli:
@ZeLonewolf how do I compile and run this on my laptop?

wipfli:
Currently, I get the following error:

```

➜  planetiler git:(main) ✗ gh pr checkout 160
➜  planetiler git:(owg_schema) ✗ ./mvnw -DskipTests=true --projects planetiler-dist -am package

[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for com.onthegomap.planetiler:planetiler-custommap:0.3-SNAPSHOT: Could not find artifact com.onthegomap.planetiler:planetiler-parent:pom:0.3-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 9, column 11
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project com.onthegomap.planetiler:planetiler-custommap:0.3-SNAPSHOT (/home/wipfli/maps/planetiler/planetiler-custommap/pom.xml) has 1 error
[ERROR]     Non-resolvable parent POM for com.onthegomap.planetiler:planetiler-custommap:0.3-SNAPSHOT: Could not find artifact com.onthegomap.planetiler:planetiler-parent:pom:0.3-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 9, column 11 -> [Help 2]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[ERROR] [Help 2] http://cwiki.apache.org/confluence/display/MAVEN/UnresolvableModelException
```

planetiler-custommap/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>planetiler-custommap</artifactId>
+
+  <parent>
+    <groupId>com.onthegomap.planetiler</groupId>
+    <artifactId>planetiler-parent</artifactId>
+    <version>0.3-SNAPSHOT</version>
```
wipfli, line 12: In the basemap profile, we use 0.4. Without this change, I can't compile the project. But with this change it work...
```suggestion
    <version>0.4-SNAPSHOT</version>
```

wipfli:
With the change above, the code compiles. But how do I run it? I tried this but it failed:

```
➜  planetiler git:(owg_schema) ✗ java -cp planetiler-dist/target/*with-deps.jar com.onthegomap.planetiler.custommap.ConfiguredProfile --force=true --area=switzerland

Error: Could not find or load main class com.onthegomap.planetiler.custommap.ConfiguredProfile
Caused by: java.lang.ClassNotFoundException: com.onthegomap.planetiler.custommap.ConfiguredProfile

```

planetiler-custommap/pom.xml
ZeLonewolf: I wish I'd seen this 5 hours ago 🤕

ZeLonewolf:
CLI should look something like:

```
generate-custom --schema=schema_file.yml --download --area=rhode-island
```

planetiler-custommap/README.md
```diff
@@ -0,0 +1,74 @@
...
+* `attribution` - An attribution statement, which may include HTML such as links
+* `sources` - A list of data sources for the schema.  See [Data Sources](#data-sources)
+* `layers` - A list of vector tile layers and their definitions.  See [Layers](#layers)
+
+### Data Sources
+
+A data source contains geospatial objects with tags that are consumed by planetiler.  The configured data sources in the schema provide complete information on how to access those data sources.
+* `name` - Name of this data source, which is referenced in other parts of the schema
+* `type` - Either `shapefile` or `osm`
+* `url` - Location to download the a shapefile from
```
msbarry, line 22: typo: "the a"

msbarry (commented):
Thanks! This is looking like a great start. Starting with some questions on the config syntax, haven't gotten into the code or tests yet...

.gitignore
```diff
@@ -22,3 +22,5 @@ bin/
 TODO
 
 data/
+
+*.mbtiles
```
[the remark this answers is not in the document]
ZeLonewolf: Done in 58da975

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+        value:
+        - river
+        - stream
+        - canal
+    attributes:
+    - key: waterway
+      tagValue: waterway
+    - key: intermittent
+      constantValue: 'true'
+      dataType: bool
```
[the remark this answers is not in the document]
ZeLonewolf: The idea is to normalize different values e.g. "true" vs "yes", similar to what imposm3 does in OpenMapTiles.  I'm not sure how real of a problem that actually is in the database though...

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+      minZoom: 7
+      maxZoom: 14
+    includeWhen:
+      geometry: linestring
+      tag:
+        key: waterway
+        value:
+        - river
+        - stream
+        - canal
```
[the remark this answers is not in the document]
ZeLonewolf: I agree that would be nicer 😅 now I just need to figure out how to make SnakeYaml handle it sanely..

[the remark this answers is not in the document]
ZeLonewolf: Actually, I can make SnakeYaml map it as a Map which, as an added bonus, lets us specify multiple keys to match on for free.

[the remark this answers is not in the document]
ZeLonewolf: Done in 1730f7b!  🎉

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Thinking more about this, it would probably be easier to drop data typing and let the user specify yes/true as a value list in cases where this problem is real.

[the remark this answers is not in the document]
msbarry: Yeah, I was thinking that with yaml or json this is the difference between `true` and `"true"` or `1` and `"1"` - the main issue would be parsing values from osm tags as boolean vs. number

msbarry:
For the tests, one idea I had was to make a yaml file where you could add examples like:

```yaml
- name: Include springs in water layer
   input:
     source: osm
     tags:
       natural: spring
   output:
     layer: water
     type: polygon
     minzoom: 10
     attributes:
       class: lake
```

Then you could run `<java/docker> test-custom --schema=schema.yml --tests=test.yml`, and maybe the owg one could be run as part of the unit tests as well. 

Sometimes when you encounter specific OSM elements you want to make sure they show up

What do you think? This can definitely wait to a subsequent PR.

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/TestAreaSourceFeature.java
```diff
@@ -0,0 +1,40 @@
+package com.onthegomap.planetiler.custommap;
+
+import com.onthegomap.planetiler.geo.GeometryException;
+import com.onthegomap.planetiler.reader.SourceFeature;
+import java.util.Collections;
+import java.util.Map;
+import org.locationtech.jts.geom.Geometry;
+
+public class TestAreaSourceFeature extends SourceFeature {
```
msbarry, line 9: Instead of `TestAreaSourceFeature` and `TestLinestringSourceFeature` you use the built-in `SimpleFeature` factory methods, something like:

```java
SimpleFeature.createFakeOsmFeature(newLineString(0, 0, 1, 1), tags, "osm", null, 1, List.of())
```

or

```java
SimpleFeature.createFakeOsmFeature(newPolygon(0, 0, 1, 0, 1, 1, 0, 0), tags, "osm", null, 1, List.of())
```

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredSingleFeatureTest.java
```diff
@@ -0,0 +1,127 @@
...
+  @BeforeAll
+  private static void setup() {
+    waterTags.put("natural", "water");
+    waterTags.put("water", "pond");
+    waterTags.put("name", "Little Pond");
+
+    motorwayTags.put("highway", "motorway");
+    motorwayTags.put("layer", "1");
+    motorwayTags.put("bridge", "yes");
+  }
```
msbarry, line 92: this could be simplified to:

```java
  private static final Map<String, Object> waterTags = Map.of(
    "natural", "water",
    "water", "pond",
    "name", "Little Pond"
  );
```

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredSingleFeatureTest.java
```diff
@@ -0,0 +1,127 @@
...
+  private static FeatureCollector linestringFeatureCollector() {
+    PlanetilerConfig config = PlanetilerConfig.defaults();
+    FeatureCollector.Factory factory = new FeatureCollector.Factory(config, Stats.inMemory());
+    return factory.get(SimpleFeature.create(TestUtils.newLineString(0, 0, 0.1, 0, 0.1, 0.1, 0, 0), new HashMap<>()));
+  }
+
+  private static Profile configureProfile(String filename) throws Exception {
+    Path staticAttributeConfig = TestConfigurableUtils.pathToResource(filename);
+    Yaml yml = new Yaml();
+    SchemaConfig schema = yml.loadAs(new FileInputStream(staticAttributeConfig.toFile()), SchemaConfig.class);
```
msbarry, line 43: In general I try to avoid `File` where possible and use nio utilities instead, for example: `java.nio.file.Files.newInputStream(staticAttributeConfig)`

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private static void run(Arguments args) throws Exception {
+    Path dataDir = Path.of("data");
+    Path sourcesDir = dataDir.resolve("sources");
+
+    String schemaFile = args.getString(
+      "schema",
+      "Location of YML-format schema definition file",
+      "");
```
msbarry, line 32: Why not `args.inputFile(...)` ? That does a check to ensure it exists as well.  You could also add an `inputFile` variation without a default value that will fail-fast if the argument is not specified.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/SchemaConfig.java
```diff
@@ -0,0 +1,55 @@
+package com.onthegomap.planetiler.custommap.configschema;
+
+import java.util.Collection;
+
+/**
+ * An object representation of a vector tile server schema. This object is mapped to a schema YML file using SnakeYAML.
+ */
+public class SchemaConfig {
```
msbarry, line 8: You can simplify these data model objects quite a bit by using records:

```java
public record SchemaConfig(
  String schemaName,
  String schemaDescription,
  String attribution,
  Collection<DataSource> sources,
  Collection<FeatureLayer> layers
) {}
```

But, to do that you need to do a 2-step parse using snake yaml to parse the YAML in a way that supports anchors and aliases, then use jackson's object mapper which supports records. For an example see the openmaptiles schema parsing code:

https://github.com/onthegomap/planetiler/blob/6a7c6791416c06fe6bbbcdebaf74cf078e1c2ea4/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/Generate.java#L117-L122

Because the record syntax is so compact, I sometimes like to just include the other data model objects as nested classes so the definition looks similar to example config files, but I don't have a strong preference on that.

msbarry (commented):
Got a bit further into the tests and main driver.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,302 @@
...
+   * @param features      - output rendered feature collector
+   */
+  public void processFeature(SourceFeature sourceFeature, FeatureCollector features) {
+
+    Feature f = geometryFactory.apply(features);
+
+    f.setBufferPixels(BUFFER_SIZE)
+      // and also whenever you set a label grid size limit, make sure you increase the buffer size so no
+      // label grid squares will be the consistent between adjacent tiles
+      .setBufferPixelOverrides(ZoomFunction.maxZoom(12, 32));
```
msbarry, line 297: I don't think we need these yet - the default buffer pixel size is 4, so we'll only need to set this when users can override it at the layer and feature-level. And buffer pixel overrides is only necessary for points using the equivalent of `LabelGrid` from postgis

https://github.com/onthegomap/planetiler/blob/6a7c6791416c06fe6bbbcdebaf74cf078e1c2ea4/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java#L217

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,302 @@
...
+        return;
+      }
+      Optional<ZoomFilter> zoomFilterMatch = zfList
+        .stream()
+        .filter(zf -> zf.getTag().matcher().test(sf))
+        .findFirst();
+
+      if (zoomFilterMatch.isPresent()) {
+        ZoomFilter zf = zoomFilterMatch.get();
+        f.setMinZoom(zf.getMinZoom());
```
msbarry, line 92: should you set maxzoom as well?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,302 @@
...
+    int minZoom = configuredMinZoom == null ? 0 : configuredMinZoom.intValue();
+    Function<SourceFeature, Object> attributeValueProducer = attributeValueProducer(attribute);
+
+    TagCriteria attrIncludeWhen = attribute.getIncludeWhen();
+    TagCriteria attrExcludeWhen = attribute.getExcludeWhen();
+
+    Predicate<SourceFeature> attributeTest =
+      attrIncludeWhen == null ? sf -> true : attrIncludeWhen.matcher();
+    Predicate<SourceFeature> attributeExcludeTest =
+      attrExcludeWhen == null ? sf -> true : attrExcludeWhen.matcher().negate();
```
msbarry, line 204: We should use `Expression` instead of `matcher()` as well.  We might even want to use an indexed `MultiExpression` that returns the attribute processors to check based on the element's tags.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,302 @@
...
+
+    if (minTileCoverage != null && minTileCoverage > 0.0) {
+      attributeZoomProducer = attributeZoomThreshold(minTileCoverage, minZoom);
+    } else {
+      attributeZoomProducer = sf -> minZoom;
+    }
+
+    return (sf, f) -> {
+      if (attributeTest.test(sf) && attributeExcludeTest.test(sf)) {
+        f.setAttrWithMinzoom(tagKey, attributeValueProducer.apply(sf), attributeZoomProducer.apply(sf));
```
msbarry, line 218: there is a performance cost to `setAttrWithMinzoom` over just `setAttr` - we should check if there's actually a minzoom configured and if not then use `setAttr`

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/FeatureGeometryType.java
```diff
@@ -0,0 +1,7 @@
+package com.onthegomap.planetiler.custommap.configschema;
+
+public enum FeatureGeometryType {
+  polygon,
+  linestring,
+  point
```
msbarry, line 6: We already have a `com.onthegomap.planetiler.geo.GeometryType` enum for these 3 geometry types, but I'm also wondering how we'll handle things like polygon centroids? If this will grow based on different kinds of geometry transformations we might do then it's fine to have separate.

If you're going to keep here, then we can also still use conventional upper-camel case but preserve the yaml representations by using the `@JsonNode` annotation, for example:

https://github.com/onthegomap/planetiler/blob/6a7c6791416c06fe6bbbcdebaf74cf078e1c2ea4/planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java#L265-L271

msbarry (commented):
Got through `ConfiguredProfile`/`ConfiguredFeature` API - done with my initial pass.

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Resolved in b0f3d60

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+    - key: waterway
+      tagValue: waterway
+    - key: intermittent
+      constantValue: 'true'
+      dataType: bool
+      includeWhen:
+        tag:
+          key: intermittent
+          value:
+          - true
```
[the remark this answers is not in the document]
ZeLonewolf: Having removed the boolean data type stuff, it's now "yes" instead of true.  Do we want to infer booleans like that?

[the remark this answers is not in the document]
msbarry: I don't quite follow - I thought that `true`/`TRUE`/`True`/`yes` were all valid aliases for true in yaml?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * A map feature, configured from a YML configuration file
+ */
+public class ConfiguredFeature {
+
+  private Collection<String> sources;
+  private Predicate<SourceFeature> geometryTest;
+  private Function<FeatureCollector, Feature> geometryFactory;
+  private Predicate<SourceFeature> tagTest;
+  private BiConsumer<SourceFeature, Feature> zoomConfig;
```
msbarry, line 33: These could be final since they are only set once from the constructor. I don't want to clutter the review with comments like this so I'm also looking into linters in #124 that add annotations to the PR like an IDE would show a warning - if you have any experience with or preference on those, let me know!

ZeLonewolf: I would personally recommend sonarlint!

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: I was thinking that the most straightforward thing to do would be to treat all OSM tag value as strings and not try to do datatype conversions.  So e.g.:

```
      includeWhen:
        tag:
          key: intermittent
          value:
          - "yes"
```
The quotes would enforce treating the value as a string rather than boolean.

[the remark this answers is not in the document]
msbarry: Ahh got it, I see how we want the data type to coerce the string to a Boolean or number. If you use the getBoolean method on the feature it uses the same logic as imposm3 to determine if a string is truthy: https://github.com/onthegomap/planetiler/blob/2742bef48d09147d6b4e1f425ac7822882499887/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Imposm3Parsers.java#L86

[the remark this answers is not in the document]
ZeLonewolf: Oh that makes things much easier.  It looks like we can support boolean, long, direction, and string in that way.

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+        tag:
+          key: highway
+          value:
+          - service
+      minZoom: 13
+    - key: name
+      tagValue: name
+      minZoom: 12
+    - key: layer
+      tagValue: layer
```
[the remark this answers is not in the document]
ZeLonewolf: Can you envision cases more complex than unit conversion?  It seems like we could just add unit of measure options into SourceFeature getXXX methods.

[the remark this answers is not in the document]
msbarry: From openmaptiles: at least stripping nonlatin characters from names (but that might be handled by separate code), computing building render height from `building:{height,min_height,level,min_level}` tags.

For simple schemas you could decide to push all of that logic to client-side style expressions, but for people to adopt the config approach we need some extension point for flexibility.  To get past this issue for now while keeping the option open for later, we could just use `contantValue`/`tagValue` as shortcuts but be open to eventually having `value` that accepts more of an expression-like or embedded script syntax.

wipfli:
> how do I compile and run this on my laptop?

Like this:
```
./scripts/build.sh
java -jar planetiler-dist/target/*-with-deps.jar generate-custom --schema=planetiler-custommap/samples/owg_simple.yml
```

This info should be in one of the readmes I think.

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Improved input data type implementation implemented in ceeb099

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+schemaDescription: Simple vector tile schema
+attribution: <a href="https://www.openstreetmap.org/copyright" target="_blank">&copy;
+  OpenStreetMap contributors</a>
+sources:
+- name: water_polygons
+  type: shapefile
+  url: https://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip
+- name: osm
+  type: osm
+  area: geofabrik:rhode-island
```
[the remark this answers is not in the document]
ZeLonewolf: How do I pass `geofabrik://rhode-island` to planetiler?

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Data typing resolved in ceeb099

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
msbarry: Just set it as the URL and Downloader should translate if it's not on disk.

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Documented in #179

planetiler-custommap/samples/owg_simple.yml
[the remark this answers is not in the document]
ZeLonewolf: Resolved in b9c89a4

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredSingleFeatureTest.java
ZeLonewolf: Fixed in c5162ef
Yet another post-Java 8 functionality I wasn't aware of!

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredProfile.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    layers.stream().forEach(layer -> {
+      String layerName = layer.getName();
+      Collection<FeatureItem> featureDefs = layer.getFeatures();
+
+      featureDefs.forEach(feature -> {
+        features.add(new ConfiguredFeature(layerName, feature));
+      });
+    });
```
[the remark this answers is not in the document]
ZeLonewolf: Ah, `var`, another new one for me.  Fixed in 0641e0a.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
ZeLonewolf: Fixed in 7cd1efb

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/TestAreaSourceFeature.java
ZeLonewolf: Fixed in d110150

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
ZeLonewolf: Fixed specific issue noted here in 97ba5b4

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
+schemaName: OWG Simple Schema
+schemaDescription: Simple vector tile schema
+attribution: <a href="https://www.openstreetmap.org/copyright" target="_blank">&copy;
+  OpenStreetMap contributors</a>
+sources:
```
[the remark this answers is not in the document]
ZeLonewolf: I changed the sources list to a map in d065e1d, is that what you were thinking?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java
ZeLonewolf: Resolved in fa7fe2f

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredSingleFeatureTest.java
ZeLonewolf: Resolved in f8f0838

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java
ZeLonewolf: Resolved in fa7fe2f

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,192 @@
...
+  type: osm
+  area: geofabrik:rhode-island
+layers:
+- name: water
+  features:
+  - sources:
+    - osm
+    zoom:
+      minZoom: 0
+      maxZoom: 14
```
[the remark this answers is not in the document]
ZeLonewolf: Fixed in 3ba11eb

planetiler-custommap/samples/owg_simple.yml
```diff
@@ -0,0 +1,160 @@
+schemaName: OWG Simple Schema
```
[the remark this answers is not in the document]
ZeLonewolf: I was thinking that we didn't want the samples packaged in the jar, though maybe for important ones like openmaptiles (someday) and the owg schema (if it matures) it might make sense for them to be pre-configured options?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/SchemaConfig.java
ZeLonewolf: Resolved in 9e494eb

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/FeatureGeometryType.java
ZeLonewolf: Resolved in 896f3cd

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
ZeLonewolf: Resolved in b7c4fd8

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `tagValue` - Value of the attribute in the tile, as copied from the value of the specified tag key.  If neither constantValue nor tagValue are specified, the default behavior is to set the tag value equal to the input value (pass-through)
+* `includeWhen` - A filter specification which determines whether to include this attribute.  If unspecified, the attribute will be included unless excluded by `excludeWhen`.  See [Tag Filters](#tag-filters)
+* `excludeWhen` - A filter specification which determines whether to exclude this attribute.  This rule is applied after `includeWhen`.  If unspecified, no exclusion filter is applied.  See [Tag Filters](#tag-filters)
+* `minZoom` - The minimum zoom at which to render this attribute.
+
+### Tag Filters
+
+A tag filter matches an object based on its tagging.  Multiple key entries may be specified:
+* `<key>:` - Match objects that contain this key.
+* `  <value>` - A single value or a list of values.  Match objects in the specified key that contains one of these values.  If no values are specified, this will match any value tagged with the specified key.
```
wipfli, line 72: Can you add an example?

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `minTileCoverSize` - include objects of a certain geometry size, where 1.0 means "is the same size as a tile at this zoom".
+* `includeWhen` - A tag specification which determines which features to include.  If unspecified, all features from the specified sources are included.  See [Tag Filters](#tag-filters)
+* `excludeWhen` - A tag specification which determines which features to exclude.  This rule is applied after `includeWhen`.  If unspecified, no exclusion filter is applied.  See [Tag Filters](#tag-filters)
+* `attributes` - Specifies the attributes that should be rendered into the tiles for this feature, and how they are constructed.  See [Attributes](#attributes)
+
+### Data Types
+
+Specifies that certain tag key should have their values treated as being a certain data type.
+* `<data type>` - One of BOOLEAN, STRING, DIRECTION, LONG
+* `<list of values>` - A list of strings corresponding to keys that are treated as this data type.
```
wipfli, line 44: Can you add an example?

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
+# Configurable Planetiler Schema
+
+It is possible to customize planetiler's output from configuration files.  This is done using the parameter:
+`--schema=schema_file.yml`
+
+The schema file provides information to planetiler about how to construct the tiles and which layers, features, and attributes will be posted to the file.  Schema files are in [YAML](https://yaml.org) format.
+
```
wipfli, line 7: Please add a link to an example schema file at the end of the intro paragraph.

wipfli: I would recommend renaming this file to "SCHEMA.md" or so. And then you could add a readme which has the structure 
- what is this good for 
- install and build instructions 
- known limitations / todos

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+
+    return MultiExpression.of(zfList
+      .stream()
+      .map(zf -> MultiExpression.entry(zf.minZoom(), zf.tag().matcher(tagValueProducer)))
+      .toList())
+      .index();
+  }
+
+  /**
+   * Produce logic that generates attribute values based on configuration and input data. If both a constantValue
```
wipfli, line 108:
```suggestion
   * Produces logic that generates attribute values based on configuration and input data. If both a constantValue
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+
+    //Configure logic for each attribute in the output tile
+    feature.attributes()
+      .stream()
+      .map(this::attributeProcessor)
+      .forEach(attributeProcessors::add);
+  }
+
+  /**
+   * Generate zoom-based inclusion logic based on tagging and overall zoom limits. Tag-based zoom logic will override
```
wipfli, line 82:
```suggestion
   * Generates zoom-based inclusion logic based on tagging and overall zoom limits. Tag-based zoom logic will override
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    if (tagVal != null) {
+      return tagValueProducer.getValueProducer(tagVal);
+    }
+
+    //Default to producing a tag identical to the input
+    return tagValueProducer.getValueProducer(attribute.key());
+  }
+
+  /**
+   * Generate logic which determines the minimum zoom level for a feature based on a configured pixel size limit.
```
wipfli, line 133:
```suggestion
   * Generates logic which determines the minimum zoom level for a feature based on a configured pixel size limit.
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+    try {
+      return (int) (Math.log(minTilePercent / sf.area()) / LOG4);
+    } catch (GeometryException e) {
+      return 14;
+    }
+  }
+
+  /**
+   * Determine the minimum zoom of a feature, based on its tagging.
```
wipfli, line 162:
```suggestion
   * Determines the minimum zoom of a feature, based on its tagging.
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    return switch (type) {
+      case POLYGON -> fc -> fc.polygon(layerName);
+      case LINE -> fc -> fc.line(layerName);
+      case POINT -> fc -> fc.point(layerName);
+      default -> throw new IllegalArgumentException("Unhandled geometry type " + type);
+    };
+  }
+
+  /**
+   * Generate a test for whether a source feature is of the correct geometry to be included in the tile.
```
wipfli, line 240:
```suggestion
   * Generates a test for whether a source feature is of the correct geometry to be included in the tile.
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    return switch (type) {
+      case POLYGON -> SourceFeature::canBePolygon;
+      case LINE -> SourceFeature::canBeLine;
+      case POINT -> SourceFeature::isPoint;
+      default -> throw new IllegalArgumentException("Unhandled geometry type " + type);
+    };
+  }
+
+  /**
+   * Test to determine whether to include a source feature in the tiles
```
wipfli, line 255:
```suggestion
   * Determines whether to include a source feature in the tiles.
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    if (!geometryTest.test(sourceFeature)) {
+      return false;
+    }
+
+    //Check for matching tags
+    return tagTest.evaluate(sourceFeature);
+  }
+
+  /**
+   * Generate a tile feature based on a source feature
```
wipfli, line 276:
```suggestion
   * Generates a tile feature based on a source feature.
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+public class TagCriteria extends HashMap<String, Object> {
+
+  private static final long serialVersionUID = 1L;
+
+  /**
+   * Returns a function that determines whether a source feature matches any of the entries in this specification
+   * 
+   * @param sf source feature
```
wipfli, line 20: Is this still up-to-date?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    //Default to producing a tag identical to the input
+    return tagValueProducer.getValueProducer(attribute.key());
+  }
+
+  /**
+   * Generate logic which determines the minimum zoom level for a feature based on a configured pixel size limit.
+   * 
+   * @param minTilePercent - minimum percentage of a tile that a feature must cover to be shown
+   * @param minZoom        - global minimum zoom for this feature
+   * @param zoomTagConfig
```
wipfli, line 137: Remove

wipfli (commented):
I am really impressed that you pushed all this so far @ZeLonewolf. Thanks a lot for your contribution! I am looking forward to making maps with this in the future :)

msbarry:
Updates are looking good! I'm working my way through the code changes, but first a couple of overall comments/questions on the config format to take advantage of my naivety reading through the config files as a new user:

1. I like the `attribute` definitions! All of the info about them is collected to one place and `- key: name` as a simple pass-through is convenient
2. As a convention should we be using camel-case? I usually see snake case in comparable config files in this space like ([imposm3](https://imposm.org/docs/imposm3/latest/mapping.html) and [maplibre](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/) uses [kebab (?)](https://stackoverflow.com/questions/11273282/whats-the-name-for-hyphen-separated-case) case) but I don't usually see camel case in yaml files.  We can use `@JsonProperty` annotation to bind our java field names to any yaml/json representation so we don't need to let the java convention leak out into config file
3. The nested `zoom` section is a bit hard to parse when you read through it... what do you think about pulling minzoom/maxzoom up to the base config for feature, and allow support either `minzoom: constant` or `minzoom: default: 12, overrides: ...` to nest `zoom_filter` under minzoom/maxzoom? That would make it a bit more clear that zoom_filter controls, and keep the min zoom logic collected into one place?
4. It's not super clear reading through the config what `dataTypes` controls.  Would a name like "input mapping/parsing" make sense since it controls what raw tags values are mapped to? Also, it's global right now, I wonder if that's too limiting... could it change by source? across multiple usages of the same tag from a source for different layers/attributes?

msbarry:
Also 5) for min tile cover size/min length/min area, does "portion of tile size" unit make sense? Openmaptiles (and planetiler code API) use pixels as their unit assuming a 256x256px tile.

We might need a neutral third party to weigh in here... @wipfli  😄

wipfli:
I would do coverage relative to the full tile area.

planetiler-core/src/main/java/com/onthegomap/planetiler/config/Arguments.java
```diff
@@ -208,6 +208,17 @@ public Path file(String key, String description, Path defaultValue) {
     return file;
   }
 
+  /** Returns a {@link Path} parsed from {@code key} argument which may or may not exist. */
```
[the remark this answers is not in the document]
ZeLonewolf: May not have been the best choice of phrase :-D

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+A layer contains a thematically-related set of features.
+* `name` - Name of this layer
+* `features` - A list of features contained in this layer.  See [Features](#features)
+
+### Features
+
+A feature is a defined set of objects that meet specified filter criteria.
+* `zoom` - Specifies the zoom inclusion rules for this feature.  See [Zoom Specification](#feature-zoom-specification).
+* `geometry` - Include objects of a certain geometry type.  Options are `POLYGON`, `LINE`, or `POINT`.
+* `minTileCoverSize` - include objects of a certain geometry size, where 1.0 means "is the same size as a tile at this zoom".
```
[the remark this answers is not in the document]
ZeLonewolf: Fixed in a37da53

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `includeWhen` - A tag specification which determines which features to include.  If unspecified, all features from the specified sources are included.  See [Tag Filters](#tag-filters)
+* `excludeWhen` - A tag specification which determines which features to exclude.  This rule is applied after `includeWhen`.  If unspecified, no exclusion filter is applied.  See [Tag Filters](#tag-filters)
+* `attributes` - Specifies the attributes that should be rendered into the tiles for this feature, and how they are constructed.  See [Attributes](#attributes)
+
+### Data Types
+
+Specifies that certain tag key should have their values treated as being a certain data type.
+* `<data type>` - One of BOOLEAN, STRING, DIRECTION, LONG
```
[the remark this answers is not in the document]
ZeLonewolf: Fixed in a37da53

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `<list of values>` - A list of strings corresponding to keys that are treated as this data type.
+
+### Feature Zoom Specification
+
+Specifies the zoom inclusion rules for this feature.
```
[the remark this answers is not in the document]
ZeLonewolf: There isn't any fundamental limit to zoom as far as I know.  Could render tiles down to z20 if you wanted to...

planetiler-custommap/README.md
ZeLonewolf: Added in b8d9659

ZeLonewolf:
Linking schema / data typing concepts in:
https://gist.github.com/msbarry/4b22aadb16bdf9e9ec82f6ccd20fe5d4

ZeLonewolf:
I'm planning to proceed with the "Define new tag names" approach, and to allow inputs/outputs to be the same.

msbarry:
> I'm planning to proceed with the "Define new tag names" approach, and to allow inputs/outputs to be the same.

Sounds good for a long term approach we can expand to but I think we want to keep the `key: type` shortcut you currently have in the config files for the most common case when input/output tags are the same.

ZeLonewolf:
> > I'm planning to proceed with the "Define new tag names" approach, and to allow inputs/outputs to be the same.
> 
> Sounds good for a long term approach we can expand to but I think we want to keep the `key: type` shortcut you currently have in the config files for the most common case when input/output tags are the same.

Agreed - I believe both approaches can co-exist.  I see the `key: type` shortcut happening post any "define new tag names" that we do.

planetiler-custommap/README.md
ZeLonewolf: At less than 100 lines, I'm not sure we really need separate files to document this?

ZeLonewolf: I've added links to the sample schemas and test cases in the opening section, which should resolve the ask for examples.

ZeLonewolf:
@msbarry tag data typing and improved zoom declarations are now implemented!  I'll pause here for review of the config YAML format to see if there's any more changes we need to make.

msbarry:
> @msbarry tag data typing and improved zoom declarations are now implemented! I'll pause here for review of the config YAML format to see if there's any more changes we need to make.

Thanks @ZeLonewolf! `snake_case` changes look good

For `tag_mappings` ... Name sounds good ✅  but I'm wondering if we might want it to be of the format:

```yaml
tag_mapping:
  output_tag_name:
    input: input_key_name
    type: boolean
```

Since we will eventually have output tags derived from multiple (or no) input tags, like:

```yaml
tag_mapping:
  length: geometry_length
  zorder:
    type: wayzorder # derived from highway, railway, tunnel, bridge tag
```

Or do you think those kinds of mappings would be handled elsewhere? (like at the attribute output configs)

msbarry:
And for `min_zoom`/`max_zoom` - the 2 use-uses are feature visibility, and attribute visibility. It's not clear to me with the latest iteration if/how feature visibility changing by tag values is handled?  I was picturing something like:

```yaml
- name: road
  features:
  - sources:
    - osm
    geometry: line
    include_when:
      highway:
      - motorway
      - trunk
      - primary
      - secondary
      - tertiary
      - motorway_link
      - trunk_link
      - primary_link
      - secondary_link
      - tertiary_link
      - unclassified
      - residential
      - living_street
      - service
      - track
    min_zoom:
      default: 4
      overrides:
      - value: 5
        tag:
          highway: trunk
      - value: 7
        tag:
          highway: primary
      - value: 8
        tag:
          highway: secondary
      - value: 9
        tag:
          highway:
          - tertiary
          - motorway_link
          - trunk_link
          - primary_link
          - secondary_link
          - tertiary_link
      - value: 11
        tag:
          highway:
          - unclassified
          - residential
          - living_street
      - value: 12
        tag:
          highway: track
      - min_zoom: 13
        tag:
          highway: service
```

Or we could possibly tie it into `include_when` so that becomes more like maplibre's [data and camera expressions](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#data-expressions) like the `[filter](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#filter)` layer attribute.

ZeLonewolf:
I was thinking about the zoom configuration, and I'm wondering if it can be even more simple, such as below.  I added a `required` field, such that if `required=yes` then the feature will be suppressed if the attribute is not set.

```
- name: road
  features:
  - sources:
    - osm
    geometry: line
    include_when:
      highway:
      - motorway
      - trunk
      - primary
      - secondary
      - tertiary
      - motorway_link
      - trunk_link
      - primary_link
      - secondary_link
      - tertiary_link
      - unclassified
      - residential
      - living_street
      - service
      - track
    attributes:
    - key: highway
      required: true
      min_zoom: 4
      min_zoom_by_value:
        trunk: 5
        primary: 7
        secondary: 8
        tertiary: 9
        motorway_link: 9
        trunk_link: 9
        primary_link: 9
        secondary_link: 9
        tertiary_link: 9
        unclassified: 11
        residential: 11
        living_street: 11
        track: 12
        service: 13
    - key: bridge
      include_when:
        bridge: true
      min_zoom: 11
    - key: tunnel
      include_when:
        bridge: true
      min_zoom: 11
    - key: service
      include_when:
        highway: service
      min_zoom: 13
    - key: name
      min_zoom: 12
    - key: layer
      min_zoom: 13
```

msbarry:
> I was thinking about the zoom configuration, and I'm wondering if it can be even more simple, such as below. I added a `required` field, such that if `required=yes` then the feature will be suppressed if the attribute is not set.

Hmm... something like that _may_ make sense for very specific scenarios, but I think that logically separating out the feature min/max zoom from the attribute min/max zoom would be the most clear.

ZeLonewolf:
Okay, I would tweak this just a touch with:

```
  zoom:
    default: 4
    overrides:
    - min: 5
      tag:
        highway: trunk
    - min: 7
      tag:
        highway: primary
    - min: 8
      tag:
        highway: secondary
    - min: 9
      tag:
        highway:
        - tertiary
        - motorway_link
        - trunk_link
        - primary_link
        - secondary_link
        - tertiary_link
    - min: 11
      tag:
        highway:
        - unclassified
        - residential
        - living_street
    - min: 12
      tag:
        highway: track
    - min: 13
      tag:
        highway: service
```

This would allow expansion to add "max" as well.

msbarry:
Hmm what does 

```yaml
zoom:
  default: 4
```

Mean? Seems like the root should have min/max zoom?

ZeLonewolf:
Oops, revised:
```
  zoom:
    min: 4
    overrides:
    - min: 5
      tag:
        highway: trunk
    - min: 7
      tag:
        highway: primary
    - min: 8
      tag:
        highway: secondary
    - min: 9
      tag:
        highway:
        - tertiary
        - motorway_link
        - trunk_link
        - primary_link
        - secondary_link
        - tertiary_link
    - min: 11
      tag:
        highway:
        - unclassified
        - residential
        - living_street
    - min: 12
      tag:
        highway: track
    - min: 13
      tag:
        highway: service
```

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/TestConfigurableUtils.java
```diff
@@ -0,0 +1,22 @@
...
+    return resolve(Path.of("planetiler-custommap", "src", "test", "resources", "invalidSchema", resource));
+  }
+
+  static Path pathToSample(String resource) {
+    return resolve(Path.of("planetiler-custommap", "src", "main", "resources", "samples", resource));
+  }
+
+  private static Path resolve(Path pathFromRoot) {
+    Path cwd = Path.of("").toAbsolutePath();
+    return cwd.resolveSibling(pathFromRoot);
```
msbarry, line 20: Would `pathFromRoot.toAbsolutePath()` work here instead?

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/SchemaYAMLLoadTest.java
```diff
@@ -0,0 +1,35 @@
...
+  @Test
+  void testSchemaLoad() throws IOException {
+    testSchemasInFolder(Paths.get("src", "main", "resources", "samples"));
+    testSchemasInFolder(Paths.get("src", "test", "resources", "validSchema"));
+  }
+
+  private void testSchemasInFolder(Path path) throws IOException {
+    var schemaFiles = Files.walk(path)
+      .filter(p -> p.getFileName().toString().endsWith(".yml"))
+      .toList();
```
msbarry, line 27: maybe also assert the list is not empty just to make sure we're not just failing to find the files.

planetiler-custommap/src/main/resources/samples/highway_areas.yml
```diff
@@ -0,0 +1,37 @@
...
+    - osm
+    geometry: polygon
+    min_zoom: 14
+    include_when:
+      man_made: bridge
+    attributes:
+    - key: man_made
+      constant_value: bridge
+    - key: layer
+    - key: surface
```
msbarry, line 37: Whenever I touch one of these files, my IDE saves in a different format than this. Can we add yaml formatting to spotless or eclipse formatter to ensure consistent formatting across IDEs?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/ZoomOverride.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Map;
+
+/**
+ * Configuration item that instructs the renderer to override the default zoom range for features which contain specific
+ * tag combinations.
+ */
+public record ZoomOverride(
+  Byte min,
+  Byte max,
```
msbarry, line 11: bytes are a little inconvenent to work with, and we're not short on space, could we just use `Integer` ?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
```diff
@@ -0,0 +1,60 @@
...
+import static com.onthegomap.planetiler.expression.Expression.matchField;
+
+import com.onthegomap.planetiler.custommap.TagValueProducer;
+import com.onthegomap.planetiler.expression.Expression;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+
+public class TagCriteria extends HashMap<String, Object> {
```
msbarry, line 13: I try to avoid extending java collections, could we do this instead through a record that wraps a hashmap? Or provide the matcher function as a static method on a utility class?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/SchemaConfig.java
```diff
@@ -0,0 +1,65 @@
+package com.onthegomap.planetiler.custommap.configschema;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * An object representation of a vector tile server schema. This object is mapped to a schema YML file using SnakeYAML.
+ */
+public class SchemaConfig {
```
msbarry, line 10: Can we make this a record?  From what I can see the main issue is that `attribution` has a default value. If there's not a built-in jackson way to set a default value with annotations, we could work around that in the constructor or the getter, something like:

```java
@Override
public String attribution() {
  return attribution == null ? "default value" : attribution;
}
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/TagValueProducer.java
```diff
@@ -0,0 +1,92 @@
...
+      LONG_DATATYPE, WithTags::getLong
+    );
+
+  static final Map<String, UnaryOperator<Object>> dataTypeParse =
+    Map.of(
+      STRING_DATATYPE, s -> s,
+      BOOLEAN_DATATYPE, Parse::bool,
+      DIRECTION_DATATYPE, Parse::direction,
+      LONG_DATATYPE, Parse::parseLong
+    );
```
msbarry, line 39: Can we rename these to tag mapping type or something since it's no longer called data type in the schema?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import org.yaml.snakeyaml.Yaml;
+
+public class ConfiguredMapMain {
```
msbarry, line 18: A little bit of javadoc on these classes in custommap package would help people find their way around.  It's not an API people will write code against, but just something to explain that this is the main driver that parses the yaml file into a profile and invokes PlanetilerRunner or something etc.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java
```diff
@@ -0,0 +1,79 @@
...
+    return mapper.convertValue(parsed, SchemaConfig.class);
+  }
+
+  private static void configureSource(Planetiler planetiler, Path sourcesDir, String sourceName, DataSource source)
+    throws URISyntaxException {
+
+    DataSourceType sourceType = source.type();
+
+    switch (sourceType) {
+      case OSM:
```
msbarry, line 63: minor: we should try to use the newer switch expression format now that it's available:

```java
switch (sourceType) {
  case OSM -> {
    do stuff, no return needed!
  }
  case SHAPEFILE -> {
    do stuff, no return
  }
  default -> ...
}
```

planetiler-custommap/README.md
```diff
@@ -0,0 +1,86 @@
+# Configurable Planetiler Schema
+
+It is possible to customize planetiler's output from configuration files.  This is done using the parameter:
+`--schema=schema_file.yml`
+
+The schema file provides information to planetiler about how to construct the tiles and which layers, features, and attributes will be posted to the file.  Schema files are in [YAML](https://yaml.org) format.
+
+For examples, see [samples](src/main/resources/samples) or [test cases](src/test/resources/validSchema).
```
msbarry, line 8: Even though planetiler is 0.4, I think we should also probably put a note in here that the format may change as we iterate on it, to set expectations around API stability.

msbarry (commented):
Looked through the updated yaml schemas and schema java files.   Will look through the meat of the logic in `custommap` package next.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeometryType.java
```diff
@@ -8,8 +9,11 @@
 
 public enum GeometryType {
   UNKNOWN(VectorTileProto.Tile.GeomType.UNKNOWN, 0),
+  @JsonAlias("point")
```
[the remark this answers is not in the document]
ZeLonewolf: Resolved in 56ef4c7

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -247,6 +247,14 @@ default boolean contains(Predicate<Expression> filter) {
    */
   boolean evaluate(WithTags input, List<String> matchKeys);
 
+  /**
+   * Returns true if this expression matches an input element.
+   *
+   * @param input the input element
+   * @return true if this expression matches the input element
+   */
+  boolean evaluate(WithTags input);
```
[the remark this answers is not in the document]
ZeLonewolf: Resolved in 0735c86

planetiler-core/src/test/java/com/onthegomap/planetiler/expression/ExpressionTestUtil.java
```diff
@@ -0,0 +1,18 @@
+package com.onthegomap.planetiler.expression;
+
+import static com.onthegomap.planetiler.TestUtils.newPoint;
+
+import com.onthegomap.planetiler.reader.SimpleFeature;
+import com.onthegomap.planetiler.reader.SourceFeature;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ExpressionTestUtil {
```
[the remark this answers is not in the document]
ZeLonewolf: Resolved in 03b1f44

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+A layer contains a thematically-related set of features.
+* `name` - Name of this layer
+* `features` - A list of features contained in this layer.  See [Features](#features)
+
+### Features
+
+A feature is a defined set of objects that meet specified filter criteria.
+* `zoom` - Specifies the zoom inclusion rules for this feature.  See [Zoom Specification](#feature-zoom-specification).
+* `geometry` - Include objects of a certain geometry type.  Options are `POLYGON`, `LINE`, or `POINT`.
```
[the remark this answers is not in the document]
ZeLonewolf: Old version, this has been resolved since.

planetiler-custommap/README.md
```diff
@@ -0,0 +1,72 @@
...
+* `minTileCoverSize` - include objects of a certain geometry size, where 1.0 means "is the same size as a tile at this zoom".
+* `includeWhen` - A tag specification which determines which features to include.  If unspecified, all features from the specified sources are included.  See [Tag Filters](#tag-filters)
+* `excludeWhen` - A tag specification which determines which features to exclude.  This rule is applied after `includeWhen`.  If unspecified, no exclusion filter is applied.  See [Tag Filters](#tag-filters)
+* `attributes` - Specifies the attributes that should be rendered into the tiles for this feature, and how they are constructed.  See [Attributes](#attributes)
+
+### Data Types
```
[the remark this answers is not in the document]
ZeLonewolf: Review is on an outdated rev.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/SchemaConfig.java
ZeLonewolf: Fixed in 4ffd340

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/SchemaYAMLLoadTest.java
ZeLonewolf: Fixed in f47ed8e

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/ZoomOverride.java
ZeLonewolf: Fixed in be4fafb

planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/TestConfigurableUtils.java
ZeLonewolf: No, because I needed to resolve from the project root, unless there's some better way to reference that point in the directory structure?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,317 @@
...
+  public void processFeature(SourceFeature sourceFeature, FeatureCollector features) {
+
+    var minZoom = zoomOverride.getOrElse(sourceFeature, featureMinZoom);
+
+    Feature f = geometryFactory.apply(features)
+      .setMinZoom(minZoom);
+    if (featureMaxZoom != null) {
+      f.setMaxZoom(featureMaxZoom);
+    }
+    attributeProcessors.forEach(p -> p.accept(sourceFeature, f));
```
msbarry, line 315: Minor comment - when the lambda version isn't much shurter than the non-lambda I usually tend to prefer non-lambda for simplicity and to avoid some allocations

```java
for (var processor : attributeProcessors) {
  processor.accept(sourceFeature, processor);
}
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,317 @@
...
+
+  /**
+   * Generates a factory method which creates a {@link Feature} from a {@link FeatureCollector} of the appropriate
+   * geometry type.
+   * 
+   * @param layerName - name of the layer
+   * @param type      - type of geometry
+   * @return geometry factory method
+   */
+  private Function<FeatureCollector, Feature> geometryMapFeature(String layerName, GeometryType type) {
```
msbarry, line 255: Can be static - also we might want to pull these into the constructors for each `GeometryType` 🤔   ... feel free to ignore though, that could be a future refactor as it doesn't impact internal or external API.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,317 @@
...
+    var attributeValueProducer = attributeValueProducer(attribute);
+
+    var attrIncludeWhen = attribute.includeWhen();
+    var attrExcludeWhen = attribute.excludeWhen();
+
+    var attributeTest =
+      Expression.and(
+        attrIncludeWhen == null ? Expression.TRUE : matcher(attrIncludeWhen, tagValueProducer),
+        attrExcludeWhen == null ? Expression.TRUE : Expression.not(matcher(attrExcludeWhen, tagValueProducer))
+      );
```
msbarry, line 224: call simplify in the preprocessing step to avoid some unnecessary work at processing time

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -0,0 +1,317 @@
...
+    var minZoom = zoomOverride.getOrElse(sourceFeature, featureMinZoom);
+
+    Feature f = geometryFactory.apply(features)
+      .setMinZoom(minZoom);
+    if (featureMaxZoom != null) {
+      f.setMaxZoom(featureMaxZoom);
+    }
+    attributeProcessors.forEach(p -> p.accept(sourceFeature, f));
+  }
+}
```
msbarry, line 317: Nice job here! Hard to believe all of the logic to evaluate config-driven schemas boils down to just ~300 lines

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  public static Expression matcher(Map<String, Object> map, TagValueProducer tagValueProducer) {
+
+    if (map.isEmpty()) {
+      return Expression.TRUE;
+    }
+
+    List<Expression> tagExpressions = new ArrayList<>();
+
+    map.entrySet()
```
msbarry, line 30: A standard `for (var entry : map.entrySet()) { ...` might be more straightforward here.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
```diff
@@ -0,0 +1,59 @@
...
+      .stream()
+      .forEach(
+        entry -> {
+          if (entry.getValue() == null) {
+            //If only a key is provided, with no value, match any object tagged with that key.
+            tagExpressions.add(
+              matchField(entry.getKey()));
+          } else if (entry.getValue() instanceof Collection) {
+            Collection<?> values =
+              (Collection<?>) entry.getValue();
```
msbarry, line 40:
```java
} else if (entry.getValue() instanceof Collection<?> values) {
```

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
```diff
@@ -0,0 +1,59 @@
...
+import static com.onthegomap.planetiler.expression.Expression.matchField;
+
+import com.onthegomap.planetiler.custommap.TagValueProducer;
+import com.onthegomap.planetiler.expression.Expression;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+public class TagCriteria {
```
msbarry, line 13: Now that this isn't a part of the schema, could we just move this static utility into `custommap` so it's near the other similar utilities?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/TagCriteria.java
ZeLonewolf: @msbarry major refactor in 848507a, please review and see what you think?

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
ZeLonewolf: Thanks!  ~300 lines of logic and 1400 lines of test cases :-D

planetiler-custommap/src/test/resources/validSchema/local_path.yml
```diff
@@ -0,0 +1,17 @@
+schema_name: Test Case Schema
+schema_description: Test case tile schema
+attribution: Test attribution
+sources:
+  osm:
+    type: osm
+    url: geofabrik:rhode-island
```
msbarry, line 7: should this specify a local path?

msbarry (commented):
Looks good! Pushed a couple of minor changes and added #247 to track some followup tasks after this lands. I'll plan to merge a couple of other in-flight changes, put out a 0.5.0 release then land this PR.

planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/FeatureItem.java
```diff
@@ -0,0 +1,17 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.onthegomap.planetiler.geo.GeometryType;
+import java.util.Collection;
+import java.util.Map;
+
+public record FeatureItem(
+  Collection<String> sources,
+  @JsonProperty("min_zoom") Integer minZoom,
+  @JsonProperty("max_zoom") Integer maxZoom,
+  GeometryType geometry,
```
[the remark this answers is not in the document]
ZeLonewolf: I think it would be a flag or enumeration.  A centroid is still a representation of an area IMO.

planetiler-custommap/src/test/resources/validSchema/local_path.yml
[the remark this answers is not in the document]
ZeLonewolf: Oh right.  Yes, it should.

fphilippe:
Hi everyone, I've understood that this PR is already integrated the main base code ? Do you know when a release will be compiled ?

wipfli:
I think it is already released. Try something like the following: 

```
mkdir data/
cd data/
https://raw.githubusercontent.com/onthegomap/planetiler/main/planetiler-custommap/src/main/resources/samples/shortbread.yml
cd ..
docker run -v "$(pwd)/data":/data ghcr.io/onthegomap/planetiler:latest generate-custom --schema=/data/shortbread.yml --download
```

fphilippe:
Thank you @wipfli for your quick answer. I got it, but the jar file is not available in [releases](https://github.com/onthegomap/planetiler/releases).

PR #190, opened 2022-04-23
Data type support for Expression / MultiExpression

ZeLonewolf:
Unblocks #160 

This PR allows developers to apply `Expression` and `MultiExpressions` to imposm3-style data types.  Data typing support is needed for declarative schemas (#160) so that users can do matching typed data.  For example, declaring an OSM tag as a boolean would map `yes`/`true`/`1` to the boolean value **`true`**.

The key change is adding a new method to `Expression` called `matchAnyTyped`, which accepts an additional parameter where the datatype may be specified, and if so, call the appropriate getBoolean/getDirection/getLong methods on SourceFeature when performing `Expression` tests.


planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -238,7 +272,7 @@ record And(List<Expression> children) implements Expression {
 
     @Override
     public String toString() {
-      return "and(" + listToString(children) + ")";
+      return "and(" + Expression.listToString(children) + ")";
```
ZeLonewolf, line 275: I had to do this (and others like it) to keep Spotless happy.

planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/generated/OpenMapTilesSchema.java
```diff
@@ -165,9 +165,11 @@ final class FieldValues {
     /** Complex mappings to generate attribute values from OSM element tags in the water layer. */
     final class FieldMappings {
       public static final MultiExpression<String> Class =
-        MultiExpression.of(List.of(MultiExpression.entry("dock", matchAny("waterway", "dock")),
+        MultiExpression.of(List.of(
+          MultiExpression.entry("dock", matchAny("waterway", "dock")),
           MultiExpression.entry("river", or(matchAny("water", "river"), matchAny("waterway", "riverbank"))),
-          MultiExpression.entry("lake", matchAny("waterway", "")), MultiExpression.entry("ocean", FALSE)));
+          MultiExpression.entry("lake", matchAny("natural", "water"))
+        ));
```
ZeLonewolf, line 172: This changes the logic to match OMT's logic - waterway=dock is a `dock`, river areas are `river` and all other water areas are water.  The previous waterway='' matchAny didn't make much sense and gave inexplicable errors I wasn't able to track down.  This version matches OMTs logic and passes all water layer test cases.

ZeLonewolf:
See also openmaptiles/openmaptiles#1377

planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/generated/OpenMapTilesSchema.java
ZeLonewolf: If the order of expressions in the list is non-deterministic, then we'll need to change this logic.

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -31,6 +34,19 @@ public interface Expression {
   String POLYGON_TYPE = "polygon";
   String RELATION_MEMBER_TYPE = "relation_member";
 
+  String STRING_DATATYPE = "string";
+  String BOOLEAN_DATATYPE = "boolean";
+  String DIRECTION_DATATYPE = "direction";
+  String LONG_DATATYPE = "long";
+
+  static final Map<String, BiFunction<SourceFeature, String, Object>> dataTypeGetter =
+    Collections.unmodifiableMap(Map.of(
```
msbarry, line 43: Also (on a side note since this code will go away) `Map.of` returns an immutable map so you don't need to wrap it with an `Collections.unmodifiableMap`

ZeLonewolf: One of many poorly-conceived attempts at patching the non-deterministic behavior!  😬

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -31,6 +34,19 @@ public interface Expression {
...
+      STRING_DATATYPE, SourceFeature::getString,
+      BOOLEAN_DATATYPE, SourceFeature::getBoolean,
+      DIRECTION_DATATYPE, SourceFeature::getDirection,
+      LONG_DATATYPE, SourceFeature::getLong
+    ));
+
```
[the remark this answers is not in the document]
ZeLonewolf: I've changed this to take in the BiFunction as a parameter.  This will allow us to not only specify which getter, but also load on any other formatting we might light to add as a pre-processing step, which I anticipate will be useful to the declarative schema effort.

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -359,11 +387,18 @@ public boolean evaluate(SourceFeature input, List<String> matchKeys) {
 
     @Override
     public String toString() {
-      return "matchAny(" + Format.quote(field) + ", " + values.stream().map(Format::quote)
+      return "matchAny(" + Format.quote(field) + ", " + values.stream().map(Expression::stringifyTypedObject)
```
msbarry, line 390: This doesn't encode the getter to use so we'll lose that when it's serialized. Only generate.java needs code to compile from `toString` so we could also rename this to `generateCode` and if the getter is not default then throw an unsupported operation exception?

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -359,11 +387,18 @@ public boolean evaluate(SourceFeature input, List<String> matchKeys) {
...
         .collect(Collectors.joining(", ")) + ")";
     }
   }
 
+  private static String stringifyTypedObject(Object o) {
+    if (o instanceof String) {
+      return Format.quote(o.toString());
+    }
+    return o.toString();
+  }
```
msbarry, line 400: `stringifyTypedObject` loses the numeric type, and `Integer.valueOf(1)` is not equal to `Long.valueOf(1)` so it might be safter to do an instanceof check for a list of supported types and emit strings like `6L` that encode the right numeric type.

_But..._ I also realize that we don't use values anywhere other than comparing string representations so maybe it's fine to just encode everything as a string?

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -317,28 +343,30 @@ public boolean evaluate(SourceFeature input, List<String> matchKeys) {
...
   record MatchAny(
-    String field, List<String> values, Set<String> exactMatches, List<String> wildcards, boolean matchWhenMissing
+    String field, List<?> values, Set<String> exactMatches, List<String> wildcards, boolean matchWhenMissing,
+    BiFunction<SourceFeature, String, Object> valueGetter
   ) implements Expression {
 
     private static final Pattern containsPattern = Pattern.compile("^%(.*)%$");
 
-    MatchAny(String field, List<String> values) {
+    MatchAny(String field, BiFunction<SourceFeature, String, Object> valueGetter, List<?> values) {
```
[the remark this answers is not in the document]
ZeLonewolf: I've made the change, however -- WithTags doesn't include any geometry checks, so I had to do some type checking when converting MatchType.

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
ZeLonewolf: @msbarry would it be easier to just undo the toString() stuff since generate.java is only ever going to invoke the non-typed version of Expression?
\

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -88,8 +89,33 @@ static MatchAny matchAny(String field, String... values) {
...
    * {@code values} can contain exact matches, "%text%" to match any value containing "text", or "" to match any value.
    */
-  static MatchAny matchAny(String field, List<String> values) {
-    return new MatchAny(field, values);
+  static MatchAny matchAny(String field, List<?> values) {
+    return new MatchAny(field, SourceFeature::getTag,
+      values.stream()
+        //Ensure that we can handle List<String>, List<Long>, etc
+        .map(Object.class::cast)
+        .toList());
```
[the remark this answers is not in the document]
ZeLonewolf: No longer needed since I switched from <Object> to <?> generics!

msbarry (approved):
looks good!

PR #201, opened 2022-04-30
Implement matchField testing for Expression.not

ZeLonewolf:
Fixes #200

This PR implements missing matchField key testing for `Expression.not` tests.

PR #203, opened 2022-04-30
Add unit tests for XOR and XNOR MultiExpression tests

ZeLonewolf:
This PR adds unit tests to verify that XOR and XNOR logic combinations within `MultiExpression`s produce the expected logical outcome.


ZeLonewolf:
Fixed by #201

PR #205, opened 2022-04-30
Add unit tests for and/or combinations with constant booleans

ZeLonewolf:
This PR adds unit tests to confirm that AND/OR Expressions when combined with Expression.TRUE/FALSE perform as expected


ZeLonewolf:
Supplanted by #206

Issue #202, opened 2022-04-30
[BUG] NOR expression + matchAny fails


ZeLonewolf:
**Describe the bug**
"NOR" logic fails to match when combining `Expression.not` + `Expression.or` + `matchAny`.  This issue is similar to #200 and is not corrected by the fix in #201.

**To Reproduce**
The following unit test added to MultiExpressionTest recreates this behavior:

```
  @Test
  void testNor() {
    var index = MultiExpression.of(List.of(
      entry("a", not(or(
        matchAny("key1", "val1"),
        matchAny("key2", "val2")
      )))
    )).index();
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1", "key2", "val2")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1", "key2", "val2", "key3", "val3")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "no", "key2", "val2")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1", "key2", "no")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key2", "val2")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags()));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key1", "no", "key2", "no")));
  }
```

The test case fails on the very last assertSameElements test.

**Expected behavior**
Test case pases

**Environment (please complete the following information):**

- Hardware: [HP Laptop]
- OS: [Windows 11]
- Java version and distribution: [Oracle JDK 17]

ZeLonewolf:
While we're at it, a NAND operation fails also:

```

  @Test
  void testNand() {
    var index = MultiExpression.of(List.of(
      entry("a", not(and(
        matchAny("key1", "val1"),
        matchAny("key2", "val2")
      )))
    )).index();
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1", "key2", "val2")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key1", "val1", "key2", "val2", "key3", "val3")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key1", "no", "key2", "val2")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key1", "val1", "key2", "no")));
[...]

PR #204, opened 2022-04-30
Fix nand/nor Expression bugs

ZeLonewolf:
Closes #202

I don't know if this fix is right or not, but it makes all the test cases pass...

Issue #200, opened 2022-04-30
[BUG] not Expression on matchField fails in MultiExpression


ZeLonewolf:
Blocks #160

**Describe the bug**
When invoking `Expression.not` on `matchField(...)` expressions within a `MultiExpression`, the inversion logic fails

**To Reproduce**
Proposed is to add the following unit tests to MultiExpressionTest, which is an expansion of the logic in testSingleMatchField.  These test cases currently fail.
```
  @Test
  void testSingleMatchField() {
    var index = MultiExpression.of(List.of(
      entry("a", matchField("key"))
    )).index();
    matchFieldCheck(index);
  }

  @Test
  void testDoubleInverseMatchField() {
    var index = MultiExpression.of(List.of(
      entry("a", not(not(matchField("key"))))
    )).index();
    matchFieldCheck(index);
  }

  private void matchFieldCheck(Index<String> index) {
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key", "value")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key", "")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key", "value2", "otherkey", "othervalue")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key2", "value", "key3", "value")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key2", "value")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key2", "no")));
    assertSameElements(List.of(), index.getMatches(featureWithTags()));
  }

  @Test
  void testInverseMatchField() {
    var index = MultiExpression.of(List.of(
      entry("a", not(matchField("key")))
    )).index();
    assertSameElements(List.of(), index.getMatches(featureWithTags("key", "value")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key", "")));
    assertSameElements(List.of(), index.getMatches(featureWithTags("key", "value2", "otherkey", "othervalue")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key2", "value", "key3", "value")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key2", "value")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags("key2", "no")));
    assertSameElements(List.of("a"), index.getMatches(featureWithTags()));
  }
```

[...]

PR #206, opened 2022-05-01
More robust multiexpression filtering

msbarry:
Proposed more robust fix for logic issues @ZeLonewolf has noticed in #205 #204 #203 #202  #201 and #200

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
@@ -13,6 +13,8 @@ import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A framework for defining and manipulating boolean expressions that match on input element.
@@ -27,6 +29,7 @@ import java.util.stream.Stream;
  * </pre>
  */
 public interface Expression {
+  Logger LOGGER = LoggerFactory.getLogger(Expression.class);
 
   String LINESTRING_TYPE = "linestring";
   String POINT_TYPE = "point";
@@ -132,8 +135,11 @@ public interface Expression {
     seen.add(simplified);
     while (true) {
       simplified = simplifyOnce(simplified);
-      if (seen.contains(simplified) || seen.size() > 100) {
+      if (seen.contains(simplified)) {
         return simplified;
+      }
+      if (seen.size() > 1000) {
+        throw new IllegalStateException("Infinite loop while simplifying expression " + initial);
       }
       seen.add(simplified);
     }
@@ -151,6 +157,8 @@ public interface Expression {
         return FALSE;
       } else if (not.child == FALSE) {
         return TRUE;
+      } else if (not.child instanceof MatchAny any && any.values.equals(List.of(""))) {
+        return matchField(any.field);
       }
       return not;
     } else if (expression instanceof Or or) {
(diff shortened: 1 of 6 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/expression/MultiExpression.java
```diff
@@ -220,38 +225,38 @@ public boolean isEmpty() {
...
+          getRelevantKeys(expression,
+            key -> keyToExpressions.computeIfAbsent(key, k -> new HashSet<>()).add(expressionValue));
         }
       }
       // create immutable copies for fast iteration at matching time
-      constantTrueExpressionList = List.copyOf(constants);
+      if (!always.isEmpty()) {
+        LOGGER.warn("{} expressions will be evaluated for every element:", always.size());
+        for (var expression : always) {
+          LOGGER.warn("    {}", expression.expression.generateJavaCode());
```
ZeLonewolf, line 251: Fix should be to surround the warn call with `if(LOGGER.isWarnEnabled()) {`

ZeLonewolf:
Fixes #202

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/MultiExpression.java
msbarry: There are a few cases like this that I think I want to tackle in one pass (and leave highlighted so I don't miss one). I'd prefer to go the route of a logger framework that takes a lambda that only gets invoked if that level is needed. There's also some other logging issues to clear up like `core` should just depend on slf4j and the runnable thing (`dist` module or a project that depends on this) should bind slf4j to whatever implementation it wants.

ZeLonewolf: Good plan.  Looks like https://github.com/kwon37xi/slf4j-lambda exists

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/MultiExpression.java
```diff
@@ -62,45 +65,65 @@ private static <T> void visitExpressions(SourceFeature input, List<Match<T>> res
...
-      acceptKey.accept(any.field());
+    // if a sub-expression must always be evaluated, then either the whole expression must always be evaluated
+    // or there is another part of the expression that limits the elements on which it must be evaluated, so we can
+    // ignore keys from this sub-expression.
+    if (!mustAlwaysEvaluate(exp)) {
+      if (exp instanceof Expression.And and) {
+        and.children().forEach(child -> getRelevantKeys(child, acceptKey));
+      } else if (exp instanceof Expression.Or or) {
+        or.children().forEach(child -> getRelevantKeys(child, acceptKey));
+      } else if (exp instanceof Expression.Not) {
```
ZeLonewolf, line 96: No need to keep this conditional, it's an unnecessary check.

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -132,9 +135,12 @@ private static Expression simplify(Expression initial) {
     seen.add(simplified);
     while (true) {
       simplified = simplifyOnce(simplified);
-      if (seen.contains(simplified) || seen.size() > 100) {
+      if (seen.contains(simplified)) {
         return simplified;
       }
+      if (seen.size() > 1000) {
+        throw new IllegalStateException("Infinite loop while simplifying expression " + initial);
+      }
```
ZeLonewolf, line 143: I was unable to compose an expression that triggers this as an actual infinite loop, though since it's only done at simplify time it's probably fine as a fail-safe.

msbarry: yeah, I'd like to leave it in just in case there's ever an unforeseen case where simplify logic goes off the rails

ZeLonewolf (approved):
This looks great, thanks for improving this logic!

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/MultiExpression.java
msbarry: also slf4j v2 api will support lambdas. I'd like to use that but it's been in alpha stage since 2019...? https://mvnrepository.com/artifact/org.slf4j/slf4j-api

msbarry:
thanks for the review! glad to fix these issues now before we start to handle arbitrary expressions coming in from user-defined schema configs.

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/MultiExpression.java
ZeLonewolf: It's a logging framework, how bad could alpha be?  🤔

Issue #167, opened 2022-04-08
[FEATURE] Deduplicate MBTiles output


msbarry:
The biggest bottleneck during mbtiles writing is writing repeated ocean tiles:

<img width="872" alt="image" src="https://user-images.githubusercontent.com/1480504/162418476-d018160b-b5b2-415a-8a9b-05ebb6193a58.png">

It's possible to reduce from ~15 minutes to ~9 minutes by using 2 separate mbtiles tables plus a view that adheres to mbtiles spec:

- tile -> blob ID
- blob ID -> blob data

This also reduces output mbtiles files size by 20%

bbilger:
are you already working on this @msbarry ? Otherwise I'd open a PR in the next few days with a (possible) fix. I already have something working, I just need to polish it a bit.

msbarry:
Thanks for reaching out @bbilger! I'm working through performance-related issues in this order: https://github.com/onthegomap/planetiler/projects/2 so was planning to do this at some point but hadn't started on it yet.  Feel free to submit a PR!

My initial thought for this was to set up a benchmark to profile mbtiles write and read performance and if the deduplicated format wasn't faster both for writes and for reads, then we'd likely want leave both formats in place so a caller can choose based on the tradeoffs.

A couple of random ideas I was thinking of trying:

1. encoding tile x/y/z as a single integer  like [TileCoord](https://github.com/onthegomap/planetiler/blob/891537e2bc43adba4439d8b233dce2b9abf5ce27/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java#L41-L81) does to see if that improves size/write/read performance
2. splitting into 2 tables: tile ID -> blob ID and blob ID -> data (I'm guessing this hurts read performance due to the 2 lookups)
3. splitting into 2 tables: tile ID -> blob and "repeated tiles" blob ID -> data, where the first table normally holds the full tile as blob data, but when it detects a small repeated tile, instead store a single byte that points to an ID in the "repeated tiles" table so the view would treat that single byte as an index in the "repeated tiles" table at read time (similar to [this approach](https://stackoverflow.com/questions/68684759/storage-optimizing-sqlite-blob-to-act-as-data-or-foreign-key))

If you're just interested in doing #2 that's fine, we might just need to be able to switch between the old and new behavior with a command-line option if reads are slower.

bbilger:
Great, will open a PR for 2 in the following days.

[...]

Issue #168, opened 2022-04-08
[FEATURE] Option to skip writing filled tiles?


msbarry:
MBtiles writing is a single-threaded bottleneck when generating output for the planet that takes up at least ~15 minutes mostly due to repeated ocean tiles:

<img width="882" alt="image" src="https://user-images.githubusercontent.com/1480504/162419450-8e091b7b-db76-447d-a4ba-a9a94d0b0398.png">

It's possible to reduce mbtiles output phase to ~4 minutes (and reduce mbtiles size by 25%) by skipping filled tiles.  Some clients can handle this by requesting parent tiles and overzooming when they get a 404. But some can't (or it might not be desirable) so this should be an opt-in feature if we want to do it at all.

PR #219, opened 2022-05-09
add compact DB mode (--compact-db) to de-duplicate mbtiles output

bbilger:
# Context

this addresses point 2 of https://github.com/onthegomap/planetiler/issues/167#issuecomment-1115483329

# Overview

the compact DB mode splits the tiles table into tiles_shallow and tiles_data
- tiles_shallow contains the coordinates plus a reference on the data ID
- tiles_data contains the data ID plus the actual tile data

this allows to de-duplicate content since multiple tiles can reference the same data

in this mode, tiles is realized as a view that joins the two tables tiles_shallow and tiles_data

# Impact

- it can significantly de-crease the mbtiles file size iff there's many duplicates (ocean tiles) - in case of Australia: 3.1GB -> 1.7GB
- it has little to no impact on writing mbtiles (if no dupes it's slower, and the more dupes the faster but not significantly)
- reading from the mbtiles is 10-15% slower because of the indirection over the view

some more insights can be found here: [PR-impact.txt](https://github.com/onthegomap/planetiler/files/8655715/PR-impact.txt)

# Concept

The main idea to realize this is to create a hash over the tile data, and build a map between that hash and a data ID.

# Failed Approaches

I was a bit surprised that the impact on writing the mbtiles is little to none. The only explanation I have is that we have to do more inserts than before (tiles_shallow, tiles_data). I tried to offload this "dual insert" from code to the DB. So in planetiler I produced just one insert but the insert was into a view - which then either just inserted into tiles_shallow or both tiles_shallow and tiles_data. This, however, increased the overall mbtiles generation time by almost 20%.

# Next

I'd like to address point 1 in https://github.com/onthegomap/planetiler/issues/167#issuecomment-1115483329  either in this or a follow-up PR

### DIFF
--- a/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
+++ b/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
@@ -11,6 +11,7 @@ import com.onthegomap.planetiler.mbtiles.TileEncodingResult;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalInt;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -52,7 +53,7 @@ class VerifyMonacoTest {
         VectorTile.encodeGeometry(point(0, 0)),
         Map.of()
       )));
-      writer.write(new TileEncodingResult(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), false, null));
+      writer.write(new TileEncodingResult(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), OptionalInt.empty()));
     }
     assertInvalid(mbtiles);
   }
--- a/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
+++ b/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
@@ -51,7 +51,7 @@ class VerifyMonacoTest {
         VectorTile.encodeGeometry(point(0, 0)),
         Map.of()
       )));
-      writer.write(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()));
+      writer.write(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), 1);
     }
     assertInvalid(mbtiles);
   }
--- a/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
+++ b/planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
@@ -7,6 +7,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import com.onthegomap.planetiler.VectorTile;
 import com.onthegomap.planetiler.geo.TileCoord;
 import com.onthegomap.planetiler.mbtiles.Mbtiles;
+import com.onthegomap.planetiler.mbtiles.TileEncodingResult;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
@@ -51,7 +52,7 @@ class VerifyMonacoTest {
         VectorTile.encodeGeometry(point(0, 0)),
         Map.of()
       )));
-      writer.write(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), 1);
+      writer.write(new TileEncodingResult(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), false, null));
     }
     assertInvalid(mbtiles);
   }
--- a/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
+++ b/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
@@ -7,20 +7,18 @@ import com.onthegomap.planetiler.mbtiles.Mbtiles;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.LongSummaryStatistics;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class BenchmarkMbtilesRead {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(BenchmarkMbtilesWriter.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(BenchmarkMbtilesRead.class);
 
   private static final String SELECT_RANDOM_COORDS =
     "select tile_column, tile_row, zoom_level from tiles order by random() limit ?";
@@ -28,8 +26,8 @@ public class BenchmarkMbtilesRead {
   public static void main(String[] args) throws Exception {
 
     Arguments arguments = Arguments.fromArgs(args);
-    int repetitions = arguments.getInteger("bench_repetitions", "number of repetitions", 100);
-    int nrTileReads = arguments.getInteger("bench_nr_tile_reads", "number of tiles to read", 10000);
+    int repetitions = arguments.getInteger("bench_repetitions", "number of repetitions", 10);
+    int nrTileReads = arguments.getInteger("bench_nr_tile_reads", "number of tiles to read", 500_000);
 
 
     List<String> mbtilesPaths = new ArrayList<>();
@@ -66,7 +64,7 @@ public class BenchmarkMbtilesRead {
       }
     }
 
-    Map<String, Double> avgIndividualReadPerDb = new HashMap<>();
+    Map<String, Double> avgReadOperationsPerSecondPerDb = new HashMap<>();
     for (String dbPathStr : mbtilesPaths) {
       Path dbPath = Path.of(dbPathStr);
       List<ReadResult> results = new LinkedList<>();
@@ -76,20 +74,14 @@ public class BenchmarkMbtilesRead {
       for (int rep = 0; rep < repetitions; rep++) {
         results.add(readEachTile(randomCoordsToFetchPerRepetition, dbPath));
       }
-      var totalStats = results.stream().mapToLong(ReadResult::totalDuration).summaryStatistics();
-      LOGGER.info("totalReadStats: {}", totalStats);
+      var readOperationsPerSecondStats =
+        results.stream().mapToDouble(ReadResult::readOperationsPerSecond).summaryStatistics();
+      LOGGER.info("readOperationsPerSecondStats: {}", readOperationsPerSecondStats);
 
-      LongSummaryStatistics individualStats = results.stream().map(ReadResult::individualReadStats)
-        .collect(Collector.of(LongSummaryStatistics::new, LongSummaryStatistics::combine, (left, right) -> {
-          left.combine(right);
-          return left;
-        }));
-      LOGGER.info("individualReadStats:  {}", individualStats);
-
-      avgIndividualReadPerDb.put(dbPathStr, individualStats.getAverage());
+      avgReadOperationsPerSecondPerDb.put(dbPathStr, readOperationsPerSecondStats.getAverage());
     }
 
-    List<String> keysSorted = avgIndividualReadPerDb.entrySet().stream()
+    List<String> keysSorted = avgReadOperationsPerSecondPerDb.entrySet().stream()
       .sorted((o1, o2) -> o1.getValue().compareTo(o2.getValue()))
       .map(Map.Entry::getKey)
       .toList();
@@ -98,36 +90,35 @@ public class BenchmarkMbtilesRead {
     for (int i = 0; i < keysSorted.size() - 1; i++) {
       for (int j = i + 1; j < keysSorted.size(); j++) {
         String db0 = keysSorted.get(i);
-        double avg0 = avgIndividualReadPerDb.get(db0);
+        double avg0 = avgReadOperationsPerSecondPerDb.get(db0);
         String db1 = keysSorted.get(j);
-        double avg1 = avgIndividualReadPerDb.get(db1);
+        double avg1 = avgReadOperationsPerSecondPerDb.get(db1);
 
         double diff = avg1 * 100 / avg0 - 100;
 
-        LOGGER.info("\"{}\" vs \"{}\": avgs reads up by {}%", db0, db1, diff);
+        LOGGER.info("\"{}\" vs \"{}\": avg read operations per second improved by {}%", db0, db1, diff);
       }
     }
   }
 
   private static ReadResult readEachTile(List<TileCoord> coordsToFetch, Path dbPath) throws IOException {
-    LongSummaryStatistics individualFetchDurations = new LongSummaryStatistics();
     try (var db = Mbtiles.newReadOnlyDatabase(dbPath)) {
       db.getTile(0, 0, 0); // trigger prepared statement creation
       var totalSw = Stopwatch.createStarted();
       for (var coordToFetch : coordsToFetch) {
-        var sw = Stopwatch.createStarted();
         if (db.getTile(coordToFetch) == null) {
           throw new IllegalStateException("%s should exist in %s".formatted(coordToFetch, dbPath));
         }
-        sw.stop();
-        individualFetchDurations.accept(sw.elapsed(TimeUnit.NANOSECONDS));
       }
       totalSw.stop();
-      return new ReadResult(totalSw.elapsed(TimeUnit.NANOSECONDS), individualFetchDurations);
+      return new ReadResult(totalSw.elapsed(), coordsToFetch.size());
     }
   }
 
-  private record ReadResult(long totalDuration, LongSummaryStatistics individualReadStats) {}
-
-
+  private record ReadResult(Duration duration, int coordsFetchedCount) {
+    double readOperationsPerSecond() {
+      double secondsFractional = duration.toNanos() / 1E9;
+      return coordsFetchedCount / secondsFractional;
+    }
+  }
 }
--- a/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
+++ b/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
@@ -28,6 +28,7 @@ public class BenchmarkMbtilesRead {
     Arguments arguments = Arguments.fromArgs(args);
     int repetitions = arguments.getInteger("bench_repetitions", "number of repetitions", 10);
     int nrTileReads = arguments.getInteger("bench_nr_tile_reads", "number of tiles to read", 500_000);
+    int preWarms = arguments.getInteger("bench_pre_warms", "number of pre warm runs", 3);
 
 
     List<String> mbtilesPaths = new ArrayList<>();
@@ -70,6 +71,10 @@ public class BenchmarkMbtilesRead {
       List<ReadResult> results = new LinkedList<>();
 
       LOGGER.info("working on {}", dbPath);
+
+      for (int preWarm = 0; preWarm < preWarms; preWarm++) {
+        readEachTile(randomCoordsToFetchPerRepetition, dbPath);
+      }
 
       for (int rep = 0; rep < repetitions; rep++) {
         results.add(readEachTile(randomCoordsToFetchPerRepetition, dbPath));
--- a/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
+++ b/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
@@ -1,0 +1,133 @@
+package com.onthegomap.planetiler.benchmarks;
+
+import com.google.common.base.Stopwatch;
+import com.onthegomap.planetiler.config.Arguments;
+import com.onthegomap.planetiler.geo.TileCoord;
+import com.onthegomap.planetiler.mbtiles.Mbtiles;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.LongSummaryStatistics;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BenchmarkMbtilesRead {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BenchmarkMbtilesWriter.class);
+
+  private static final String SELECT_RANDOM_COORDS =
+    "select tile_column, tile_row, zoom_level from tiles order by random() limit ?";
+
+  public static void main(String[] args) throws Exception {
+
+    Arguments arguments = Arguments.fromArgs(args);
+    int repetitions = arguments.getInteger("bench_repetitions", "number of repetitions", 100);
+    int nrTileReads = arguments.getInteger("bench_nr_tile_reads", "number of tiles to read", 10000);
+
+
+    List<String> mbtilesPaths = new ArrayList<>();
+    for (int i = 0;; i++) {
+      String mbtilesPathStr = arguments.getString("bench_mbtiles" + i, "the mbtiles file to read from", null);
+      if (mbtilesPathStr == null) {
+        break;
+      }
+      mbtilesPaths.add(mbtilesPathStr);
+    }
+
+    if (mbtilesPaths.isEmpty()) {
+      throw new IllegalArgumentException("pass one or many paths to the same mbtiles file");
+    }
+
+    mbtilesPaths.stream().map(File::new).forEach(f -> {
+      if (!f.exists() || !f.isFile()) {
+        throw new IllegalArgumentException("%s does not exists".formatted(f));
+      }
+    });
+
+    List<TileCoord> randomCoordsToFetchPerRepetition = new LinkedList<>();
+
+    try (var db = Mbtiles.newReadOnlyDatabase(Path.of(mbtilesPaths.get(0)))) {
+      try (var statement = db.connection().prepareStatement(SELECT_RANDOM_COORDS)) {
+        statement.setInt(1, nrTileReads);
+        var rs = statement.executeQuery();
+        while (rs.next()) {
+          int x = rs.getInt("tile_column");
+          int y = rs.getInt("tile_row");
+          int z = rs.getInt("zoom_level");
+          randomCoordsToFetchPerRepetition.add(TileCoord.ofXYZ(x, (1 << z) - 1 - y, z));
+        }
+      }
+    }
+
+    Map<String, Double> avgIndividualReadPerDb = new HashMap<>();
+    for (String dbPathStr : mbtilesPaths) {
+      Path dbPath = Path.of(dbPathStr);
+      List<ReadResult> results = new LinkedList<>();
+
+      LOGGER.info("working on {}", dbPath);
+
+      for (int rep = 0; rep < repetitions; rep++) {
+        results.add(readEachTile(randomCoordsToFetchPerRepetition, dbPath));
+      }
+      var totalStats = results.stream().mapToLong(ReadResult::totalDuration).summaryStatistics();
+      LOGGER.info("totalReadStats: {}", totalStats);
+
+      LongSummaryStatistics individualStats = results.stream().map(ReadResult::individualReadStats)
+        .collect(Collector.of(LongSummaryStatistics::new, LongSummaryStatistics::combine, (left, right) -> {
+          left.combine(right);
+          return left;
+        }));
+      LOGGER.info("individualReadStats:  {}", individualStats);
+
+      avgIndividualReadPerDb.put(dbPathStr, individualStats.getAverage());
+    }
+
+    List<String> keysSorted = avgIndividualReadPerDb.entrySet().stream()
+      .sorted((o1, o2) -> o1.getValue().compareTo(o2.getValue()))
+      .map(Map.Entry::getKey)
+      .toList();
+
+    LOGGER.info("diffs");
+    for (int i = 0; i < keysSorted.size() - 1; i++) {
+      for (int j = i + 1; j < keysSorted.size(); j++) {
+        String db0 = keysSorted.get(i);
+        double avg0 = avgIndividualReadPerDb.get(db0);
+        String db1 = keysSorted.get(j);
+        double avg1 = avgIndividualReadPerDb.get(db1);
+
+        double diff = avg1 * 100 / avg0 - 100;
+
+        LOGGER.info("\"{}\" vs \"{}\": avgs reads up by {}%", db0, db1, diff);
+      }
+    }
+  }
+
+  private static ReadResult readEachTile(List<TileCoord> coordsToFetch, Path dbPath) throws IOException {
+    LongSummaryStatistics individualFetchDurations = new LongSummaryStatistics();
+    try (var db = Mbtiles.newReadOnlyDatabase(dbPath)) {
+      db.getTile(0, 0, 0); // trigger prepared statement creation
+      var totalSw = Stopwatch.createStarted();
+      for (var coordToFetch : coordsToFetch) {
+        var sw = Stopwatch.createStarted();
+        if (db.getTile(coordToFetch) == null) {
+          throw new IllegalStateException("%s should exist in %s".formatted(coordToFetch, dbPath));
+        }
+        sw.stop();
+        individualFetchDurations.accept(sw.elapsed(TimeUnit.NANOSECONDS));
+      }
+      totalSw.stop();
+      return new ReadResult(totalSw.elapsed(TimeUnit.NANOSECONDS), individualFetchDurations);
+    }
+  }
+
+  private record ReadResult(long totalDuration, LongSummaryStatistics individualReadStats) {}
+
+
+}
--- a/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
+++ b/planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
@@ -4,11 +4,11 @@ import com.google.common.base.Stopwatch;
 import com.onthegomap.planetiler.config.Arguments;
 import com.onthegomap.planetiler.geo.TileCoord;
 import com.onthegomap.planetiler.mbtiles.Mbtiles;
-import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Duration;
-import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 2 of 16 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/MbtilesWriter.java
```diff
@@ -59,6 +65,18 @@ public class MbtilesWriter {
   private final FeatureGroup features;
   private final AtomicReference<TileCoord> lastTileWritten = new AtomicReference<>();
   private final MbtilesMetadata mbtilesMetadata;
+  private final AtomicLong tileDataIdGenerator = new AtomicLong(1);
+  //  private final LoadingCache<ByteBuffer, Long> tileDataIdByHash = CacheBuilder.newBuilder()
```
bbilger, line 69: de-duping over a shared cache would allow to reduce the size of Australia by another ~100MB but it adds some overhead and more importantly it would also make the output again less predictable
=> would like to get your input here

shermp:
Would using a faster non-cryptographic hashing algorithm improve performance much? Or is performance so limited by IO that it doesn't matter?

shermp:
I got around to testing this using `australia-oceania` as the region.

Here are the file size differences:
```
-rw-rw-r--  1 xxxx xxxx 4.9G May 10 14:56 compact.mbtiles
-rw-rw-r--  1 xxxx xxxx 2.3G May 10 14:56 compact.mbtiles.zst
-rw-rw-r--  1 xxxx xxxx  12G May 10 15:16 full.mbtiles
-rw-rw-r--  1 xxxx xxxx 2.4G May 10 15:16 full.mbtiles.zst
```
When uncompressed, compacted mbtiles are a LOT smaller. There isn't much difference when compressing the files with zstandard, which is probably to be expected. 

I tried serving the tiles with tileserver-gl, and while I have no performance figures, the compacted tiles do feel slightly slower than the full tileset. But by the time you add some tile caching into the mix, it may not matter much.

msbarry:
Thanks for this change @bbilger! This looks like a huge improvement

> Would using a faster non-cryptographic hashing algorithm improve performance much? Or is performance so limited by IO that it doesn't matter?

Have you tried samplig with visualvm to see how much of a hot spot the hash is?

I'm also looking into generating pmtiles output, and it uses the [FNV hash](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) to solve this problem.  I haven't checked for collisions but with a 64 bit hash over 200-300 million tiles it should theoretically be a 1 or 2 out of 1000 chance of a collision, bit it is very simple and fast:

```java
  private static final long FNV1_64_INIT = 0xcbf29ce484222325L;
  private static final long FNV1_PRIME_64 = 1099511628211L;
  public static long hash(byte[] data) {
    long hash = FNV1_64_INIT;
    for (byte datum : data) {
      hash ^= (datum & 0xff);
      hash *= FNV1_PRIME_64;
    }
    return hash;
  }
```

Maybe [@bdon](https://github.com/bdon) has a better sense of fnv hash's suitability for planet-scale tilesets?

shermp:
I guess at the scales we're talking about (millions of tiles), one would really want the chance of collisions to be so low, that you don't have to worry about checking for collisions, otherwise I think you'd have to do a binary compare to know for certain. The last thing we would want for a data tile to collide with an ocean tile and not get included...

I think you end up going with a faster hash algorithm, a 128 bit hash would probably be the safer option, and I imagine it should still be faster than sha1/md5.

bbilger:
@shermp 
I wrote "com.onthegomap.planetiler.mbtiles.Compare" to compare reading from equal mbtiles files (e.g. compact vs. non-compact) and to get some idea about the performance decrease. On my machine, reading from australia-compact is 10-15% slower than from australia-not-compact.

@msbarry / @shermp 
What I did so far was swapping SHA-1 by Murmur3 128 - and it had zero impact on the performance which made me believe this is not the bottleneck. But all I did was checking the stats produced by planetiler at the end, and haven't analyzed it any further but will try to do so now.

@msbarry 
re: 1.: The threshold sounds like a good idea. Do you want it to be configurable as CLI argument?
re 2.: Do I understand the idea correctly that you effectively want to hash twice - in case of a duplicate? First layer would hash into a small space and store the result to a BitSet, and the second layer would use a "real" hash function (e.g. FNV 32) and store the result in a HashMap (hash, dataId). Is the idea here (in case of no duplicate) that the first hashing would be even cheaper, and looking up from a BitSet is a bit faster than from a HashMap? Sorry, just want to double-check that I get the idea, since it adds some complexity and FNV-32 already looks pretty fast.

bbilger:
I tried to address your feedback in https://github.com/onthegomap/planetiler/pull/219/commits/0faa63aad43eb62d8cf3dcb4e86323abc9d544a3 => please let me know what you think.

Testing with Australia, I can barely notice any difference but it seems a tiny bit faster. As I said before, I don't think the hashing is the bottleneck and at least I cannot identify it as one in VisualVM but it's probably the right thing to do to not use SHA-1.

One reason I decided to use SHA-1 in the beginning was that I wanted to avoid any collisions. As such I share @shermp 's concerns that using FNV 32 might be a bit problematic since a collision would be rather bad. Maybe it would be better to use FNV 64, FNV 128, or Murmur3-128?

msbarry:
> re: 1.: The threshold sounds like a good idea. Do you want it to be configurable as CLI argument?

Yes, if we go with a size-based approach then configurable would be good, probably with a default of ~20kb.

> re 2.: Do I understand the idea correctly that you effectively want to hash twice - in case of a duplicate? First layer would hash into a small space and store the result to a BitSet, and the second layer would use a "real" hash function (e.g. FNV 32) and store the result in a HashMap (hash, dataId). Is the idea here (in case of no duplicate) that the first hashing would be even cheaper, and looking up from a BitSet is a bit faster than from a HashMap? Sorry, just want to double-check that I get the idea, since it adds some complexity and FNV-32 already looks pretty fast.

The downside to option 1 is that a repeated fill tile with lots of attributes could get above 20kb, so option 2 is an adaptive way to identify tiles we might have seen before. On second thought, we'd probably want to use a bloom filter for that, so it's probably simpler just to use a size threshold.

bdon:
Instead of using a size threshold, could we limit the deduplication only to tiles that have a single layer, or a single feature on a single layer? We know a priori that the deduplication is useful for ocean squares and land squares, which might have arbitrary tags that increase the total tile size. We would need the ability to introspect on the PBF content instead of treating it as a blob. Are there good counterexamples on why this won't work?

msbarry:
It is possible that we can pass through an "is fill" bit for each feature, then inspect easily if all features in a tile are fill polygons without deserializing - that's what I was planning for #168.

So maybe to start on this PR, we could limit the deduping to just tiles with less than some small number of features (1? 2? 5?) and do the more robust fill identification as part of #168 ?  There's a `getNumFeaturesToEmit` method to get the number of features in `TileFeatures` argument to the tile encoder without deserializing anything.

bbilger:
"NumFeaturesToEmit" seems like the right thing to do and I implemented it accordingly.
Duplicates in Australia where spread as follows per "NumFeaturesToEmit":
{1=13074826, 2=1359105, 3=397821, 4=13822, 5=73, 7=4, 8=2, 9=2, 18=2}
=> trying to de-duplicate/hash only if less than 5 "NumFeaturesToEmit"

...
gonna tackle packing tiles_shallow x,y,z into one int in the DB now next

shermp:
Ran some numbers today, looks like it's possible to encode tiles up to zoom level 29 in a 64 bit int. 

Considering that at each zoom level, each axis has 2<sup>zoom</sup> tiles, you can encode tiles such:

Bits 0-4 could be `zoom`
Bits 5-33 could be `y`
bits 34-62 could be `x`

Or any other similar scheme.

bbilger:
:+1: 
Someone please correct me if I am wrong but I think the max supported zoom level is 14. This - I think - allows to pack into a 32bit integer.
For "packing" x,y,z into on int, I am using 
[TileCoord#encode](https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java#L56)
and on the generated columns, I am currently experimenting with, I am basically using [TileCoord#decode](https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java#L41)

shermp:
Are you generating the x,y columns as part of the SQL view?

bbilger:
Well, yes but mostly no. Technically the view remains unchanged to what it is now for the compact mode:
```
CREATE VIEW tiles AS
select
  tiles_shallow.zoom_level as zoom_level,
  tiles_shallow.tile_column as tile_column,
  tiles_shallow.tile_row as tile_row,
  tiles_data.tile_data as tile_data
from tiles_shallow
join tiles_data on tiles_shallow.tile_data_id = tiles_data.tile_data_id
```

The key change is to have generated and most importantly indexed columns in `tiles_shallow`:
```
CREATE TABLE tiles_shallow (
  tile_id integer,
  tile_data_id integer,

  zoom_level integer as ( (tile_id >> 28) + 8 ),
  tile_column integer as ( (tile_id >> 14) & ((1 << 14) - 1) ),
  tile_row integer as ( (1 << zoom_level) - 1 - (((1 << zoom_level) - 1) - ((tile_id) & ((1 << 14) - 1))) )
)
CREATE UNIQUE INDEX tiles_shallow_index on tiles_shallow (zoom_level, tile_column, tile_row)
```

With a view-only approach like the following we would end up with a full table scan which would hurt read performance badly
```
CREATE VIEW tiles AS
select
  ( (tiles_shallow.tile_id >> 28) + 8 ) as zoom_level,
  ( (tiles_shallow.tile_id >> 14) & ((1 << 14) - 1) ) as tile_column,
   ( (1 << ( (tiles_shallow.tile_id >> 28) + 8 )) - 1 - (((1 << ( (tiles_shallow.tile_id >> 28) + 8 )) - 1) - ((tile_id) & ((1 << 14) - 1))) ) as tile_row,
  tiles_data.tile_data as tile_data
from tiles_shallow
join tiles_data on tiles_shallow.tile_data_id = tiles_data.tile_data_id
```

What I can say so far is that:
- write performance is mostly unaffected - maybe a slight improvement
- file size is slightly lower
haven't tested impact on read yet

msbarry:
> :+1: 
> 
> Someone please correct me if I am wrong but I think the max supported zoom level by planetiler is 14. This allows to pack it into a 32bit integer.
For now yes although it will increase to 15 or 16 at some point in the future.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.LongSummaryStatistics;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.LongSupplier;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BenchmarkMbtilesWriter {
```
msbarry, line 32: When I run with one repetition, 10m tiles, and without compact-db I see this output at the end of each stage:

compact db:
```
0:00:15 INF [mbtiles] - Finished in 13s cpu:33s avg:2.6
0:00:15 INF [mbtiles] -   read    1x(4% 0.5s wait:12s)
0:00:15 INF [mbtiles] -   encode 10x(13% 2s wait:10s)
0:00:15 INF [mbtiles] -   write   1x(97% 12s)
```

original:
```
0:00:34 INF [mbtiles] - Finished in 19s cpu:2m38s avg:8.2
0:00:34 INF [mbtiles] -   read    1x(2% 0.3s wait:18s)
0:00:34 INF [mbtiles] -   encode 10x(73% 14s sys:2s block:5s)
0:00:34 INF [mbtiles] -   write   1x(76% 15s wait:4s)
```

but the summary statistics show 19 seconds for all stages on the original, and 13 for all on compact - so they are measuring the longest phase but the main bottleneck we care about here is the write thread.  Would it be possible to have the benchmark isolate just the `write` phase by invoking the compact vs. non-compact batched writer directly? The tile `byte[]` payloads don't need to actually be real tiles.

Also I usually find it a little more straightforward to report the performance in "operations per second" to make it easier to compare across different input sizes.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.LongSummaryStatistics;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BenchmarkMbtilesRead {
```
msbarry, line 21: Overall this read benchmark is looking good, the main thing we care about is throughput of "read operations per second" - could we have the benchmark report that metric? I don't think the per-request stat hurts (unless getting the system time for each request adds overhead) but the main metric I'm interested in is overall read throughput.

msbarry (commented):
Some comments on the high-level benchmark and verification code. Planning to get down into the implementation details next...

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/MbtilesWriter.java
```diff
@@ -225,43 +233,131 @@ private void tileEncoderSink(Iterable<TileBatch> prev) throws IOException {
...
+
+    private final Supplier<Long> tileDataIdGenerator;
+    private final Map<Integer, Long> tileDataIdByHash = new HashMap<>((int) MAX_TILES_PER_BATCH);
+
+    private CompactTileFeaturesEncoder(Supplier<Long> tileDataIdGenerator) {
+      this.tileDataIdGenerator = tileDataIdGenerator;
+    }
+
+    @Override
+    public TileEncodingResult encode(TileFeatures tileFeatures) throws IOException {
```
msbarry, line 307: Can we move the tile cache to the single-threaded writer so that we deduplicate a smaller number of entries for the unique tiles?  The breakdown of work would roughly be:

`encode` (in workers)
- (only when using --compact-db) if num feature is < `MAX_FEATURES_HASHING_THRESHOLD` then compute the hash and pass it along on `TileEncodingResult` otherwise some value indicating "not repeated"
- use the old `hasSameContents ` logic to avoid encoding a tile if it's the same as the last tile - I think that should be more than sufficient to avoid unnecessary work since it's a cheap check and will likely avoid 99%+ of the repeat encoding work. We could also use the hasSameContents check to memoize the hash between identical tiles.

`tileWriter` / compact batch writer (single-threaded)
- keep a shared hppc `IntIntHashMap` (?) from tile hash to tile ID, and append new values to it when we encounter a repeated tile where the hash was computed in encode
- append to the `tiles_data` table except when you encounter a repeated tile where we've already added the hash to it.

bbilger: Oookay, so far I was under the assumption that the current tile memoization isn't that effective. Thus I tried to knock-off 2 things with the hashmap, de-duping for writing and avoiding to re-encode+gzip. 
I'll try to verify the effectiveness of the current memoization and if it's that effective, then I'll refactor accordingly.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
bbilger: Despite the name `BenchmarkMbtilesWriter`, I wanted to check both stages encode and write. And it's specifically designed to make the current memoization not work.
This is related to https://github.com/onthegomap/planetiler/pull/219#discussion_r871772440
=> if that's the case, then I'll adjust according

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
bbilger: You are absolutely right, per request stats probably hurts, and I also agree that "read/s" is the better metric. I'll adjust accordingly.

bbilger:
> > Someone please correct me if I am wrong but I think the max supported zoom level by planetiler is 14. This allows to pack it into a 32bit integer.
> 
> For now yes although it will increase to 15 or 16 at some point in the future.
> 
> > write performance is mostly unaffected - maybe a slight improvement
> 
> +1 I would expect some write performance increase to come from a simpler index structure, but most to come from being able to fit more tiles into each batch with the 999 limit on prepared statement params
> 
> Curious how read performance turns out. To query from application layer, it would be relatively efficient to map from x/y/z to tile ID, then lookup the data for it but I'm not sure how possible it is to express that in a sqlite view...

I cannot see a noticeable impact on read performance (BenchmarkMbtilesRead shows something like 1% faster).

The approach - let's call it approach no.1 -  presented here https://github.com/onthegomap/planetiler/pull/219#issuecomment-1124366496 relies on generated x,y,z columns that need to be indexed. Thus the index structure is/should be (mostly) identical.

Unless you want to build a sqlite file that breaks with mbtiles spec, approach no.1 is the only feasible option I can see. 
A full table-scan - let's call it approach no.2 - as described at the end of https://github.com/onthegomap/planetiler/pull/219#issuecomment-1124366496 is obviously a no-go. 

This bring us to approach no.3:

```
create view tiles as
  select zoom_level, tile_column, tile_row, tile_data
  from tiles_shallow ts
  join tiles_data on ts.tile_data_id = tiles_data.tile_data_id
  left join zoom_levels 
  left join tile_columns 
  left join tile_rows
  where ts.tile_id = ((zoom_level-8) << 28) | (tile_column << 14) | (tile_row);
```
(bear with me if the formula is not quite correct, haven't tested it much but it worked at least for my test coordinates)
zoom_levels, tile_columns, tile_rows are views on a number table (`create table numbers (number integer primary key)` and it contains numbers from 0 to 1 << 14). I just created it manually but while it's relatively quick (~5ms), it's definitely slower, and it breaks pretty quickly if you don't filter it completely with zoom_level, tile_column, tile_row.

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/MbtilesWriter.java
msbarry: I think the current memoization is pretty effective since the tiles are processed in order in consecutive batches, but definitely let me know if you find otherwise. I tested Australia on my m1 mac without memoization and encode time went from 140 seconds to 370 cpu seconds.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
msbarry: I think the options would either be to benchmark the encode/write separately, or try to extract the cpu time from each thread since they all end at the same time.  I'm mostly interested in the write thread performance since that is a single threaded bottleneck.

msbarry:
> > > Someone please correct me if I am wrong but I think the max supported zoom level by planetiler is 14. This allows to pack it into a 32bit integer.
> > 
> > 
> > For now yes although it will increase to 15 or 16 at some point in the future.
> > > write performance is mostly unaffected - maybe a slight improvement
> > 
> > 
> > +1 I would expect some write performance increase to come from a simpler index structure, but most to come from being able to fit more tiles into each batch with the 999 limit on prepared statement params
> > Curious how read performance turns out. To query from application layer, it would be relatively efficient to map from x/y/z to tile ID, then lookup the data for it but I'm not sure how possible it is to express that in a sqlite view...
> 
> I cannot see a noticeable impact on read performance (BenchmarkMbtilesRead shows something like 1% faster).

Cool, I would expect the fastest one to be the original, so mostly interested how much read qps drops with the more compact formats. If we can get a compact format with same read performance as original that would be ideal.

> 
> The approach - let's call it approach no.1 - presented here [#219 (comment)](https://github.com/onthegomap/planetiler/pull/219#issuecomment-1124366496) relies on generated x,y,z columns that need to be indexed. Thus the index structure is/should be (mostly) identical.

OK, is there any benefit of this one then? It seems the same as x/y/z columns, but with some added complexity?

> 
> Unless you want to build a sqlite file that breaks with mbtiles spec, approach no.1 is the only feasible option I can see. A full table-scan - let's call it approach no.2 - as described at the end of [#219 (comment)](https://github.com/onthegomap/planetiler/pull/219#issuecomment-1124366496) is obviously a no-go.
> 
> This bring us to approach no.3:
> 
> ```
> create view tiles as
>   select zoom_level, tile_column, tile_row, tile_data
>   from tiles_shallow ts
>   join tiles_data on ts.tile_data_id = tiles_data.tile_data_id
>   left join zoom_levels 
>   left join tile_columns 
>   left join tile_rows
>   where ts.tile_id = ((zoom_level-8) << 28) | (tile_column << 14) | (tile_row);
> ```
> 
> (bear with me if the formula is not quite correct, haven't tested it much but it worked at least for my test coordinates) zoom_levels, tile_columns, tile_rows are views on a number table (`create table numbers (number integer primary key)` and it contains numbers from 0 to 1 << 14). The number table is a trick to get hold of the requested parameters zoom_level, tile_column, tile_row to reconstruct the tile_id. I just created it manually but while it's relatively quick (~5ms), it's definitely slower, and it breaks pretty quickly if you don't filter it completely with zoom_level, tile_column, tile_row.

OK cool thanks for exploring this, but if it's looking like there might not a good way to express this through sqlite views then feel free to ignore!

msbarry:
Also @bbilger let me know if any of those need more clarification, or if you'd rather have me push the change for one to your branch!

bbilger:
@msbarry all is clear. Sorry, I just didn't have to work on it over the weekend. I should have addressed the majority of your comments absolutely latest by Friday.

bbilger:
> OK, is there any benefit of this one then? It seems the same as x/y/z columns, but with some added complexity?

The advantage is that write performance increases since the prepared statement will have just 1 vs 3 arguments. Let me re-write the write benchmark then a decision can be based on that.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
bbilger: addressed here https://github.com/onthegomap/planetiler/pull/219/commits/b8f58f2accc3dfad16d2dd5a86aaa8941c4c4473

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/MbtilesWriter.java
bbilger: addressed in https://github.com/onthegomap/planetiler/pull/219/commits/5414f33ddd27d0889c0efbd8d85ca0b1b24e16ee

I left a comment in code here, https://github.com/onthegomap/planetiler/pull/219/commits/5414f33ddd27d0889c0efbd8d85ca0b1b24e16ee#diff-38e31206bb20aac6140acd652533b5194401ffbf1d95ede4cb620d2294bc40a8R673

I think by trying to de-dupe memoized only we miss quite some opportunities and I think we should just try to de-dupe whenever there's a hash - but I understand if you want to avoid potential unnecessary map lookups

results on my machine
```

memoized, only
[m[0m0:06:13 INF - 	mbtiles          1m9s cpu:10m46s gc:4s avg:9.4
[m[0m0:06:13 INF - 	  read    1x(22% 15s sys:2s wait:46s done:3s)
[m[0m0:06:13 INF - 	  encode 12x(63% 43s wait:13s done:3s)
[m[0m0:06:13 INF - 	  write   1x(56% 38s sys:4s wait:21s)
[m[0m0:06:13 INF - ----------------------------------------
[m[0m0:06:13 INF - 	features	3.6GB
[m[0m0:06:13 INF - 	mbtiles	1.8GB

1814077440

all that have a hash
[m[0m0:06:11 INF - 	mbtiles          1m7s cpu:10m28s gc:4s avg:9.3
[m[0m0:06:11 INF - 	  read    1x(22% 15s sys:2s wait:46s done:3s)
[m[0m0:06:11 INF - 	  encode 12x(62% 42s wait:13s done:3s)
[m[0m0:06:11 INF - 	  write   1x(54% 37s sys:3s wait:22s)
[m[0m0:06:11 INF - ----------------------------------------
[m[0m0:06:11 INF - 	features	3.6GB
[m[0m0:06:11 INF - 	mbtiles	1.7GB

1725030400
```

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
bbilger: addressed in https://github.com/onthegomap/planetiler/pull/219/commits/44d481e74624329f27d86709e3aa86e09c95f35a

I tried to make the the test somewhat realistic... the "key" is the "dupe spread". If I am not mistaken, then the dupe spread for Australia is 0.17%

results are as following:
```
--bench-tiles-to-write=1000000 --bench-repetitions=10 --bench-no-dupe-tiles=10 --bench-distinct-tile-data-size=800 --bench-dupe-tile-data-size=100

  --bench-dupe-spread=10

    (non-compact)
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=5535826.018941, min=494772.881352, average=553582.601894, max=572869.469085}

    --compact-db
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=5893505.249478, min=516823.356896, average=589350.524948, max=614381.936503}

  --bench-dupe-spread=1

    (non-compact)
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=5515072.624620, min=500652.172801, average=551507.262462, max=572924.290469}

    --compact-db
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=6341896.356866, min=553440.630765, average=634189.635687, max=652745.072077}

  --bench-dupe-spread=50

    (non-compact)
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=5517730.613009, min=493311.037536, average=551773.061301, max=578336.895536}

    --compact-db
    tileWritesPerSecondsStats: DoubleSummaryStatistics{count=10, sum=4996218.743876, min=434994.065552, average=499621.874388, max=517147.994776}
```

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
bbilger: One thing here however that I cannot wrap my mind around is that after all the refactorings, reading from the compact-db actually seems to be faster. This cannot really be the case but I can't spot an issue in the benchmark.

```
0m0:00:11 INF - working on /.../australia-not-compact.mbtiles
0m0:00:54 INF - readOperationsPerSecondStats: DoubleSummaryStatistics{count=10, sum=1168952.601015, min=86929.648036, average=116895.260102, max=125684.624005}
0m0:00:54 INF - working on /.../australia-compact.mbtiles
0m0:01:35 INF - readOperationsPerSecondStats: DoubleSummaryStatistics{count=10, sum=1241796.798647, min=96633.936167, average=124179.679865, max=130121.770334}
0m0:01:35 INF - diffs
0m0:01:35 INF - "/.../australia-not-compact.mbtiles" vs "/.../australia-compact.mbtiles": avg read operations per second improved by 6.231578386342335%
```

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/MbtilesWriter.java
msbarry: sounds good! I think it's fine to always do the map lookup in the writer.

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Compare.java
```diff
@@ -0,0 +1,160 @@
...
+              ---
+              %s
+              >>>
+              """.formatted(
+              coord, featureCountMatches, features0.size(), features1.size(),
+              getDiffJoined(features0, features1, "\n"),
+              getDiffJoined(features1, features0, "\n"));
+
+            if (failOnFeatureDiff) {
+              throw new RuntimeException(msg);
```
msbarry, line 102: to fix sonar's complaint here and below, this could be a more specific exception like assertion error?

msbarry (commented):
Sorry for the delay, this looks great! Leaving a few minor comments, but I see no major issues remaining before we merge.  Thanks for putting this together and improving the code long the way!

planetiler-basemap/src/test/java/com/onthegomap/planetiler/basemap/util/VerifyMonacoTest.java
```diff
@@ -51,7 +52,7 @@ void testStilInvalidWithOneTile() throws IOException {
         VectorTile.encodeGeometry(point(0, 0)),
         Map.of()
       )));
-      writer.write(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()));
+      writer.write(new TileEncodingResult(TileCoord.ofXYZ(0, 0, 0), gzip(tile.encode()), false, null));
```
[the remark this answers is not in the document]
bbilger: Yes, sure. I am okay with it being re-licensed.

planetiler-core/src/test/java/com/onthegomap/planetiler/util/HashingTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+class HashingTest {
+
+  @ParameterizedTest
+  @ArgumentsSource(TestArgs.class)
+  void testFnv32(boolean expectSame, byte[] data0, byte[] data1) {
+    var hash0 = Hashing.fnv32(data0);
```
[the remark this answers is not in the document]
bbilger: hmm, I am a big fan of parameterized tests and the argument provider could have been re-used for other hash function but I rewrote it now - tho it's not so compact because of the required cast.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
bbilger: A "dupe spread" of 50 would mean that in the "dupe group" always two tiles would share the same data. At that level the compact mode has disadvantages because of the additional overhead (more inserts). 
The less spread dupes are (0 would mean all tiles in the "dupe group" would share the tile data), the more efficient the compact mode gets.

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesRead.java
bbilger: Benchmarking sqlite is a bit tricky it seems. I don't quite get it either but currently on my machine it seems to be faster reading from the DB in compact-db mode.

bbilger:
> Sorry for the delay, this looks great! Leaving a few minor comments, but I see no major issues remaining before we merge. Thanks for putting this together and improving the code long the way!

Absolutely no problem. Thank you for the review and the feedback. I tried to incorporate your feedback in this commit https://github.com/onthegomap/planetiler/pull/219/commits/c65c7b7d23007973e0d4cf8cab4ed673d91be3d9 . Let me know if there's anything else.

msbarry:
Ran some tests this morning on the planet with this branch merged with #225.  `--compact-db` appears to improve write throughput by 5%, hurt read throughput by 5%, and make the mbtiles file about 25% smaller.

`--compact-db=false`:

```
0:38:41 INF - 	mbtiles          14m13s cpu:5h45m29s gc:45s avg:24.3
0:38:41 INF - 	  read    2x(24% 3m22s sys:58s wait:8m20s done:28s)
0:38:41 INF - 	  merge   1x(35% 5m3s sys:13s wait:8m26s done:13s)
0:38:41 INF - 	  encode 62x(33% 4m38s sys:8s wait:8m53s done:13s)
0:38:41 INF - 	  write   1x(87% 12m25s sys:1m29s wait:55s)
0:38:41 INF - ----------------------------------------
0:38:41 INF - 	features	184GB
0:38:41 INF - 	mbtiles	104GB
```

`--compact-db=true`:

```
0:37:38 INF - 	mbtiles          13m29s cpu:5h50m42s gc:46s avg:26
0:37:38 INF - 	  read    2x(26% 3m29s sys:1m1s wait:7m18s done:25s)
0:37:38 INF - 	  merge   1x(32% 4m22s sys:14s wait:8m23s done:11s)
0:37:38 INF - 	  encode 62x(35% 4m44s sys:9s wait:8m2s done:11s)
0:37:38 INF - 	  write   1x(87% 11m40s sys:1m18s wait:53s)
0:37:38 INF - ----------------------------------------
0:37:38 INF - 	features	184GB
0:37:38 INF - 	mbtiles	84GB
```

The read benchmark shows about 67.7k reads/s with regular and 64.7k reads/s with compact (I was running short on time so did a smaller test)

```
0:00:00 DEB - argument: bench_repetitions=3 (number of repetitions)
0:00:00 DEB - argument: bench_nr_tile_reads=50000 (number of tiles to read)
0:00:00 DEB - argument: bench_pre_warms=1 (number of pre warm runs)
0:00:00 DEB - argument: bench_mbtiles=data/output.mbtiles,data/compact.mbtiles (the mbtiles file to read from)
0:00:36 INF - working on data/output.mbtiles
0:00:43 INF - readOperationsPerSecondStats: DoubleSummaryStatistics{count=3, sum=203141.152027, min=62994.850298, average=67713.717342, max=70282.853583}
0:00:43 INF - working on data/compact.mbtiles
0:00:51 INF - readOperationsPerSecondStats: DoubleSummaryStatistics{count=3, sum=193959.728756, min=63488.766430, average=64653.242919, max=65851.192021}
0:00:51 INF - diffs
0:00:51 INF - "data/compact.mbtiles" to "data/output.mbtiles": avg read operations per second improved by 4.73367504185164%
```

Previous tests before this change took about 12m +/- 10s to write mbtiles so I don't have any concerns about `--compact-db=false` performance being different from what's on main now.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Hashing.java
```diff
@@ -0,0 +1,23 @@
+package com.onthegomap.planetiler.util;
+
+public final class Hashing {
```
msbarry, line 3: Can you add a short javadoc to this class and the public members that explain what it's doing/how to use?

planetiler-benchmarks/src/main/java/com/onthegomap/planetiler/benchmarks/BenchmarkMbtilesWriter.java
msbarry: Sounds good, a real-world test on planet data showed about a 5% improvement going to compact, so 10 seems like a realistic default value to use :thumbsup:

msbarry (commented):
Looks good! 2 minor followup comments then it should be good to merge.

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
```diff
@@ -341,6 +343,22 @@ public TileCoord tileCoord() {
...
+    /**
+     * Generates a hash over the feature's relevant data: layer, geometry, and attributes. The coordinates are
+     * <b>not</b> part of the hash.
+     * <p>
+     * Used as an optimization to avoid writing the same (ocean) tiles over and over again.
+     */
+    public int generateContentHash() {
+      int hash = Hashing.FNV1_32_INIT;
+      for (var feature : entries) {
+        long layerId = extractLayerIdFromKey(feature.key());
```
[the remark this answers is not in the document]
bbilger: sure! sorry, got fooled by #hasSameContents in which the result of #extractLayerIdFromKey is assigned to a variable of type long

PR #226, opened 2022-05-17
Handle elevations in units besides meters

### LINKED ISSUE #224: [BUG] Elevations in feet are misinterpreted as meters
When parsing an `ele` tag expressed in feet and inches, Planetiler discards the unit and interprets any number at the beginning of the value as a measurement in meters for the `ele` field, then converts it to feet for the `ele_ft` field. As a result, both fields are inaccurate. Instead, it should recognize that a value of the form `432'1"` is expressed in feet and inches, one of the established [unit specifications](https://wiki.openstreetmap.org/wiki/Map_features/Units) for OSM tags and convert it to meters as appropriate. This will enable more accurate labeling of peaks and other features primarily in the United States, as described in [this proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/ele_with_units).

https://github.com/onthegomap/planetiler/blob/9c1472fd30e4c84c8fc5d8a69fa8fd1c1aa08543/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/layers/MountainPeak.java#L115-L120 https://github.com/onthegomap/planetiler/blob/9c1472fd30e4c84c8fc5d8a69fa8fd1c1aa08543/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/Utils.java#L63-L66 https://github.com/onthegomap/planetiler/blob/9c1472fd30e4c84c8fc5d8a69fa8fd1c1aa08543/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Parse.java#L14 https://github.com/onthegomap/planetiler/blob/9c1472fd30e4c84c8fc5d8a69fa8fd1c1aa08543/planetiler-basemap/src/main/java/com/onthegomap/planetiler/basemap/util/Utils.java#L52-L57

msbarry:
Fix #224 by changing elevation handling to accept units besides meters according to https://wiki.openstreetmap.org/wiki/Map_features/Units


planetiler-core/src/test/java/com/onthegomap/planetiler/util/ParseTest.java
```diff
@@ -52,6 +53,39 @@ void testDirection(String in, int out) {
...
+    "1mi,1609.344",
+    "1 mi,1609.344",
+    "328',99.974",
+    "328'11\",100.254",
+    "garbage,null",
+    "1nmi,1852",
+    "1.5 nmi,2778",
+    "1..5 nmi,null",
+    "36\",0.9144",
+    "1'11\",0.584",
```
1ec5, line 78: Another case to test would be decimal feet or inches, such as [`132.74'`](https://www.openstreetmap.org/node/9292162885) or `132'8.88"`.

1ec5 (commented):
Thanks for hopping on this issue so quickly – the proposal probably hinges on people feeling comfortable about data consumers handling these cases correctly. 👍

PR #238, opened 2022-05-29
Change terminology around "rendering" to be consistent with MapLibre vision

ZeLonewolf:
In maplibre/maplibre#27, there was a discussion regarding the term "rendering" with regard to the scope of the mapLibre project, and what was considered "rendering" versus not in an mvt technology stack.  There seemed to be a consensus, at least from @wipfli and @nyurik that "rendering" was really describing a technology that produced graphical pixels rather than data.  In order to align with mapLibre, we should change the terminology used in planetiler to align with those definitions.  This PR updates the project's documentation to rephrase usages of the term "render" to be consistent with that definition.

If we agree on drawing a bright line in terminology between rendering and vector tile generation, it would be appropriate also to rename usages of "render" in the code itself in order to avoid developer confusion down the line.  However, this PR does not include any code renames.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5,7 +5,7 @@
 Planetiler builds a map in 3 phases:
 
 1. [Process Input Files](#1-process-input-files) according to
-   the [Profile](planetiler-core/src/main/java/com/onthegomap/planetiler/Profile.java) and write rendered tile features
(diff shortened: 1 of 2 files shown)


nyurik:
@ZeLonewolf is the Mr. :zap:, thx for keeping things in sync :)

msbarry:
Looks good, thanks! Might be a bigger project to remove references to rendering from the code...

PR #244, opened 2022-05-31
fix: merge line if it needs to be simplified

farfromrefug:
This ensure we still simplify passed line to `mergeLineStrings` even if there is only one

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
@@ -91,7 +91,8 @@ public class FeatureMerge {
       // - only 1 element in the group
       // - it doesn't need to be clipped
       // - and it can't possibly be filtered out for being too short
-      if (groupedFeatures.size() == 1 && buffer == 0d && lengthLimit == 0) {
+      // - and it does not need to be simplified
+      if (groupedFeatures.size() == 1 && buffer == 0d && lengthLimit == 0 && tolerance == 0) {
         result.add(feature1);
       } else {
         LineMerger merger = new LineMerger();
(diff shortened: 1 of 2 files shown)


msbarry (commented):
Thanks! This is pretty subtle, would it be possible to add a test to verify so I retain this behavior when I refactor later?

farfromrefug:
@msbarry i will but only in a few days/week. sorry

msbarry:
Added a test for this, but I'm not 100% sure why it's needed - all features get simplified during the initial processing phase so if just 1 feature ends up in a group then it should already be simplified.  Are you simplifying with a smaller tolerance during initial processing, and larger during postprocessing?

No rush on an answer, I'll target this for next release so it's not holding anything up.

farfromrefug:
@msbarry maybe i understood wrong. But what i understand for when we use `mergeLineStrings` is that:
* we dont simplify features in "process"
* we pass the lines features to `mergeLineStrings` in "postProcess"

So if this is true we can get into `mergeLineStrings` where `groupedFeatures.size() == 1`. Without the test added here the line would then not be simplified. 

I think this is why you did the same for `lengthLimit` ?

msbarry:
You _can_ opt-out of simplification or min size limits in process, but in practice we still simplify, but only opt-out of the length limit if we're going to merge in postProcess. 

So this will only make a difference if you set simplify tolerance to 0 in process.

farfromrefug:
@msbarry yes this i agree. isn't it what we do for all layers where we merge lines ?

msbarry:
In the openmaptiles config, it sets the min length to 0 for all line layers that get merged in postProcess, but process also simplifies them like normal so this code would have no effect on the output, just add processing time because we're re-simplifying lines that are already simplified.

If there was a use-case for setting the simplification tolerance to 0 in process then this might make sense but I'm not seeing that use-case...

farfromrefug:
@msbarry oh if that is so yes no need. Could we see it as a safeguard? Otherwise simply close this PR!

msbarry:
Added a `resimplify` parameter you can set to true to opt-into this behavior (disabled by default). How does that look?

farfromrefug:
@msbarry that seems good to me ! thanks

PR #250, opened 2022-06-02
place layer changes from OMT

mmc1718:
- Copied changes from https://github.com/ACMEAtronOmatic/openmaptiles/blob/master/layers/place/city.sql


mmc1718:
opened by mistake

Issue #133, opened 2022-03-12
[FEATURE] generate low details tiles for whole planet when generating detailed tiles for certain region


caspg:
**First of all, huge thanks for building this great application!**

---

**Is your feature request related to a problem? Please describe.**

I've generated tiles for Europe and it works great. I wonder if there would be an option to create tiles (country contours, rivers, etc) for the whole planet and the detailed tiles for the desired region?

<img width="1675" alt="Screenshot 2022-03-12 at 12 50 00" src="https://user-images.githubusercontent.com/10419439/158016881-4288c704-a911-42d1-9f57-6cc9209a0ecd.png">

msbarry:
Hello! That would probably be doable for some sources, for example natural earth and water polygons but any low zoom features rendered from OSM won't show up. 

I think this would require a new config attribute, and a logic tweak in SimpleReader to not filter geometries when reading, and also a tweak to how TileExtents computes the tile x/y range limits by zoom level when that attribute is specified. 

I'm working on a few [other tasks first](https://github.com/onthegomap/planetiler/projects/2) so it might be some time before I could take a look at this, but I'd be happy to review a PR. 

As a workaround you can run with `--bounds=planet` to generate the whole planet for all zoom levels but only OSM details from your extract. Or to save space, run planetiler a second time with `--bounds=planet --maxzoom=5` and combine the two mbtiles with a tool like tilelive-copy.

caspg:
@msbarry Thanks for the reply.  I'm afraid I won't be able to help with development since I know nothing about Java 😬 

I will try to play with suggested workarounds though.

caspg:
Merging two mbtiles using `tilelive-copy` gave me broken results (empty tiles) but it looked ok using `tile-join` (from https://github.com/mapbox/tippecanoe). It saves space but tileserver is returning empty tiles after zoom level 5.

Generating tiles with `--bounds=planet` gives better results IMO. It allows moving around the map and zoom everywhere. Of course, detailed tiles are returned only for the specific region. Low details tiles for the whole planet add around 25GB to the `output.mbtiles`.

msbarry:
Responded in slack - `tilelive-copy` is probably messing with the metadata. You may need to manually ensure that minzoom=0, maxzoom=14, bounds=-180,-90,180,90 when you combine using tilelive-copy

makitsune:
Thank you so much for this awesome program. I'm so happy it exists and I'd love to give support if I could.

[...]

PR #272, opened 2022-06-09
feat: added `maxzoomForRendering` and `minzoomForRendering`

farfromrefug:
This will allow rendering with `—maxzoom=6` while having zoom 6 rendered the same way as if `—maxzoom=14`.

`maxzoomForRendering` will be used for simplifly algorithms. The default is `MAX_ZOOM`(14) so by default whatever `maxzoom` is used you will get the same render as if `maxzoom=14`


wipfli:
Interesting that no tests break. Maybe you want to add some for the new behavior? Also, maybe this change could be somewhere reflected in the documentation?

farfromrefug:
@wipfli actually it is not normal for no test to break as the rendering is exactly the same as with maxzoom 14. About the doc yes. I am waiting on @msbarry to see first if he wants some refactoring

planetiler-core/src/main/java/com/onthegomap/planetiler/config/PlanetilerConfig.java
```diff
@@ -99,6 +101,8 @@ public static PlanetilerConfig from(Arguments arguments) {
       arguments.getDuration("loginterval", "time between logs", "10s"),
       arguments.getInteger("minzoom", "minimum zoom level", MIN_MINZOOM),
       arguments.getInteger("maxzoom", "maximum zoom level (limit 14)", MAX_MAXZOOM),
+      arguments.getInteger("minzoomForRendering", "minimum zoom level", MIN_MINZOOM),
+      arguments.getInteger("maxzoomForRendering", "maximum rendering zoom level (limit 14)", MAX_MAXZOOM),
```
msbarry, line 105: For naming, I think `maxzoom_for_render` or `render_maxzoom` (let's stick with lower_camel_case) makes sense, but maybe add to the description that minzoom/maxzoom are "minimum/maximum tile zoom level to generate" and something in the render max zoom that it's the zoom level that contains fine details for overzooming?

On a side note I almost wonder if it is more a property of the profile? Maybe we can revisit as we get deeper into config-driven schemas.

planetiler-core/src/main/java/com/onthegomap/planetiler/config/PlanetilerConfig.java
```diff
@@ -99,6 +101,8 @@ public static PlanetilerConfig from(Arguments arguments) {
       arguments.getDuration("loginterval", "time between logs", "10s"),
       arguments.getInteger("minzoom", "minimum zoom level", MIN_MINZOOM),
       arguments.getInteger("maxzoom", "maximum zoom level (limit 14)", MAX_MAXZOOM),
+      arguments.getInteger("minzoomForRendering", "minimum zoom level", MIN_MINZOOM),
```
msbarry, line 104: I don't think min zoom  for rendering is really necessary - I see it's only used in CacheByZoom which just uses the thresholds to set a size for the value array ahead of time - so just using min zoom would be fine.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CacheByZoom.java
```diff
@@ -29,7 +29,7 @@ private CacheByZoom(int minzoom, int maxzoom, IntFunction<T> supplier) {
    * @return a cache for {@code supplier} by zom
    */
   public static <T> CacheByZoom<T> create(PlanetilerConfig config, IntFunction<T> supplier) {
-    return new CacheByZoom<>(config.minzoom(), config.maxzoom(), supplier);
+    return new CacheByZoom<>(config.minzoomForRendering(), config.maxzoomForRendering(), supplier);
```
msbarry, line 32: these zoom levels don't matter too much - they just define the size of the object array that stores cached values, you can just leave with minzoom/maxzoom so that minzoomForRendering isn't necessary.

msbarry (commented):
Looks good! A couple of small comments

msbarry:
> Interesting that no tests break. Maybe you want to add some for the new behavior? Also, maybe this change could be somewhere reflected in the documentation?

Good point, 1 new test in PlanetilerTest might make sense similar to the compact DB tests - basically run planetiler with `--maxzoom=14` and `--maxzoom=8` and `--render_maxzoom=8 --maxzoom=8` and compare the number of features in some tile at z8 to make sure it's the same for first too, and larger for last one.

That kind of smoke test also ensures we don't throw an exception when someone uses this option.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CacheByZoom.java
farfromrefug: @msbarry actually it fails without this. I end up with an inifinite amount of error about out of bounds (sorry dont have the error here).

farfromrefug:
@msbarry i am looking at writing the test. I get the idea of what you want to do but to be honest i am a bit lost on the features to add. what we need for the test is 2 have one feature only visible at zoom 8 and one visible at zoom 14. 
That way:
* -`-render_maxzoom=8 --maxzoom=8` should show on the zoom 8  feature at zoom 8 
* `--maxzoom=14` : same thing
* `--maxzoom=8` is the failure as it will show both features.

The tricky part is that i cant do it with `zoomRange` it has to be "filtered" by minPixelSize. And that i am a bit lost with.
i can push the skeleton of the test if you want but i did not go much further :s really sorry

msbarry:
> @msbarry i am looking at writing the test. I get the idea of what you want to do but to be honest i am a bit lost on the features to add. what we need for the test is 2 have one feature only visible at zoom 8 and one visible at zoom 14. That way:
> 
> * -`-render_maxzoom=8 --maxzoom=8` should show on the zoom 8  feature at zoom 8
> * `--maxzoom=14` : same thing
> * `--maxzoom=8` is the failure as it will show both features.
> 
> The tricky part is that i cant do it with `zoomRange` it has to be "filtered" by minPixelSize. And that i am a bit lost with. i can push the skeleton of the test if you want but i did not go much further :s really sorry

No problem, I'll take a stab at adding one this week.

msbarry:
No problem, I'll take a look at adding a test for this.

msbarry:
Looks good, thanks for adding!

msbarry:
I think this logic is correct:

https://github.com/onthegomap/planetiler/blob/76c7880d6b4c5535157582416d8dedf648ed3f70/planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java#L1789-L1792

We don't want --maxzoom=8 to cause small features to show up in the z8 unless you explicitly set --render-maxzoom=8

farfromrefug:
@msbarry I think you got this wrong (or this PR is not doing what i think it is). The point of `render_maxzoom` is to have zoom 8 tiles to be exactly like they are when you render with `maxzoom=14` so the assert for `baseResult` should be the same as `renderMaxzoomResult` and should be false because the feature will be removed because of  simplification (in both cases).
while with `maxzoomResult` (maxzoom=8) the feature is not simplified at zoom 8 because 8 is the maxzoom and thus all simplifications are disabled.

msbarry:
I was interpreting `render_maxzoom` as the zoom level past which clients will overzoom, so it should set a lower simplification/size threshold at that zoom level.  This makes `minzoom`/`maxzoom` only control which tiles get generated and not affect the contents of those tiles which is likely what most users want.

The test verifies that the contents of z8 tiles is the same with maxzoom=8 or maxzoom=14, but only changes if you set render_maxzoom=8 (which clients would very rarely do).

farfromrefug:
@msbarry damin Mike i am sorry you are 100% right, it is me who got the test mixed up in my head  :( i am so so so sorry. Indeed in my case i use render_maxzoom=14 to get what i described before. But your test uses 8.
What i describe would be a test with `render_maxzoom=14` 
Again so so sorry for wasting your time ...

zokipirlo:
Great feature. I have a few questions, because I'm not really sure if understand this correctly.

1. Should this work like [Overzooming in tilelive-vector](https://github.com/mapbox/tilelive-vector)? If I generate mbtile to zoom level 8, currently I get on level 9-14 empty tiles and everything is rendered white (because they doesn't exists).
With overzooming it generate tiles with same details as of max zoom tile, so if I put render_maxzoom to 8 and maxzoom to 14, it will generate all tiles to z 14, but simplified, with details as of z 8 right?

2. I get a lot of errors ("Error processing OSM Way ..." and "Error processing OSM Relation ... java.lang.ArrayIndexOutOfBoundsException") if I set render_maxzoom to 8. Is that normal? (tested on currently latest commit in main branch)

msbarry:
> Great feature. I have a few questions, because I'm not really sure if understand this correctly.
> 
> 
> 
> 1. Should this work like [Overzooming in tilelive-vector](https://github.com/mapbox/tilelive-vector)? If I generate mbtile to zoom level 8, currently I get on level 9-14 empty tiles and everything is rendered white (because they doesn't exists).
> 
> With overzooming it generate tiles with same details as of max zoom tile, so if I put render_maxzoom to 8 and maxzoom to 14, it will generate all tiles to z 14, but simplified, with details as of z 8 right?
> 
> 
> 
> 2. I get a lot of errors ("Error processing OSM Way ..." and "Error processing OSM Relation ... java.lang.ArrayIndexOutOfBoundsException") if I set render_maxzoom to 8. Is that normal? (tested on currently latest commit in main branch)

Hello! Overzooming only happens in the client. Planetiler assists the client by including more detail at the max zoom level for rendering that it generates. This PR made it so that setting max zoom lower just omits zoom levels above that, without including more detail at the max zoom level. If you actually do want more detail at that max zoom level, then use the max zoom level for rendering flag instead. 

Setting max zoom level for rendering less than max zoom results in undefined, I'll try adding an error message at startup that explains a little better.

zokipirlo:
Aha, now I get it, thanks. Client (at least mine) will overzoom if you say that generated map is to zoom 8 (source in style.json) and you want max zoom 14 (for example in mapbox map). But when you have merged map with some parts (planet) to level 8 and some region (country) to 14 will:
- when set as maxzoom 8 in source inside style.json: overzoom everything even if tiles are generated higher than 8
- when set as maxzoom 14 in source inside style.json: empty tiles for zoom level above 8 if zoomed outside generated region
 
I thought [it's solving that problem](https://github.com/onthegomap/planetiler/issues/133). My mistake, sorry.

PR #275, opened 2022-06-19
Allow more than 256 attribute keys

ZeLonewolf:
Currently, planetiler is limited to 256 total attribute keys due to this being packed in a byte.  This PR changes the attribute key index to an integer, which increases this limit to 2 billion.  Since there's only 83,000 unique keys in the OSM database according to taginfo, this should effectively uncap the attribute limit.

This PR retains the byte limit for layers, because only a madman could possibly want more than 256 layers (famous last words).

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
@@ -40,14 +40,18 @@ public class CommonStringEncoder {
   public int encode(String string) {
     // optimization to avoid more expensive computeIfAbsent call for the majority case when concurrent hash map already
     // contains the value.
-    return stringToId.computeIfAbsent(string, s -> {
-      int id = stringId.getAndIncrement();
-      if (id >= MAX_STRINGS) {
-        throw new IllegalArgumentException("Too many strings");
-      }
-      idToString[id] = string;
-      return id;
-    });
+    Integer result = stringToId.get(string);
+    if (result == null) {
+      result = stringToId.computeIfAbsent(string, s -> {
+        int id = stringId.getAndIncrement();
+        if (id >= MAX_STRINGS) {
+          throw new IllegalArgumentException("Too many strings");
+        }
+        idToString[id] = string;
+        return id;
+      });
+    }
+    return result;
   }
 
   /**
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
@@ -11,78 +11,60 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public class CommonStringEncoder {
 
-  private final Map<String, Byte> stringToId = new ConcurrentHashMap<>(255);
-  private final String[] idToString = new String[255];
-  private final AtomicInteger layerId = new AtomicInteger(0);
+  private static final int MAX_STRINGS = 100_000;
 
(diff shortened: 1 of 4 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
```diff
@@ -1,41 +1,58 @@
...
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * A utility for compressing up to 250 commonly-used strings (i.e. layer name, tag attributes) into a single byte.
+ * A utility for compressing commonly-used strings (i.e. layer name, tag attributes).
  */
 @ThreadSafe
 public class CommonStringEncoder {
```
msbarry, line 12: Let's not have this class store two separate encodings internally - we construct a separate instance for layers and attributes anyway.  Can we just make this one support large values, but maybe add a nested class for bytes, something like:

```java
public static class AsByte {
  private final CommonStringEncoder = new CommonStringEncoder();
  public String decode(byte id) { ... }
  public byte encode(String string) { range check then encode()... }
}
```

?

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
```diff
@@ -1,41 +1,58 @@
...
 
-  private final ConcurrentMap<String, Byte> stringToId = new ConcurrentHashMap<>(255);
+  private final Map<String, Byte> stringToId = new ConcurrentHashMap<>(255);
   private final String[] idToString = new String[255];
   private final AtomicInteger layerId = new AtomicInteger(0);
 
+  private final Map<String, Integer> stringToIdMap = new ConcurrentHashMap<>(255);
+  private final Map<Integer, String> idToStringMap = new ConcurrentHashMap<>();
```
msbarry, line 19: Actually if we just raise the limit to 100k, then a `String[]` would only take up 800kb of memory - let's keep the array for idToString instead of concurrent hash map.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
```diff
@@ -51,4 +68,21 @@ public byte encode(String string) {
...
+   * Returns a int value to each unique string passed in.
+   *
+   * @param string the string to store
+   * @return an int that can be converted back to a string by {@link #decodeInt(int)}.
+   * @throws IllegalArgumentException if called for too many values
+   */
+  public int encodeInt(String string) {
+    // optimization to avoid more expensive computeIfAbsent call for the majority case when concurrent hash map already
+    // contains the value.
+    return stringToIdMap.computeIfAbsent(string, s -> {
```
msbarry, line 82: make sure you don't remove the `stringToIdMap.get(string) == null` optimization here - this is called billions of times for the planet, ConcurrentHashMap is optimized for reads, and all but 100-100k are read-only.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CommonStringEncoder.java
```diff
@@ -1,54 +1,74 @@
...
+   */
+  public static class AsByte {
+    private final CommonStringEncoder encoder = new CommonStringEncoder();
+
+    public String decode(byte id) {
+      return encoder.decode(id & 0xff);
+    }
+
+    public byte encode(String string) {
+      if (encoder.stringId.get() > 255) {
```
msbarry, line 68: If we add a constructor parameter to CommonStringEncoder with the max number of allowed strings (could default to 100k), then we could get rid of this check here.

msbarry (approved):
Looks good! Added one test

PR #278, opened 2022-06-20
OSM QA Tiles Example Profile

msbarry:
Now that we can support an arbitrary number of unique attributes (#275), add an example profile that generates something similar to [OSM QA tiles](https://osmlab.github.io/osm-qa-tiles/).


msbarry:
> I've been exploring something like this in https://github.com/ZeLonewolf/planetiler-osm-crudo
> 
> If it's of general interest, I'd be happy to contribute it back to the main repo. 

Interesting, I think they serve a slightly different purpose though, osm qa tiles appear to be a distribution of raw osm data as tiles at a single zoom level with no filtering, simplification, or grouping into layers.  

I think there's also a use case for a "raw" schema with some filtering/grouping applied based off of the tagging conventions on the osm wiki or common usage patterns from taginfo. The info on the wiki is also available in a [structured format](https://wiki.openstreetmap.org/wiki/Item:Q4822), although I'm not exactly sure how you'd know which data to filter out at lower zoom levels...

PR #280, opened 2022-06-20
feat: `--polygon` argument to constrain mbtiles to a poly shape

farfromrefug:
This PR add support for a new argument `poly_shape`. You can use it with the path of a [Osmosis poly file](https://wiki.openstreetmap.org/wiki/Osmosis/Polygon_Filter_File_Format)
When passed only tiles intersecting that poly "shape" will be generated.

As a example for France using [that](https://download.geofabrik.de/europe/france.poly) file the generated MBTiles went from 2.69Gb to 2.68Gb so no big difference there. Build time got from ~18min to ~20min on my computer. So it is a bit slower but it is expected.

The idea behind that PR is mostly to allow to generate small regions from "bigger" pbf files. Consequently you would end up with a mbtiles without "half filled" tiles.

NOTE: for now we dont parse "holes" from `.poly` files. I do not fully understand how they work (to which polygon they are removed from) + i dont see a real usecase for now.

I still need to add tests for this. I am not so good with tests so it will take me a bit of time ;)


farfromrefug:
@msbarry i ran some tests with that PR.
I built France using europe pbf file: 3.1Gb while it was 5Gb without that PR. 
Now it is not 100% finished and it should go even smaller. I dont know why but there are still some features outside the shape being rendered, seems only points. Is there another place i should add the `shape test` except [here](https://github.com/onthegomap/planetiler/blob/a61fdc5b6423567d03d1ecd19a2c70f14ef7fb19/planetiler-core/src/main/java/com/onthegomap/planetiler/render/TiledGeometry.java#L454=) ? 
Screenshot 2022-06-21 at 09 19 32

msbarry:
Nice @farfromrefug ! For points, you also need to add the test to this method:

https://github.com/onthegomap/planetiler/blob/a61fdc5b6423567d03d1ecd19a2c70f14ef7fb19/planetiler-core/src/main/java/com/onthegomap/planetiler/render/TiledGeometry.java#L110-L131

planetiler-core/src/main/java/com/onthegomap/planetiler/render/TiledGeometry.java
```diff
@@ -121,6 +122,12 @@ private void slicePoint(Coordinate coord) {
       if (extents.testX(wrappedX)) {
         for (int y = minY; y <= maxY; y++) {
           TileCoord tile = TileCoord.ofXYZ(wrappedX, y, z);
+          if (shape != null) {
+            TileCoord tileID = TileCoord.ofXYZ(x, y, z);
```
msbarry, line 126: I think we want to use wrappedX here

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PolyFileReader.java
```diff
@@ -0,0 +1,82 @@
...
+
+    File file = new File(filePath);
+    try (
+      BufferedReader br =
+        new BufferedReader(new FileReader(file))
+    ) {
+      boolean inRing = false;
+      String line;
+      List<Polygon> polygons = new ArrayList<>();
+      List<Coordinate> currentRing = null;
```
msbarry, line 38: FYI I made a `MutableCoordinateSequence` if you wanted to use that, just call `add(x, y)` then you can pass it into createLinearRing directly.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PolyFileReader.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.CoordinateXY;
+import org.locationtech.jts.geom.MultiPolygon;
+import org.locationtech.jts.geom.Polygon;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PolyFileReader {
```
msbarry, line 18: Can you add a quick blurb to the javadoc here, and maybe an `@see` link to the spec for it?

msbarry, line 18: Can you add a unit test for this class? If you wanted to keep the tests compact and not require many extra files, you could extract the logic that parses a multipolygon from a `BufferedReader` and use that directly from the test. Then you could inline the test files as multiline strings and wrap with a StringReader.

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -515,6 +517,13 @@ public void run() throws Exception {
       return; // exit only if just fetching wikidata or downloading sources
     }
 
+    String osmPolyFilePath =
+      arguments.getString("poly_shape", "a .poly file defining the \"bounds\" to generate", null);
```
msbarry, line 521: Let's add one integration test in `PlanetilerTests` that uses this option and ensures it's doing what we expect.  We want to make sure that 1) the options works without throwing exceptions and 2) points, lines, polygons, and polygon fills don't show up outside the bounding shape. If the integration test gets too complex verifying the cases for 2 then you could add some unit tests to `FeatureRendererTest` to verify the different point, line, polygon, and polygon fill variations.

planetiler-core/src/main/java/com/onthegomap/planetiler/config/Bounds.java
```diff
@@ -52,11 +55,17 @@ public Bounds addFallbackProvider(Provider latLonProvider) {
     return this;
   }
 
+  public Bounds setShape(MultiPolygon shape) {
+    this.shape = shape;
+    set(shape.getEnvelopeInternal());
```
msbarry, line 60: Would it ever make sense to set a shape and bounding box? If so would we take the union? intersection?

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileExtents.java
```diff
@@ -52,9 +59,9 @@ public boolean test(TileCoord tileCoord) {
 
   /**
    * X/Y extents within a given zoom level. {@code minX} and {@code minY} are inclusive and {@code maxX} and {@code
-   * maxY} are exclusive.
+   * maxY} are exclusive. shape is an optional polygon defining a more refine shape
    */
-  public record ForZoom(int minX, int minY, int maxX, int maxY) {
+  public record ForZoom(int minX, int minY, int maxX, int maxY, MultiPolygon shape) {
```
msbarry, line 64: One contributing factor to making pass2 slow is that we're projecting tile coordinates back to lat/lon, and also doing an expensive "intersects" test on a raw JTS geometry.  In order to speed this up:

1) we could project the shape to world web mercator coordinates (and possibly create a separate instance scaled per zoom level to go into `ForZoom`)
2) we could construct a `PreparedGeometry` ahead of time that's optimized for queries like `intersects`

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java
```diff
@@ -118,6 +122,17 @@ public Coordinate getLatLon() {
...
+  /** Returns the latitude/longitude of the northwest corner of this tile. */
+  public Geometry getEnvelope() {
+    double worldWidthAtZoom = Math.pow(2, z);
+    return JTS_FACTORY.toGeometry(new Envelope(
+      GeoUtils.getWorldLon(x / worldWidthAtZoom),
+      GeoUtils.getWorldLon((x + 1) / worldWidthAtZoom),
+      GeoUtils.getWorldLat(y / worldWidthAtZoom),
+      GeoUtils.getWorldLat((y + 1) / worldWidthAtZoom)
+    ));
+  }
```
msbarry, line 134: this will need to change based on re-projecting the geometry we test against, but let's make sure to add 1 test to TileCoordTests to make sure it does what we expect

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileExtents.java
msbarry, line 64: also let's make sure we update the tests in `TileExtentsTests` to verify the logic once it's settled

msbarry (commented):
Looking good, made some suggestions for cleanup/performance. We also want to add an integration test in PlanetilerTests and unit tests for the affected classes.

Might be easiest to start with the integration tests since they should remain stable regardless of implementation, then make the changes and when things solidify add unit tests to make sure the corner cases are covered?

planetiler-core/src/main/java/com/onthegomap/planetiler/render/TiledGeometry.java
```diff
@@ -121,6 +122,12 @@ private void slicePoint(Coordinate coord) {
       if (extents.testX(wrappedX)) {
         for (int y = minY; y <= maxY; y++) {
           TileCoord tile = TileCoord.ofXYZ(wrappedX, y, z);
+          if (shape != null) {
```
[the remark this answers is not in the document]
farfromrefug: @msbarry the thing is `extents` is actually a `ForZoom` which has no notion of `z` . Should i add `z` to `ForZoom` and add the `test(x, y, z)` there?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PolyFileReader.java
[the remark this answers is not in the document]
farfromrefug: @msbarry no i wrote it from scratch and actually did not find the `PolygonFileReader.java` :D
Do you want to switch to it? Not sure how to handle holes with jts though.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileExtents.java
farfromrefug: @msbarry not sure i get the necessary test here. I added the `shape` to `ForZoom` just so that i can get it in `TiledGeometry`

msbarry: This would be after we push the test(x,y,z) logic that compares to the shape into TileExtents

msbarry: You should be able to use `GeoUtils.latLonToWorldCoords`

farfromrefug:
@msbarry i just pushed the changes for prepared / simplified shape per zoom using tile coordinates.
France now builds in 1h!
```
1:03:26 DEB [mbtiles] - Tile stats:
1:03:26 DEB [mbtiles] - z0 avg:4.7k max:4.7k total:4.7k
1:03:26 DEB [mbtiles] - z1 avg:3.6k max:3.8k total:7.3k
1:03:26 DEB [mbtiles] - z2 avg:4.7k max:4.8k total:9.5k
1:03:26 DEB [mbtiles] - z3 avg:6.2k max:7.5k total:12k
1:03:26 DEB [mbtiles] - z4 avg:145k max:228k total:290k
1:03:26 DEB [mbtiles] - z5 avg:138k max:184k total:552k
1:03:26 DEB [mbtiles] - z6 avg:81k max:156k total:817k
1:03:26 DEB [mbtiles] - z7 avg:67k max:137k total:1.9M
1:03:26 DEB [mbtiles] - z8 avg:38k max:96k total:3.6M
1:03:26 DEB [mbtiles] - z9 avg:39k max:182k total:13M
1:03:26 DEB [mbtiles] - z10 avg:28k max:133k total:35M
1:03:26 DEB [mbtiles] - z11 avg:20k max:72k total:100M
1:03:26 DEB [mbtiles] - z12 avg:12k max:63k total:242M
1:03:26 DEB [mbtiles] - z13 avg:5.8k max:67k total:443M
1:03:26 DEB [mbtiles] - z14 avg:9.2k max:655k total:2.8G
1:03:26 DEB [mbtiles] - all avg:9k max:0
1:03:26 DEB [mbtiles] -  # features: 142,115,234
1:03:26 DEB [mbtiles] -     # tiles: 405,597
1:03:26 INF [mbtiles] - Finished in 2m28s cpu:16m21s gc:18s avg:6.6
1:03:26 INF [mbtiles] -   read    1x(13% 19s sys:4s wait:1m49s done:4s)
1:03:26 INF [mbtiles] -   encode 15x(37% 55s sys:7s)
1:03:26 INF [mbtiles] -   write   1x(5% 8s sys:5s wait:2m17s)
1:03:27 INF - Finished in 1h3m27s cpu:7h11m13s gc:1m17s avg:6.8
1:03:27 INF - FINISHED!
1:03:27 INF - 
1:03:27 INF - ----------------------------------------
1:03:27 INF -   overall          1h3m27s cpu:7h11m13s gc:1m17s avg:6.8
1:03:27 INF -   lake_centerlines 1s cpu:3s avg:2.5
1:03:27 INF -     read     1x(81% 1s)
1:03:27 INF -     process 15x(2% 0s wait:1s)
1:03:27 INF -     write    1x(0% 0s wait:1s)
1:03:27 INF -   water_polygons   10s cpu:27s avg:2.7
1:03:27 INF -     read     1x(93% 10s)
1:03:27 INF -     process 15x(4% 0.4s wait:10s)
1:03:27 INF -     write    1x(0% 0s wait:10s)
1:03:27 INF -   natural_earth    6s cpu:12s avg:1.8
1:03:27 INF -     read     1x(92% 6s)
1:03:27 INF -     process 15x(3% 0.2s wait:6s)
1:03:27 INF -     write    1x(0% 0s wait:6s)
1:03:27 INF -   osm_pass1        4m13s cpu:21m33s gc:5s avg:5.1
1:03:27 INF -     read     1x(9% 23s sys:15s wait:3m49s)
1:03:27 INF -     parse   14x(29% 1m12s wait:2m52s)
1:03:27 INF -     process  1x(86% 3m37s sys:15s wait:2s)
1:03:27 INF -   osm_pass2        55m38s cpu:6h29m5s gc:50s avg:7
1:03:27 INF -     read     1x(1% 22s sys:15s wait:51m42s done:3m30s)
1:03:27 INF -     process 15x(46% 25m28s sys:1m36s block:24s wait:3m13s done:2s)
1:03:27 INF -     write    1x(1% 36s sys:10s wait:54m33s)
1:03:27 INF -   boundaries       3s cpu:4s avg:1.4
1:03:27 INF -   sort             34s cpu:3m15s gc:5s avg:5.7
1:03:27 INF -     worker  5x(26% 9s sys:2s wait:4s done:5s)
1:03:27 INF -   mbtiles          2m28s cpu:16m21s gc:18s avg:6.6
1:03:27 INF -     read    1x(13% 19s sys:4s wait:1m49s done:4s)
1:03:27 INF -     encode 15x(37% 55s sys:7s)
1:03:27 INF -     write   1x(5% 8s sys:5s wait:2m17s)
1:03:27 INF - ----------------------------------------
1:03:27 INF -   features        7.8GB
1:03:27 INF -   mbtiles 2.7GB
```

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -515,6 +517,13 @@ public void run() throws Exception {
       return; // exit only if just fetching wikidata or downloading sources
     }
 
+    Path osmPolyFilePath =
+      arguments.file("poly_shape", "a .poly file defining the \"bounds\" to generate", null);
```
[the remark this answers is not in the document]
farfromrefug: maybe yes. For now i use poly extracts from geofabrik. I think it could be added later?

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileExtents.java
```diff
@@ -52,14 +85,35 @@ public boolean test(TileCoord tileCoord) {
...
+          x,
+          x + 1.0f,
+          y,
+          y + 1.0f
+        )));
+      }
       return testX(x) && testY(y);
     }
 
+    public boolean testOverShape(int x, int y, int z) {
```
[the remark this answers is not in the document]
farfromrefug: @msbarry i did that to prevent another test on `shape != null`. in `TiledGeometry` we only want to test if shape exists.
But in `test` we only want to test if `shape` too. So it would mean testing for shape twice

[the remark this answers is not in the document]
farfromrefug: actually i was doing the test for shape in `testOverShape` :s Is that ok if i remove the test for `shape!=null` in `testOverShape` ? 
Or do you prefer to use `test` but then have 2 tests `shape!=null`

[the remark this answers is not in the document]
farfromrefug: @msbarry if i use `test` then test will always be called even if no shape.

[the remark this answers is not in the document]
msbarry: Yes, I think that's fine though since if shape is null it will just do a range check on x/y which should be cheap/free

farfromrefug:
@msbarry this looks perfect! thanks a lot for the tests. Lot to learn from here. I ll try to do better next time

farfromrefug:
@msbarry why dors sonarcloud keeps on running on merge PRs? Quite a loss of energy

msbarry:
I'm not sure, it happens periodically... I deleted these branches in sonar cloud - hopefully that fixes it!

farfromrefug:
@msbarry still running :(

msbarry:
Just asked the sonar support people: https://community.sonarsource.com/t/sonarcloud-keeps-analyzing-branches-that-havent-changed/80419

PR #309, opened 2022-07-28
Lazy primitive group deserialization

msbarry:
- Reduce strain on JVM's garbage collector when reading an osm.pbf file by deferring parsing each primitive group (see https://wiki.openstreetmap.org/wiki/PBF_Format) so there's less deserialized data in-memory at any given point in time.
- Pull in osm protobuf definitions from https://github.com/openstreetmap/OSM-binary/tree/master/osmpbf to replace an LGPL dependency with MIT and give more flexibility for future tweaks
- Replace `LongHashSet` with more memory-efficient `Roaring64Bitmap` for storing IDs of ways that are members of multipolygons

Previously `osm_pass1` was taking ~3 minutes and spending 1.5 minutes in GC, now it takes 1:45-2:00 and spends 14 seconds in GC on a cg6d.metal instance. This also helps running with a smaller heap in memory-mapped mode (which should probably be the default now unlesss you have 150GB+ of RAM).


wipfli:
What is GC pressure and primitive group?

msbarry:
> What is GC pressure and primitive group?

I didn't think anyone would read this 😉 ! Updated the description, does that make more sense now?

PR #374, opened 2022-10-26
Create openmaptiles.yml

wipfli:
Adds a yml config file for the OpenMapTiles schema. Shortbread looks really good I think so we might get quite far with the yml already now. I am sharing this config file here just to track progress, this is far from complete...

Monaco looks already nice:

image


wipfli:
I plan to move this to the OpenMapTiles repo...

Issue #326, opened 2022-08-22
[FEATURE] Adding `-V/--version` flag to jar would be super nice! or how-do-i get the version if it is already avalible?


jessekrubin:
Adding a `-V`/`--version` cli flag would be super nice! 

Does it already exist in the cli? if so how do I get the current planetiler jar version?

wipfli:
Thanks @jessekrubin for your interest in Planetiler. I think it would be easier to understand the issue if you put the question in the content of the issue and keep the title short...

jessekrubin:
> Thanks @jessekrubin for your interest in Planetiler. I think it would be easier to understand the issue if you put the question in the content of the issue and keep the title short...

Done!

PR #376, opened 2022-10-29
Cli add version argument

ttomasz:
I saw https://github.com/onthegomap/planetiler/issues/326 and even though I have no experience with Java I thought it would be easy enough to add.

Let me know if this approach is ok or not.

Running `$ java -jar planetiler-dist/target/planetiler-dist-*-with-deps.jar --version` results in logger printing arguments and versions of modules at the end:
```
(...)
0:00:00 DEB - argument: version=true (show version then exit)
0:00:00 INF - module: planetiler-benchmarks - version: 0.6-SNAPSHOT
0:00:00 INF - module: planetiler-core - version: 0.6-SNAPSHOT
0:00:00 INF - module: planetiler-custommap - version: 0.6-SNAPSHOT
0:00:00 INF - module: planetiler-dist - version: 0.6-SNAPSHOT
0:00:00 INF - module: planetiler-examples - version: 0.6-SNAPSHOT
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,16 +161,32 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>de.westemeyer</groupId>
-        <artifactId>artifact-version-maven-plugin</artifactId>
-        <version>1.1.0</version>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>buildnumber-maven-plugin</artifactId>
+        <version>3.0.0</version>
         <executions>
           <execution>
+            <phase>validate</phase>
             <goals>
-              <goal>generate-service</goal>
+              <goal>create</goal>
             </goals>
           </execution>
         </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifest>
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+            </manifest>
+            <manifestEntries>
+              <Implementation-Build>${buildNumber}</Implementation-Build>
+              <Implementation-Timestamp>${maven.build.timestamp}</Implementation-Timestamp>
+            </manifestEntries>
+          </archive>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>com.diffplug.spotless</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,18 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>de.westemeyer</groupId>
+        <artifactId>artifact-version-maven-plugin</artifactId>
+        <version>1.1.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>generate-service</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
         <groupId>com.diffplug.spotless</groupId>
         <artifactId>spotless-maven-plugin</artifactId>
         <version>2.27.2</version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
     <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
     <sonar.exclusions>planetiler-benchmarks/**/*, planetiler-openmaptiles/**/*</sonar.exclusions>
     <revision>0.6-SNAPSHOT</revision>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -160,6 +160,18 @@
...
+        <artifactId>artifact-version-maven-plugin</artifactId>
+        <version>1.1.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>generate-service</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
msbarry, line 174: I think it would be useful to print the git hash and build time as well as version.  It looks like this approach would give us that: https://jdonkervliet.com/2021/10/22/git-version-in-jar.html - you'd just have to add the buildnumber-maven-plugin, and configure maven-jar-plugin to add the Implementation-Build and Implementation-Timestamp properties.  I tried it locally and got:

```
./mvnw -DskipTests clean package && unzip -p planetiler-dist/target/planetiler-dist-0.6-SNAPSHOT.jar META-INF/MANIFEST.MF
...
Implementation-Title: planetiler-dist
Implementation-Version: 0.6-SNAPSHOT
Implementation-Vendor-Id: com.onthegomap.planetiler
Implementation-Build: bd9f3ab22258da4eb9e6a5b0cbefd00fb0ff6df9
Implementation-Timestamp: 2022-11-11T11:11:20Z
```

Then we'd just have to figure out how to get those properties out of MANIFEST.MF.

msbarry (commented):
Thanks for getting this started! Sorry for the delay... a couple of comments and tweaks, let me know what you think.

pom.xml
ttomasz: Which POMs did you add the plugin to to get it to add stuff to manifest?

This should allow getting info from manifest https://stackoverflow.com/a/3777116

msbarry: Great! That snippet looks like it should do it.  I was able to get those in the manifest by adding to the root pom.xml build/plugins section:

```xml
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>buildnumber-maven-plugin</artifactId>
        <version>3.0.0</version>
        <executions>
          <execution>
            <phase>validate</phase>
            <goals>
              <goal>create</goal>
            </goals>
          </execution>
        </executions>
      </plugin>

...

      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-jar-plugin</artifactId>
        <configuration>
          <archive>
            <manifest>
              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
            </manifest>
            <manifestEntries>
              <Implementation-Build>${buildNumber}</Implementation-Build>
              <Implementation-Timestamp>${maven.build.timestamp}</Implementation-Timestamp>
            </manifestEntries>
          </archive>
        </configuration>
      </plugin>
```

ttomasz: I see. I checked and adding those two plugins makes build add version info to JAR but only to the smaller JAR without dependencies. Fat JAR with dependencies doesn't have that info.

Is that expected?

ttomasz: Also how do you run the no-deps file?
I'm running:
`$ java -cp planetiler-dist/target/planetiler-dist-0.6-SNAPSHOT.jar com.onthegomap.planetiler.Main --version`
and get:
`Exception in thread "main" java.lang.NoClassDefFoundError: org/openmaptiles/OpenMapTilesMain`

Running the jar with dependencies work but since version is not in manifest it just prints null values.

msbarry: Good point, this approach looks problematic because the jars get built so many different ways (jib, assembly plugin, regular maven).  Here's another approach that appears to work: https://github.com/onthegomap/planetiler/commit/c10f08da060b9b5f7840509766f0106bfccd41dc - basically it just builds a resource file in planetiler-core that captures the build info from maven and that goes everywhere the class files go so we don't need to worry about adding it to all the manifests.  What do you think?

msbarry: > Also how do you run the no-deps file

you need to add all of the transitive dependencies to the classpath as well.  It gets a little tricky to find them all, something like `java -cp "planetiler-dist/target/planetiler-dist-0.6-SNAPSHOT.jar:planetiler-core/target/planetiler-core-0.6-SNAPSHOT.jar:..." com.onthegomap.planetiler.Main --version

ttomasz: > What do you think?
Looks better than the previous solution :)

msbarry:
Fixes #326

msbarry:
Looks great, thanks @ttomasz !

msbarry:
oops

Issue #215, opened 2022-05-06
Deterministic ordering of points with same sort order


msbarry:
When 2 points have the same sort key, include the bytes when comparing them so that the ordering is deterministic.

_Originally posted by @nyurik in https://github.com/onthegomap/planetiler/discussions/214#discussioncomment-2697226_

nyurik:
Technically this might not fully solve the issue. Osm allows two points to have identical coordinates (or they could get the same x,y during simplification). You might want to add tag sorting as the last fallback.

bbilger:
Would you have some pointer for me where this sorting is happening currently? Then I'll take a look into this.

PR #382, opened 2022-11-09
215 deterministic ordering v2

phanecak-maptiler:
This PR contains changes which bring deterministic ordering of features with `.mbtiles` file generated by Planetiler:
* reuse source IDs for feature IDs, thus avoiding random changes of IDs between runs
* fix for random ordering of attributes in `SortableFeature`'s encoded value
* rework/adjustment of `SortableFeature` sorting to take into account also `value` when `key` is the same

This contributes towards fixing:
* fully issue #215 : some differences in generated tiles still remain, but so far seems not to be caused by ordering
* partially issue #214 : see above


phanecak-maptiler:
Stopping work on this. Will be covered by series of new pull-requests, starting with https://github.com/onthegomap/planetiler/pull/785 .

PR #387, opened 2022-11-18
Move NumberFormat to thread-local variable

erik:
The `NumberFormat` class isn't thread safe ([StackOverflow], [Docs]), and sharing one instance across threads allowed for races which resulted in otherwise valid OSM tags not being parsed correctly.

This surfaced as some rather strange exceptions during Planetiler runs that I wasn't able to reproduce when running each input individually. For example, both of these exceptions were not reproducible in unit tests:

```
java.lang.NumberFormatException: For input string: ""
java.lang.NumberFormatException: For input string: "1."
```

Given that it's a race condition, I'm not sure how to write a proper test for this, but this change reliably fixes the issue for me during local runs, and the existing tests for `Parse` should at least confirm that there aren't regressions.

[StackOverflow]: https://stackoverflow.com/a/1285297
[Docs]: https://docs.oracle.com/javase/8/docs/api/java/text/NumberFormat.html

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
@@ -17,7 +17,15 @@ import org.locationtech.jts.geom.Coordinate;
 public class Format {
 
   public static final Locale DEFAULT_LOCALE = Locale.getDefault(Locale.Category.FORMAT);
-  public static final ConcurrentMap<Locale, Format> instances = new ConcurrentHashMap<>();
+
+  // `NumberFormat` instances are not thread safe, so we need to wrap them inside a `ThreadLocal`.
+  //
+  // Ignore warnings about not removing thread local values since planetiler uses dedicated worker threads that release
+  // values when a task is finished and are not re-used.
+  @SuppressWarnings("java:S5164")
+  private static final ThreadLocal<ConcurrentMap<Locale, Format>> instances = ThreadLocal.withInitial(
+    ConcurrentHashMap::new);
+
   private final NumberFormat pf;
   private final NumberFormat nf;
   private final NumberFormat intF;
@@ -32,11 +40,7 @@ public class Format {
   }
 
   public static Format forLocale(Locale locale) {
-    Format format = instances.get(locale);
-    if (format == null) {
-      format = instances.computeIfAbsent(locale, Format::new);
-    }
-    return format;
+    return instances.get().computeIfAbsent(locale, Format::new);
   }
 
   public static Format defaultInstance() {
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
@@ -18,29 +18,39 @@ public class Format {
 
   public static final Locale DEFAULT_LOCALE = Locale.getDefault(Locale.Category.FORMAT);
 
+  private static final ConcurrentMap<Locale, Format> instances = new ConcurrentHashMap<>();
+
   // `NumberFormat` instances are not thread safe, so we need to wrap them inside a `ThreadLocal`.
   //
   // Ignore warnings about not removing thread local values since planetiler uses dedicated worker threads that release
   // values when a task is finished and are not re-used.
   @SuppressWarnings("java:S5164")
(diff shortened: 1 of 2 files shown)


erik:
> SonarCloud Quality Gate failed.    [Quality Gate failed](https://sonarcloud.io/dashboard?id=onthegomap_planetiler&pullRequest=387)
> 
> [Bug](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG) [C](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG) [1 Bug](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG) 

From my reading of [the issue](https://sonarcloud.io/organizations/onthegomap/rules?open=java%3AS5164&rule_key=java%3AS5164), I believe this warning can be safely suppressed, but would like a second opinion. `Parse` is a static utility class that has no obvious lifecycle, and as far as I can tell we'd want the `NumberFormat` instance available for the duration of the thread.

msbarry:
Yikes! Good catch.  I'm a bit concerned about other locations that do this as well.  It looks like both `Format` and `Parse` classes use number format instances. Could you update those as well? Looking at a recent run on the planet there's only a small amount of time spent in any of those, so the overhead of getting the thread local variable should be fine.

msbarry:
> > SonarCloud Quality Gate failed.    [Quality Gate failed](https://sonarcloud.io/dashboard?id=onthegomap_planetiler&pullRequest=387)
> > [Bug](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG) [C](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG) [1 Bug](https://sonarcloud.io/project/issues?id=onthegomap_planetiler&pullRequest=387&resolved=false&types=BUG)
> 
> From my reading of [the issue](https://sonarcloud.io/organizations/onthegomap/rules?open=java%3AS5164&rule_key=java%3AS5164), I believe this warning can be safely suppressed, but would like a second opinion. `Parse` is a static utility class that has no obvious lifecycle, and as far as I can tell we'd want the `NumberFormat` instance available for the duration of the thread.

I think that's fine, can you put `@SuppressWarnings("java:S5164")` on that thread local definition with a short comment explaining that?

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
```diff
@@ -17,7 +17,15 @@
...
+  //
+  // Ignore warnings about not removing thread local values since planetiler uses dedicated worker threads that release
+  // values when a task is finished and are not re-used.
+  @SuppressWarnings("java:S5164")
+  private static final ThreadLocal<ConcurrentMap<Locale, Format>> instances = ThreadLocal.withInitial(
+    ConcurrentHashMap::new);
+
   private final NumberFormat pf;
   private final NumberFormat nf;
   private final NumberFormat intF;
```
msbarry, line 31: I think we want to do these 3 as well.

erik: Ah, you're right. I was hoping there wasn't a case where one thread constructs a `Format` instance and then shares it to another thread, but seems that very likely is the case with `ProgressLoggers`

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Format.java
```diff
@@ -17,26 +17,40 @@
...
+    nf = ThreadLocal.withInitial(() -> {
+      var f = NumberFormat.getNumberInstance(locale);
+      f.setMaximumFractionDigits(1);
+      return f;
+    });
+    intF = ThreadLocal.withInitial(() -> {
+      var f = NumberFormat.getNumberInstance(locale);
+      f.setMaximumFractionDigits(0);
+      return f;
+    });
```
msbarry, line 49: minor thing - you could probably just inline these in the field definitions since they don't need anything outside the initializer

erik: Is that possible if they need access to the `locale` parameter passed to the constructor?

msbarry: oh yeah, good point!

PR #397, opened 2022-11-29
Remove outdated reference to "relation_member" type

erik:
Support was removed in #323, and trying to use `matchType("relation_member")` will currently throw an exception


msbarry:
Thank you!

PR #399, opened 2022-12-01
Fix max tile size in stats output.

erik:
Another tiny fix - `printTileStats()` was always printing something like `0:01:12 DEB [mbtiles] - all avg:715 max:0` because the variable wasn't being updated.


msbarry:
Thanks!

PR #403, opened 2022-12-01
Switch ordering of translation providers

msbarry:
Prefer OSM translations and only use wikidata as a fallback.


ZeLonewolf:
Thanks!

PR #406, opened 2022-12-06
Add shapefile directory source

erik:
This PR adds `Planetiler.addShapefileDirectorySource`, which can be used to add all Shapefiles matching a glob pattern in a directory to a single named source. Other than grouping related shapefiles together, the main benefit here is that the files can then be processed concurrently.

Was mentioned in Slack a couple weeks back: https://osmus.slack.com/archives/C031V9E9RMG/p1668717797875339

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -25,7 +25,9 @@ import com.onthegomap.planetiler.util.ResourceUsage;
 import com.onthegomap.planetiler.util.Translations;
 import com.onthegomap.planetiler.util.Wikidata;
 import com.onthegomap.planetiler.worker.RunnableThatThrows;
+import com.onthegomap.planetiler.worker.WorkerPipeline;
 import java.io.IOException;
+import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Instant;
@@ -249,6 +251,49 @@ public class Planetiler {
   public Planetiler addShapefileSource(String name, Path defaultPath, String defaultUrl) {
     return addShapefileSource(null, name, defaultPath, defaultUrl);
   }
+
+  /**
+   * Adds a new ESRI shapefile directory source that will process all files under {@param basePath} matching
+   * {@param globPattern} using an explicit projection.
+   *
+   * @param projection  the Coordinate Reference System authority code to use, parsed with
+   *                    {@link org.geotools.referencing.CRS#decode(String)}
+   * @param name        string to use in stats and logs to identify this stage
+   * @param basePath    path to the directory containing shapefiles to process
+   * @param globPattern string to match filenames against, as described in {@link FileSystem#getPathMatcher(String)}.
+   * @return this runner instance for chaining
+   * @see ShapefileReader
+   */
+  public Planetiler addShapefileDirectorySource(String projection, String name, Path basePath, String globPattern) {
+    Path dirPath = getPath(name, "shapefile directory", basePath, null);
+    return addStage(name, "Process all features matching " + dirPath + "/" + globPattern, ifSourceUsed(name, () -> {
+      var pathMatcher = dirPath.getFileSystem().getPathMatcher("glob:" + globPattern);
+
+      try (var dirWalker = Files.walk(dirPath)) {
+        WorkerPipeline.start(name, stats)
+          .readFrom("reader", dirWalker.filter(pathMatcher::matches)::iterator)
+          .addBuffer("read_queue", 1000)
+          .sinkToConsumer("process", config.featureProcessThreads(),
+            path -> ShapefileReader.processWithProjection(projection, name, path, featureGroup, config, profile, stats))
+          .await();
+      }
+    }));
+  }
+
+  /**
+   * Adds a new ESRI shapefile directory source that will process all files under {@param basePath} matching
+   * {@param globPattern}.
+   *
+   * @param name        string to use in stats and logs to identify this stage
+   * @param basePath    path to the directory containing shapefiles to process
+   * @param globPattern string to match filenames against, as described in {@link FileSystem#getPathMatcher(String)}.
+   * @return this runner instance for chaining
+   * @see ShapefileReader
+   */
+  public Planetiler addShapefileDirectorySource(String name, Path basePath, String globPattern) {
+    return addShapefileDirectorySource(null, name, basePath, globPattern);
+  }
+
 
   /**
    * Adds a new ESRI shapefile source that will be processed with an explicit projection when {@link #run()} is called.
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -250,6 +253,62 @@ public Planetiler addShapefileSource(String name, Path defaultPath, String defau
...
+
+          WorkerPipeline.start(sourceName, stats)
+            .readFrom("reader", pathStream::iterator)
+            .addBuffer("read_queue", 1000)
+            .sinkToConsumer("process", config.featureProcessThreads(), path -> {
+              LogUtil.setStage(sourceName, path.getFileName().toString());
+              ShapefileReader.processWithProjection(projection, sourceName, path, featureGroup, config, profile, stats,
+                false);
+              LogUtil.clearStage();
+            })
```
msbarry, line 289: Internally `processWithProjection` is kicking off its own worker pipeline.  I think we only want one worker pipeline in order to spread work out appropriately across all cores.  Can we push this handling down into the reader - possibly refactoring `SimpleReader.process` to be multi-file aware? Ideally this will let us support glob patterns for any new kind of reader (geopackage, flatgeobuf, etc.) without having to add special handling for globs for that file type. I'm picturing we get a count from each file ahead of time for progress reporting, then the reader thread iterates through each file and enqueues all of the elements and the rest of the pipeline should remain unaffected.

erik: So to make sure I'm following, that would mean that `processWithProjection` takes a `Stream<Path>` or `List<Path>` instead of a single `Path`?

msbarry: Yes exactly, at first I was thinking a `List<Path>` so you could traverse once to get the total count, and a second time to actually extract the elements.  Would that list be too big?

acalcutt:
I'm interested in this for contour lines. I was taking the jaxa DEM files and running them through gdal_contour, which by default export to shapefile. would love to just feed those into planetiler.

erik:
@msbarry refactored so that `SimpleReader` can handle multiple file inputs without the child classes having to reimplement all the logic. IMO it's still kind of clunky - `NaturalEarthReader` for example wouldn't ever need to support multiple files as far as I can tell, so it's a bit of an unnecessary change.

I tried to avoid creating additional levels of abstraction, but the refactor is still a bit bulkier than I'd like. Open to suggestions

msbarry:
> @msbarry refactored so that `SimpleReader` can handle multiple file inputs without the child classes having to reimplement all the logic. IMO it's still kind of clunky - `NaturalEarthReader` for example wouldn't ever need to support multiple files as far as I can tell, so it's a bit of an unnecessary change.
> 
> I tried to avoid creating additional levels of abstraction, but the refactor is still a bit bulkier than I'd like. Open to suggestions

Thanks for making that change - the only other thing I could think of would be to keep each Reader wrapping a single file, but extract the logic from `SimpleReader.process` so it can operate on a list of readers so the readers wouldn't need to change much (or at all?)

I agree it doesn't make sense for natural earth reader, but I think it will for geojson, geopackage, etc...

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/SimpleReader.java
```diff
@@ -49,15 +54,18 @@ protected SimpleReader(Profile profile, Stats stats, String sourceName) {
...
     int processThreads = config.featureProcessThreads();
     Envelope latLonBounds = config.bounds().latLon();
     AtomicLong featuresRead = new AtomicLong(0);
     AtomicLong featuresWritten = new AtomicLong(0);
 
     var pipeline = WorkerPipeline.start(sourceName, stats)
-      .fromGenerator("read", read())
-      .addBuffer("read_queue", 1000)
+      .readFrom("source_paths", sourcePaths)
+      .addBuffer("read_queue", 1000, 1)
```
msbarry, line 66: This allocates a new thread to iterate through the list, we could skip that by either moving the logic to iterate through the paths into the read thread, or change it to use `readFromTiny` which just copies every element from sourcePaths into a queue and makes that the input to the next stage.  I might suggest the latter since it allows us to split the work of read thread among multiple CPUs since reading shapefiles can sometimes be a bottleneck.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/SimpleReader.java
```diff
@@ -125,8 +133,25 @@ private FeatureRenderer newFeatureRenderer(FeatureGroup writer, PlanetilerConfig
...
+    }
+    return numFeatures;
+  }
+
+  /** Returns the number of features to be read from a single path to use for displaying progress. */
+  public abstract long getCountForPath(Path path);
 
   /** Returns a source that initiates a {@link WorkerPipeline} with elements from this data provider. */
-  public abstract WorkerPipeline.SourceStep<? extends SourceFeature> read();
+  public abstract void readPath(Path path, Consumer<F> next) throws Exception;
```
msbarry, line 148: Sonar is complaining but I think this is fine since a bunch of different kinds of exceptions can be thrown and we want them all to terminate the whole thing - you can add `@SuppressWarnings("java:S112")` so sonar stops bugging us

erik:
> Thanks for making that change - the only other thing I could think of would be to keep each Reader wrapping a single file, but extract the logic from `SimpleReader.process` so it can operate on a list of readers so the readers wouldn't need to change much (or at all?)

I think this makes sense, I was prototyping something similar initially, but didn't want to make too many design decisions in an unrelated PR. I can try this route, but maybe the readers should be lazily constructed rather than being passed in all at once?

planetiler-core/src/test/java/com/onthegomap/planetiler/reader/ShapefileReaderTest.java
```diff
@@ -54,35 +55,29 @@ void testReadShapefileUnzipped() throws IOException {
   }
 
   private static void testReadShapefile(Path path) {
```
msbarry, line 57: Can we add another test that verifies it handles multiple shapefiles? It might be sufficient just to pass in `List.of(path, path)`? Alternatively we could add a test just for `SourceFeatureProcessor` that tests its handling with a stub implementation of a reader.

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -92,23 +92,25 @@ private static <T extends OsmElement> T with(T elem, Consumer<T> fn) {
...
-
-      @Override
-      public long getCount() {
-        return features.size();
-      }
+  private <F extends SourceFeature> void processReaderFeatures(FeatureGroup featureGroup, Profile profile,
+    PlanetilerConfig config, List<F> features) {
+    SourceFeatureProcessor.processFiles(
+      "test",
+      List.of(Path.of("dummy-path")), path -> new SimpleReader<F>("test") {
```
msbarry, line 99: minor nit: instead of dummy-path, maybe mock-path?

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1661,6 +1663,7 @@ public void processFeature(SourceFeature source, FeatureCollector features) {
       .addOsmSource("osm", tempOsm)
       .addNaturalEarthSource("ne", TestUtils.pathToResource("natural_earth_vector.sqlite"))
       .addShapefileSource("shapefile", TestUtils.pathToResource("shapefile.zip"))
+      .addShapefileDirectorySource("shapefiledir", TestUtils.pathToResource(""), "shapefile*.zip")
```
msbarry, line 1666: Is the intention here to include that same file as the line above just through glob matching? If so, you could add another variation to the ValueSource arguments of the test to switch whether this runs with `addShapefileSource` or `addShapefileDirectorySource` to make sure they both produce the same results?

msbarry (commented):
Looks great! A couple of comments on the tests but should be good to go after that.

One thing I'm wondering is if this would be flexible enough to allow us to switch over to the shapefile distribution of natural earth and deprecate the natural earth reader? It's basically a zip file with different `.shp` files nested in it where we'd want to map `xyz.shp` to `source_layer=xyz`... 

Looks like there's also a geopackage source, so we could use that instead after merging #413.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/ShapefileReader.java
```diff
@@ -150,34 +132,30 @@ private ShapefileDataStore open(Path path) {
...
+      while (iter.hasNext()) {
+        org.opengis.feature.simple.SimpleFeature feature = iter.next();
+        Geometry source = (Geometry) feature.getDefaultGeometry();
+        Geometry latLonGeometry = source;
+        if (transformToLatLon != null) {
+          latLonGeometry = JTS.transform(source, transformToLatLon);
+        }
+        if (latLonGeometry != null) {
+          SimpleFeature geom = SimpleFeature.create(latLonGeometry, new HashMap<>(attributeNames.length),
+            sourceName, null, featureId.incrementAndGet());
```
msbarry, line 151: I don't think this ID needs to be globally unique, would probably be fine to just increment a local counter here to convey what order it was contained in the file.

erik: Is it okay if different features in the same source layer (potentially also the same tile) have the same id?

msbarry: There's a different process that generates vector tile feature IDs. The only use of this ID is if you need conditional logic to handle a specific feature differently and all you have to go by is the order that it appears in an input file.

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
erik: Good idea, added a more explicit test.

erik:
> One thing I'm wondering is if this would be flexible enough to allow us to switch over to the shapefile distribution of natural earth and deprecate the natural earth reader? It's basically a zip file with different `.shp` files nested in it where we'd want to map `xyz.shp` to `source_layer=xyz`...

Even if we don't swap out the `NaturalEarthReader` I think this would be nice functionality to add. But at this point it should be in a separate PR.

> Looks like there's also a geopackage source, so we could use that instead after merging #413.

That might just work out of the box honestly. Worth testing.

msbarry:
I think there might be some issue in the geotools shapefile reader where it doesn't fully close the resource then windows complains when you try to delete the file...

erik:
Yeah... I'm running out of ideas to try here. As far as I can tell the relevant `close()` methods should be called everywhere. It is odd that the other tests didn't have the same behavior previously though. What do you think?

msbarry:
> Yeah... I'm running out of ideas to try here. As far as I can tell the relevant `close()` methods should be called everywhere. It is odd that the other tests didn't have the same behavior previously though. What do you think?

Yeah it looks like you're doing the right thing. Maybe just get rid of that new test and go back to testing this in `testPlanetilerRunner`, add another `@ValueSource` argument that runs it with a `--use-glob` argument or something, then in the body of the test check if that argument is present and if so then reference that shapefile using a glob matcher instead of addShapefileSource.

Alternatively if you wanted to keep going with the separate test, you could add a second shapefile resource and have it match them with a glob so it doesn't need to delete the files to clean up.

erik:
Adding the copy of the file feels gross, but I think it's useful to test that the directory source actually does iterate over multiple files in this test.

I still think the failure on Windows is really bizarre... seems like too obvious of a bug to go unnoticed in gt-shapefile 🤷

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1684,6 +1686,46 @@ public void processFeature(SourceFeature source, FeatureCollector features) {
...
+          switch ((String) feature.attrs().get("source")) {
+            case "shapefile" -> fileCount++;
+            case "shapefile-dir" -> dirCount++;
+          }
+        }
+      }
+
+      // Input file was copied twice into test directory, directory source should have
+      // 2x the number of features.
+      assertEquals(2 * fileCount, dirCount);
```
msbarry, line 1725: might want to also assert that fileCount/dirCount are >0 just to make sure we're not saying things are ok if for some reason it doesn't read anything

msbarry (commented):
Looks good! One minor suggestion on the tests then should be good to merge.

msbarry:
> Adding the copy of the file feels gross, but I think it's useful to test that the directory source actually does iterate over multiple files in this test.
> 
> I still think the failure on Windows is really bizarre... seems like too obvious of a bug to go unnoticed in gt-shapefile 🤷

Yeah it's annoying, it could just be that we are missing some step when closing the resource. I looked into it in the past and couldn't quite figure it out but realized the impact is not too big in practice because the process lives for a short period of time then releases resources when it's done.

One additional clue here is that sometimes I see this message [in the logs of a full planet run](https://github.com/onthegomap/planetiler/blob/e9f78d6471ed60a37d0c531488e8bec595bef01b/planet-logs/v0.5.0-planet-c6gd-32gb.txt#L1225-L1227), exactly 1 hour after the shapefile is originally opened:

```
Jun 05, 2022 4:34:08 PM org.geotools.referencing.factory.DeferredAuthorityFactory disposeBackingStore
INFO: Disposing class org.geotools.referencing.factory.epsg.hsql.ThreadedHsqlEpsgFactory backing store
1:00:01 INF - Database closed
```

Issue #46, opened 2021-12-30
[FEATURE] Geojson input


msbarry:
Add support for newline-separated geojson input files, for example contour lines generated by gdal-contour.

durkie:
do you envision that planetiler might pick up where tippecanoe left off? i regularly use tippecanoe with newline separated geojson, and it's a great piece of software, but it doesn't seem like it's getting much maintenance/development these days.

msbarry:
Interesting, I doesn't realize the last commit to tippecanoe was almost a year ago. My guess is it's deprecated in favor of their new proprietary tiling service. 

I think there are some jobs people use tippecanoe for that planetiler could handle, but not sure about all of the density-based geometry filtering. Planetiler actually looks more like an alternative to the tiling service. 

The initial motivation was talking with someone trying to generate contour lines from DEM data. They used gdal contour to generate 1x1 degree contour tiles then they tried to use tippecanoe but it fell over trying to hold the planet in memory. Planetiler should be able to fill that gap since it does the same job but only holds a single input and output tile in memory at a time.

Supporting more input formats, and exposing more through config (#18) so you can do more without writing Java code would open up handling more use-cases from tippecanoe.

flother:
What do you think about supporting [GeoPackage](https://www.geopackage.org/) as an input format too? It's essentially a standardised SQLite database format — a bit like MBTiles — and it's supposed to be a replacement for the venerable shapefile. I've seen it used quite a bit for government data.

msbarry:
Makes sense! As long as you can read shapes with key/value tags from it then we should be able to use it as an input source.

wipfli:
Can planetiler handle shapefiles (https://en.wikipedia.org/wiki/Shapefile)?

The Swiss government distributes shapefiles (https://data.geo.admin.ch/) for example for natural protected areas. In the past, I  converted shapefiles to GeoJSON and then to mbtiles with `tippecanoe`.

msbarry:
> Can planetiler handle shapefiles (https://en.wikipedia.org/wiki/Shapefile)?

Yes, in fact planetiler currently can only handle shapefiles, natural earth sqlite file, and .osm.pbf files.  The openmaptiles profile uses shapfiles for [water polygons](https://osmdata.openstreetmap.de/data/water-polygons.html) and [lakelines](https://github.com/lukasmartinelli/osm-lakelines):

[...]

Issue #126, opened 2022-03-08
Remove geotools dependency for reading shapefiles


msbarry:
Geotools is not in maven central which adds an extra step for people to build the project, and the licensing is not great. Geotools is really only needed to read shapefiles (including coordinate system transformation) which there should be another option for.

[Apache SIS](https://sis.apache.org/) looks promising.

bchapuis:
I also started playing with Apache SIS in Baremaps. 

https://github.com/baremaps/baremaps/issues/419

Unfortunately, the shapefile reader in Apache SIS is rudimentary and has not yet been released in maven central. Therefore, I asked Geotoolkit (LGPL) about their intent submit some of their implementation of a shapefile store modules to Apache SIS.

https://github.com/Geomatys/geotoolkit/issues/22

PR #413, opened 2022-12-12
Add basic support for reading GeoPackage files.

erik:
This PR adds `Planetiler.addGeoPackageSource`, which can be used to consume features from a [geopackage] file. 

Mostly implemented as a direct port of `ShapefileReader`, with a couple differences:

- Always reads the CRS from the GeoPackage
- May read multiple vector layers from the same file.

The file `geopackage.gpkg` was converted from `shapefile.zip` using `ogr2ogr`.

See also: #46.

[geopackage]: https://www.geopackage.org/


planetiler-core/pom.xml
```diff
@@ -138,6 +138,11 @@
       <artifactId>guava</artifactId>
       <version>31.1-jre</version>
     </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-geopkg</artifactId>
+      <version>${geotools.version}</version>
+    </dependency>
```
msbarry, line 145: I'm trying to eventually remove the dependency on geotools since it has licensing issues and the artifacts aren't in maven central (#126) - is there another library we could use instead that provides equivalent functionality?

planetiler-core/src/test/java/com/onthegomap/planetiler/reader/GeoPackageReaderTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+class GeoPackageReaderTest {
+
+  @Test
+  @Timeout(30)
+  void testReadGeoPackage() {
+    testReadGeoPackage(TestUtils.pathToResource("geopackage.gpkg"));
+  }
+
+  private static void testReadGeoPackage(Path path) {
```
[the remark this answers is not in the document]
erik: It doesn't - copy paste artifact from `ShapefileReaderTest` :)

planetiler-core/pom.xml
erik: 👍 wasn't aware of that, just went with it because `geotools` was already used.

There's http://ngageoint.github.io/geopackage-java/ (MIT License)

msbarry: Thanks, that looks promising! Also please update `NOTICE.md` with a blurb about the dependency you end up going with.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -0,0 +1,108 @@
...
+    long id = 0;
+
+    for (var featureName : geoPackage.getFeatureTables()) {
+      FeatureDao features = geoPackage.getFeatureDao(featureName);
+
+      MathTransform transform = CRS.findMathTransform(
+        CRS.decode("EPSG:" + features.getSrsId()),
+        latLonCRS);
+
+      for (var feature : features.queryForAll()) {
```
msbarry, line 80: Not strictly necessary for first pass, but because geopackage seems to have a spatial index built-in, would it be possible to limit this query by bounding box by passing `config.bounds` into the constructor if it's set?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -0,0 +1,108 @@
...
+import mil.nga.geopackage.geom.GeoPackageGeometryData;
+import org.geotools.geometry.jts.JTS;
+import org.geotools.geometry.jts.WKBReader;
+import org.geotools.referencing.CRS;
+import org.locationtech.jts.geom.Geometry;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.MathTransform;
+
+/**
+ * Utility that reads {@link SourceFeature SourceFeatures} from the vector geometries contained in a GeoPackage file.
```
msbarry, line 23: I'd like to try migrating natural earth reader to this after we merge. That comes in a `.gpkg.zip` file. I'm wondering if we want this to have the same handling that [natural earth reader has](https://github.com/onthegomap/planetiler/blob/cb082a0566547934281891d4426ea5dbfd245f1c/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/NaturalEarthReader.java#L89-L106) where it detects that and unzips it first if necessary or if that unzip step should be a reusable thing we could tack onto any reader...?  

The ideal situation is where the reader is OK reading directly out of the zip like shapefiles could do and we don't need the intermediate file. Other formats would likely be the same, but as far as I could tell sqlite needs to point at an actual file on disk.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        next.accept(geom);
+      }
+    }
+  }
+
+  @Override
+  public void close() {
+    geoPackage.close();
```
msbarry, line 106: Hopefully this one actually works 🤣

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1663,6 +1663,7 @@ public void processFeature(SourceFeature source, FeatureCollector features) {
       .addOsmSource("osm", tempOsm)
       .addNaturalEarthSource("ne", TestUtils.pathToResource("natural_earth_vector.sqlite"))
       .addShapefileSource("shapefile", TestUtils.pathToResource("shapefile.zip"))
+      .addGeoPackageSource("geopackage", TestUtils.pathToResource("geopackage.gpkg"), null)
```
msbarry, line 1666: Could we add something to verify these features actually make it to the output, for example something in the mock profile above that emits some features for the stations and assert they show up in output below? For each new user-facing feature want to make sure there's one integration test here that verifies it works end to end.

msbarry (commented):
This looks great! And it's a good template for adding support for other kinds of input formats (#46).

There's one more step of exposing this source type to the yaml configs in planetiler-custommap project. Do you want to tackle that in this review or a followup?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
erik: Great idea. I looked through the code a bit, and seems that `mil.nga.geopackage` will try to modify the file in place to build the index if it doesn't already exist, which we likely do not want. It'd be possible to first check if a gpkg file already has an index, and if not falling back to `queryForAll()`, but there might be some more nuance to it. I'd suggest we leave it as a follow up.

Something like this:

```suggestion
      Iterable<FeatureRow> featureIter = () -> {
        FeatureIndexManager indexManager = new FeatureIndexManager(geoPackage, featureName);
        if (indexManager.isIndexed()) {
          return indexManager.query(configuredBbox).iterator();
        }

        return features.queryForAll().iterator();
      };

      for (FeatureRow feature : featureIter) {
```

msbarry (commented):
@erik looks good! Thanks for adding the tests and yaml bindings.  Last remaining comment is about the zip file handling - what do you think there?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
erik: Ah! Misread your comment last time. Sure, I think that'd be useful to add here. Since sqlite files can't be read directly from a zip file I think we might need to change behavior in #423 if we want to support GPKG glob matching, but I'd like to get that PR merged first (or do this change as a follow up)

msbarry: Ok I merged #423, did you want to include the zip handling in this PR?

erik: How about doing it as a quick follow up? I can create a PR focused on adding natural earth support

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -0,0 +1,108 @@
...
+      for (var feature : features.queryForAll()) {
+        GeoPackageGeometryData geometryData = feature.getGeometry();
+        if (geometryData == null) {
+          continue;
+        }
+
+        Geometry featureGeom = (new WKBReader()).read(geometryData.getWkb());
+        Geometry latLonGeom = (transform.isIdentity()) ? featureGeom : JTS.transform(featureGeom, transform);
+
+        SimpleFeature geom = SimpleFeature.create(latLonGeom, new HashMap<>(),
```
msbarry, line 89: Minor tweak, could you set the hash map initial size from column count to avoid resizing?

PR #421, opened 2022-12-20
Expose `min_size` in configurable schema

wipfli:
Exposes the `Feature::setMinPixelSize` function in the configurable schema as `min_size`.

This follows the implementation of `min_zoom`.

Related to https://github.com/onthegomap/planetiler/issues/349

PR #422, opened 2022-12-20
Add a custommap standalone pom file

wipfli:
My workflow is usually to code in visual studio code without much java support. Then I go to the terminal, compile and run the application.

Maybe this is a bit strange if you are used to working with intelij or another good ide, but this is how I usually code...

So for me it is super helpful if I can run the configurable schema standalone, like I can run the things in the examples folder standalone...

Let me know if this is the right approach.

### DIFF
--- a/planetiler-custommap/README.md
+++ b/planetiler-custommap/README.md
@@ -4,9 +4,7 @@ You can define how planetiler turns input sources into vector tiles by running p
 file as the first argument:
 
 ```bash
-# from a java build
-java -jar planetiler.jar schema.yml
-# or with docker (put the schema in data/schema.yml to include in the attached volume)
+# With docker (put the schema in data/schema.yml to include in the attached volume)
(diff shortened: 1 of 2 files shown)


msbarry:
@wipfli Are you making changes to the java code in this workflow? If not, it might just be best to download the latest compiled jar file, or run the latest docker build against the yaml file you are working on.

If we do keep this, we'd want to update a few more places to ensure this pom stays correct and up-to-date: https://github.com/search?q=repo%3Aonthegomap%2Fplanetiler+standalone.pom.xml&type=code

wipfli:
I was changing the source code of `planetiler-custommap` to expose a `min_pixel_size` feature, see #421. I first did this in intellij and everything worked well. It somehow found all the relevant classes. Then I wanted to switch to the command line to compile and run my updated version of planetiler-custommap but could not figure out how to do it without adding a standalone pom file.

PR #423, opened 2022-12-21
Extract ZIP file walker out of ShapefileReader.

erik:
This PR pulls ZIP file handling out of `ShapefileReader` so that it can be reused for different readers. The changes here can be applied to other reader types as we add them.

Added:

- support for reading multiple `.shp` files from a single `.zip`
- support glob matching within a `.zip` file

There's a breaking change in naming: `addShapefileDirectorySource` becomes `addShapefileGlobSource`. I figured `addShapefileDirectorySource` was new enough that the more accurate name was worth the churn, as the pattern can now be applied both to ZIP archives and directories.

Another change in behavior: previously, if you had a `.zip` with two `.shp` files, only one would be read. Now (unless a glob pattern filters one out), both will be read.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -328,7 +328,10 @@ public class Planetiler {
     Path path = getPath(name, "shapefile", defaultPath, defaultUrl);
     return addStage(name, "Process features in " + path,
       ifSourceUsed(name, () -> {
-        var sourcePaths = FileUtils.hasExtension(path, "zip") ? globMatchPath(path, "*.shp") : List.of(path);
+        List<Path> sourcePaths = List.of(path);
+        if (FileUtils.hasExtension(path, "zip") || Files.isDirectory(path)) {
+          sourcePaths = globMatchPath(path, "*.shp");
+        }
 
         ShapefileReader.processWithProjection(projection, name, sourcePaths, featureGroup, config, profile, stats);
       }));
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -26,18 +26,14 @@ import com.onthegomap.planetiler.util.Translations;
 import com.onthegomap.planetiler.util.Wikidata;
 import com.onthegomap.planetiler.worker.RunnableThatThrows;
 import java.io.IOException;
-import java.io.UncheckedIOException;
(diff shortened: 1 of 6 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -752,6 +764,50 @@ private void ensureInputFilesExist() {
     }
   }
 
+  // Helper for globMatchPath which will not recurse into zip files.
+  private List<Path> globMatchPath(Path basePath, String pattern) {
```
msbarry, line 768: nit: instead of inline comment, maybe rename this to something to indicate it doesn't recurse into nested zips?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -752,6 +764,50 @@ private void ensureInputFilesExist() {
...
+  }
+
+  /**
+   * Returns a list of paths matching {@param pattern} within {@param basePath}, which may be either a ZIP archive or a
+   * directory.
+   * <p/>
+   * When {@param basePath} is a directory, the additional {@param onZipFile} function will be called whenever a ZIP
+   * archive matching {@param pattern} is found. This allows a different pattern to be used within the archive.
+   */
+  private List<Path> globMatchPath(Path basePath, String pattern, Function<Path, List<Path>> onZipFile)
```
msbarry, line 783: We might want to move this into FileUtils with a simple test or two just to verify it's doing what we expect without all of the mapping logic included.

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1700,30 +1700,35 @@ public void processFeature(SourceFeature source, FeatureCollector features) {
...
       }
 
-      // Input file was copied twice into test directory, directory source should have
-      // 2x the number of features.
       assertTrue(fileCount > 0);
-      assertEquals(2 * fileCount, dirCount);
+      // `shapefile` and `shapefile-glob-zip` both match only one file.
+      assertEquals(fileCount, globZipCount);
+      // `shapefile-glob` matches two input files, should have 2x number of features of `shapefile`.
+      assertEquals(2 * fileCount, globCount);
```
msbarry, line 1731: Thanks, this is a good e2e test to convince me this all works!

msbarry (commented):
Looks good overall!  One comment about extracting the glob stuff to FileUtils so we can have a test to show what it's doing.

msbarry:
Looks great, thanks @erik!

PR #430, opened 2023-01-02
Support unzipping GeoPackage sources at runtime

erik:
Extension to #413 which allows us to read GeoPackage sources from ZIP archives.

Testing locally, I was able to use the NaturalEarth data from https://naciscdn.org/naturalearth/packages/natural_earth_vector.gpkg.zip as a direct replacement for `addNaturalEarthSource()`. 

We should be able to remove `NaturalEarthReader`, but since there are several references to `natural_earth_vector.sqlite.zip` which will break, I've left it as a `@Deprecated` annotation for now. At the very least I think we'd need to update the `planetiler-openmaptiles` repo first to keep tests from breaking.


planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -358,8 +361,18 @@ public Planetiler addShapefileSource(String projection, String name, Path defaul
   public Planetiler addGeoPackageSource(String name, Path defaultPath, String defaultUrl) {
     Path path = getPath(name, "geopackage", defaultPath, defaultUrl);
     return addStage(name, "Process features in " + path,
-      ifSourceUsed(name,
-        () -> GeoPackageReader.process(name, List.of(path), featureGroup, config, profile, stats)));
+      ifSourceUsed(name, () -> {
+        List<Path> sourcePaths = List.of(path);
+        if (FileUtils.hasExtension(path, "zip")) {
+          sourcePaths = FileUtils.unzipMatchingFiles(path, "*.gpkg", tmpDir);
+        }
```
msbarry, line 368: Can we push the unzip handling inside geopackage reader so addGeoPackageSource enumerates the files in the zip file and passes that list of paths into the geopackage reader, then geopackage reader extracts them first if necessary?

That would move toward making the reader usage more consistent and encapsulate the logic of whether they can read into a zip file or need to extract first inside the reader.  It would also let the reader explicitly delete the unzipped file on close to free up disk space as soon as it's not needed since we can't always trust deleteOnExit to happen.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/FileUtils.java
```diff
@@ -314,4 +322,29 @@ public static void unzip(InputStream input, Path destDir) {
...
+   * @return list of paths to extracted files
+   */
+  public static List<Path> unzipMatchingFiles(Path zipPath, String pattern, Path tmpLocation) throws IOException {
+    PathMatcher matcher = createGlobMatcher(zipPath.getFileSystem(), pattern);
+
+    try (var zipFs = FileSystems.newFileSystem(zipPath)) {
+      return walkFileSystemWithMatcher(zipFs, matcher)
+        .map(path -> {
+          try {
+            Path unzippedPath = tmpLocation.resolve(path.getFileName().toString());
```
msbarry, line 339: Probably unlikely but there could be 2 files with the same name in different directories with the same name they will clobber eachother.

Also, I think `resolve` can take a path so you don't need toString

planetiler-core/src/main/java/com/onthegomap/planetiler/util/FileUtils.java
```diff
@@ -314,4 +322,29 @@ public static void unzip(InputStream input, Path destDir) {
...
+            Files.copy(path, unzippedPath, StandardCopyOption.REPLACE_EXISTING);
```
msbarry, line 340: Can we use the `unzip` utility here that has some safety checks built in?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -74,8 +74,14 @@ public void readFeatures(Consumer<SimpleFeature> next) throws Exception {
...
 
+      // If left unset (e.g. in NaturalEarth's data), assume latlon
+      long srsId = features.getSrsId();
+      if (srsId == 0) {
+        srsId = 4326;
+      }
+
       MathTransform transform = CRS.findMathTransform(
-        CRS.decode("EPSG:" + features.getSrsId()),
+        CRS.decode("EPSG:" + srsId),
```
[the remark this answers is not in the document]
erik: From [the spec](https://www.geopackage.org/spec131/#r11):

> The record with an srs_id of 0 SHALL be used for undefined geographic coordinate reference systems.

So adding an explicit projection would be required for correctness, but if not given, maybe we can fallback to 4326 rather than throwing?

planetiler-core/src/main/java/com/onthegomap/planetiler/util/FileUtils.java
erik: Would you want to extend `unzip` to only extract certain files, or unzip every file in the archive? The latter is probably fine by itself, but might conflict with the idea of pushing the extraction into the individual GeoPackageReader instead as you suggested here https://github.com/onthegomap/planetiler/pull/430#discussion_r1060510769

msbarry:
Hey @erik  just checking in on this, were you planning to make more changes?

erik:
@msbarry yeah, hopefully will have a chance this upcoming week, holidays have kept me busy. Couple of questions for you in the comments above which might reduce number of review iterations

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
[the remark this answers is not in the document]
erik: Updated to match shapefile reader behavior (will throw rather than assuming a default projection, can be overridden unconditionally)

planetiler-core/src/main/java/com/onthegomap/planetiler/util/FileUtils.java
erik: Since we only wanted to extract a single path, most of the safety checks in `unzip` weren't relevant. I added a new helper which preserves the max size check though

erik:
Hey @msbarry, just making sure this didn't skip through the cracks, any other changes you'd like to see here?

msbarry:
> Hey @msbarry, just making sure this didn't slip through the cracks, any other changes you'd like to see here?

Sorry for the delay! Looks good, just have a couple of minor tweaks to the tests, I'll try pushing them then land.

One thing I noticed is that it's 2-3x slower than natural earth reader for the natural earth geopackage file. Not a big deal in the grand scheme of things since it only takes ~30 seconds no matter the size of the extract, I'm just curious if there's any obvious reason for the slowdown...

erik:
@msbarry when I was first working on this I believe I noticed that the `NaturalEarthReader` class wasn't actually reading all of the possible features out of the sqlite DB (maybe due to missing SRIDs?). Unfortunately I don't remember which features weren't loading, but perhaps that could be the cause of the slowdown if they are now being read.

msbarry:
> @msbarry when I was first working on this I believe I noticed that the `NaturalEarthReader` class wasn't actually reading all of the possible features out of the sqlite DB (maybe due to missing SRIDs?). Unfortunately I don't remember which features weren't loading, but perhaps that could be the cause of the slowdown if they are now being read.

OK got it, that makes sense. One other thing is that it now unzips the zip file twice - once to get the count, then again to read the file. Another option would be send the unzipped file to a deterministic location, and keep it around between those.  @bdon  also pointed out that we could keep it around between runs to save from having to unzip it each time.

Issue #98, opened 2022-02-27
[FEATURE]  PMTiles output format


wipfli:
**Is your feature request related to a problem? Please describe.**

Currently, Planetiler outputs mbtiles which requires a tileserver, i.e., static file hosting like on github pages is not enough to serve tiles.

**Describe the solution you'd like**

[PMTiles](https://github.com/protomaps/PMTiles) can be used to serve vector tiles with range requests and they don't require a tile server. Planetiler could implement a pmtiles writer.

**Describe alternatives you've considered**

@bdon created a mbtiles to pmtiles converter:

```
pip install pmtiles
pmtiles-convert TILES.mbtiles TILES.pmtiles
```

**Additional context**
the powerlines example uses pmtiles https://github.com/wipfli/powerlines-switzerland

msbarry:
I think this makes sense, probably migrate the `--mbtiles=output.mbtiles` option to `--output=output.mbtiles` or `--output=output.pmtiles` and switch the writer implementation based on file type. It sounds like the format is pretty straightforward, but Brandon pointed out that it would be beneficial to change the tiles order to collocate nearby tiles. He recommended hilbert curve order, but tile pyramid order might satisfy a similar goal and be a bit easier to implement?

bdon:
Would it make sense to structure this as a separate Java library? If so, can that live alongside the python/js implementations at https://github.com/protomaps/PMTiles or should it live in its own Git repository?

msbarry:
@bdon a separate library would be nice, then the wrapper in planetiler would be pretty minimal.  At the simplest it would need an API like:

```
try (var pmtiles = new PMTiles(file)) {
  for (var tile : tiles) {
    pmtiles.writeTile(tile.x, tile.y, tile.z, tile.data);
  }
} // close() flushes the index leaves
```

But for performance optimizations it might make sense to expose the hashing function and an API to write a tile with a known hash as well so the writer could avoid hashing the same bytes over and over again.

msbarry:
Also @bdon could you elaborate on the tile ordering optimization? Is the main reason to put nearby tiles into the same index leaves? Planetiler packs tile x/y/z coordinates into a 32 bit integer that defines the order in which tiles are emitted, so I'd have to express a different ordering strategy as a different mapping from x/y/z to int.

bdon:
[...]

PR #443, opened 2023-01-13
Tile archive refactor

bdon:
This is a pure shuffling and renaming of classes to facilitate #98 but does not introduce any new archive formats yet.

The gist of it is extracting out `MbtilesWriter` into a new package `writer` and class `TileArchiveWriter`, and a new interface `TileArchive` implemented by `Mbtiles` and future archive formats.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -492,30 +492,29 @@ public class Planetiler {
   }
 
   /**
-   * Sets the location of the output {@code .mbtiles} file to write rendered tiles to. Fails if the file already exists.
+   * Sets the location of the output archive to write rendered tiles to. Fails if the archive already exists.
    * <p>
-   * To override the location of the file, set {@code argument=newpath.mbtiles} in the arguments.
+   * To override the location of the file, set {@code argument=newpath} in the arguments.
    *
    * @param argument the argument key to check for an override to {@code fallback}
    * @param fallback the fallback value if {@code argument} is not set in arguments
    * @return this runner instance for chaining
-   * @see MbtilesWriter
+   * @see TileArchiveWriter
    */
   public Planetiler setOutput(String argument, Path fallback) {
-    this.output = arguments.file(argument, "mbtiles output file", fallback);
+    this.output = arguments.file(argument, "output tile archive", fallback);
     return this;
   }
 
   /**
-   * Sets the location of the output {@code .mbtiles} file to write rendered tiles to. Overwrites file if it already
-   * exists.
+   * Sets the location of the output archive to write rendered tiles to. Overwrites file if it already exists.
    * <p>
-   * To override the location of the file, set {@code argument=newpath.mbtiles} in the arguments.
+   * To override the location of the file, set {@code argument=newpath} in the arguments.
    *
    * @param argument the argument key to check for an override to {@code fallback}
    * @param fallback the fallback value if {@code argument} is not set in arguments
    * @return this runner instance for chaining
-   * @see MbtilesWriter
+   * @see TileArchiveWriter
    */
   public Planetiler overwriteOutput(String argument, Path fallback) {
     this.overwrite = true;
@@ -524,7 +523,7 @@ public class Planetiler {
 
   /**
    * Reads all elements from all sourced that have been added, generates map features according to the profile, and
-   * writes the rendered tiles to the output mbtiles file.
+   * writes the rendered tiles to the output archive.
    *
    * @throws IllegalArgumentException if expected inputs have not been provided
    * @throws Exception                if an error occurs while processing
@@ -582,7 +581,7 @@ public class Planetiler {
         }
       }
       LOGGER.info("  sort: Sort rendered features by tile ID");
-      LOGGER.info("  mbtiles: Encode each tile and write to {}", output);
+      LOGGER.info("  archive: Encode each tile and write to {}", output);
     }
 
     // in case any temp files are left from a previous run...
@@ -619,7 +618,7 @@ public class Planetiler {
     stats.monitorFile("nodes", nodeDbPath);
     stats.monitorFile("features", featureDbPath);
     stats.monitorFile("multipolygons", multipolygonPath);
-    stats.monitorFile("mbtiles", output);
+    stats.monitorFile("archive", output);
 
     for (Stage stage : stages) {
       stage.task.run();
@@ -668,7 +667,7 @@ public class Planetiler {
     readPhase.addDisk(featureDbPath, featureSize, "temporary feature storage");
     writePhase.addDisk(featureDbPath, featureSize, "temporary feature storage");
     // output only needed during write phase
-    writePhase.addDisk(output, outputSize, "mbtiles output");
+    writePhase.addDisk(output, outputSize, "archive output");
     // if the user opts to remove an input source after reading to free up additional space for the output...
     for (var input : inputPaths) {
       if (input.freeAfterReading()) {
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -4,9 +4,8 @@ import com.onthegomap.planetiler.collection.FeatureGroup;
 import com.onthegomap.planetiler.collection.LongLongMap;
 import com.onthegomap.planetiler.collection.LongLongMultimap;
 import com.onthegomap.planetiler.config.Arguments;
-import com.onthegomap.planetiler.config.MbtilesMetadata;
 import com.onthegomap.planetiler.config.PlanetilerConfig;
-import com.onthegomap.planetiler.mbtiles.MbtilesWriter;
+import com.onthegomap.planetiler.mbtiles.Mbtiles;
 import com.onthegomap.planetiler.reader.GeoPackageReader;
 import com.onthegomap.planetiler.reader.NaturalEarthReader;
 import com.onthegomap.planetiler.reader.ShapefileReader;
@@ -27,6 +26,9 @@ import com.onthegomap.planetiler.util.ResourceUsage;
 import com.onthegomap.planetiler.util.Translations;
 import com.onthegomap.planetiler.util.Wikidata;
 import com.onthegomap.planetiler.worker.RunnableThatThrows;
+import com.onthegomap.planetiler.writer.TileArchiveMetadata;
+import com.onthegomap.planetiler.writer.TileArchiveWriter;
+import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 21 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -616,7 +619,7 @@ public void run() throws Exception {
     stats.monitorFile("nodes", nodeDbPath);
     stats.monitorFile("features", featureDbPath);
     stats.monitorFile("multipolygons", multipolygonPath);
-    stats.monitorFile("mbtiles", output);
+    stats.monitorFile("archive", output);
```
msbarry, line 622: I'm thinking we might want a third file output that just writes files to disk. Do you think "archive" is appropriate for that output as well? If not we might change to "output"

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -633,7 +636,12 @@ public void run() throws Exception {
 
     featureGroup.prepare();
 
-    MbtilesWriter.writeOutput(featureGroup, output, mbtilesMetadata, config, stats);
+    try (Mbtiles db = Mbtiles.newWriteToFileDatabase(output, config.compactDb())) {
```
msbarry, line 639: This is fine as-is since there's only one available output, but when we have more what do you think about ways to choose between the two? It could just go by file extension ".pmtiles" vs. ".mbtiles" but ultimately I think we'll want composable layers for input/output file types, encodings, and locations, for example:

- download from <s3/http/etc..>
- unwrap <zip/tar/tgz/bz2>
- read <shp/geopackage/flatgeobug/geojson/osm>
- then do planetiler stuff...
- output to <mbtiles/pmtiles>
- wrap <zip/tar/tgz/bz2>
- upload to <s3/http/...>

and each layer might have some parameters like s3 credentials/headers, mbtiles settings, gzip level, etc... I was thinking we could encode these as strings like `output.mbtiles?compact=false` or `input.shp.zip?src_proj=EPSG:3875` but not sure if you have any better ideas there?

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchive.java
```diff
@@ -0,0 +1,28 @@
...
+    void close();
+
+    default void printStats() {}
+  }
+
+  void initialize(PlanetilerConfig config, TileArchiveMetadata metadata, LayerStats layerStats);
+
+  TileWriter newTileWriter();
+
+  void finalize(PlanetilerConfig config);
```
msbarry, line 27: can you add javadoc explaining what these do for callers and implementers?

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchive.java
```diff
@@ -0,0 +1,28 @@
...
+import com.onthegomap.planetiler.config.PlanetilerConfig;
+import com.onthegomap.planetiler.util.LayerStats;
+
+/**
+ * A TileArchive is a portable on-disk stored tileset, e.g. MBTiles
+ */
+public interface TileArchive {
+
+  /**
+   * A high-throughput writer that accepts new tiles and queues up the writes to execute them in fewer large-batches.
```
msbarry, line 12: the performance/batching language is mbtiles-specific

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveMetadata.java
```diff
@@ -27,7 +27,7 @@ public MbtilesMetadata(Profile profile) {
     );
   }
 
-  public MbtilesMetadata(Profile profile, Arguments args) {
+  public TileArchiveMetadata(Profile profile, Arguments args) {
     this(
```
msbarry, line 31: looks like you need to update these to "tileset_" for the tests to pass. That's technically a breaking change since people might have scripts that set these - could you make then check `mbtiles_*` as a fallback?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -550,7 +552,8 @@ public void run() throws Exception {
       throw new IllegalArgumentException("Can only run once");
     }
     ran = true;
-    mbtilesMetadata = new MbtilesMetadata(profile, config.arguments());
+    tileArchiveMetadata = new TileArchiveMetadata(profile, config.arguments());
+    TileArchiveMetadata tileArchiveMetadata = new TileArchiveMetadata(profile, config.arguments());
```
msbarry, line 556: looks like you're constructing a second one here, causing tests to fail.

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java
```diff
@@ -145,6 +150,40 @@ public static Mbtiles newReadOnlyDatabase(Path path) {
     }
   }
 
+  public void initialize(PlanetilerConfig config, TileArchiveMetadata tileArchiveMetadata, LayerStats layerStats) {
```
msbarry, line 153: make sure you include `@Override` on any methods you override.

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveWriter.java
```diff
@@ -79,29 +77,19 @@ private MbtilesWriter(Iterable<FeatureGroup.TileFeatures> inputTiles, Mbtiles db
...
-    try (Mbtiles output = Mbtiles.newWriteToFileDatabase(outputPath, config.compactDb())) {
-      writeOutput(features, output, () -> FileUtils.fileSize(outputPath), mbtilesMetadata, config, stats);
-    } catch (IOException e) {
-      throw new IllegalStateException("Unable to write to " + outputPath, e);
-    }
+    stats.counter("tileset_tiles_written", "zoom", () -> countsByZoom);
   }
 
   /** Reads all {@code features}, encodes them in parallel, and writes to {@code output}. */
   public static void writeOutput(FeatureGroup features, Mbtiles output, DiskBacked fileSize,
```
msbarry, line 90: could you change `Mbtiles output` to `TileArchive output` ?

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveWriter.java
```diff
@@ -316,36 +304,12 @@ private void tileEncoder(Iterable<TileBatch> prev, Consumer<TileBatch> next) thr
...
-      .setType(mbtilesMetadata.type())
-      .setBoundsAndCenter(config.bounds().latLon())
-      .setMinzoom(config.minzoom())
-      .setMaxzoom(config.maxzoom())
-      .setJson(layerStats.getTileStats());
-
-    for (var entry : mbtilesMetadata.planetilerSpecific().entrySet()) {
-      metadata.setMetadata(entry.getKey(), entry.getValue());
-    }
+    db.initialize(config, tileArchiveMetadata, layerStats);
```
msbarry, line 307: could you rename `db` to `archive`  or something similar?

msbarry (commented):
Looking good! I pointed out the 2 issues causing the unit test failures, but after that we're going to run into a planetiler-openmaptiles compile failure (TileEncodingResult symbol not found), which is a pain due to the dependencies between the two projects... we could have one briefly broken, or try to make this change in a backwards compatible way, upgrade OMT, then remove the backwards-compatible part. Let me know if you have any better idea for dealing with that.

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
bdon: Yeah, I guess technically `TileArchive` is a misnomer for directory of tiles on disk but I can live with it... `output` and `tileset` felt too generic to me: e.g.: two tile archives in MBTiles and PMTiles format represent the same output tileset

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
bdon: I don't think string encoding is going to work for specifying S3 endpoints, because those are fully formed URLs and will need to have a level of escaping to fit in string query parameters like that. Most `TileArchive` formats are going to be compressed internally so adding another wrapper of compression won't be that effective (any counterexamples, other than a dirtiles archive format?)

I say we just switch based on the output extension for now

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchive.java
bdon: Maybe we should drop the `AutoCloseable` part of this interface? any cleanup that happens in `close` could be redundant with what's done in `finalize` though it's the caller's job to `finalize` - what vacuums the SQLite db or persists the PMTiles archive

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveMetadata.java
bdon: Instead of checking mbtiles_* as a fallback we could make planetiler explicitly exit until the user changes their configuration to `tileset_`? do we plan to keep the `mbtiles_*` around forever?

bdon: changed to leave as `mbtiles_` for now and keeping this PR an internal refactor; can change this next time we bump the user-facing API

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveWriter.java
bdon: I get a ` try-with-resources not applicable to variable type` - is this related to my JDK version?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
msbarry: yeah switch on extension makes sense to start with.

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchive.java
msbarry: I think we want `AutoCloseable` (or `Closeable` since it throws an IOException), it's fine now since we have to do this to use it:

```java
    try (Mbtiles archive = Mbtiles.newWriteToFileDatabase(output, config.compactDb())) {
      TileArchiveWriter.writeOutput(featureGroup, archive, () -> FileUtils.fileSize(output), tileArchiveMetadata,
        config, stats);
    } catch (IOException e) {
      throw new IllegalStateException("Unable to write to " + output, e);
    }
```

but when there's a pmtiles version, we'll want to use try-with-resources to ensure that it closes, but we'll also need to be working with a generic reference to a `TileArchive`:

```java
    try (TileArchive archive = createTileArchive(output, config.compactDb())) {
      TileArchiveWriter.writeOutput(featureGroup, archive, () -> FileUtils.fileSize(output), tileArchiveMetadata,
        config, stats);
    } catch (IOException e) {
      throw new IllegalStateException("Unable to write to " + output, e);
    }
```

which needs TileArchive to implement closeable/autocloseable to compile

msbarry (commented):
Looks good overall, a few minor comments then should be good to merge!

msbarry:
Looks great, thanks! Looking forward to native pmtiles output!

PR #455, opened 2023-01-23
FeatureGroup can emit in either TMS or Hilbert order. [#98]

bdon:
Another prerequisite for #98 that is not strictly required but PMTiles is much more efficient with clustered order. 

This isolates the ordering to `FeatureGroup` only; from the TileArchive side you're still dealing with `TileCoord` that is TMS order. This should also mean non-PMTiles archive formats should never touch the slower Hilbert codepaths.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
@@ -7,12 +7,12 @@ import com.onthegomap.planetiler.config.PlanetilerConfig;
 import com.onthegomap.planetiler.geo.GeometryException;
 import com.onthegomap.planetiler.geo.GeometryType;
 import com.onthegomap.planetiler.geo.TileCoord;
+import com.onthegomap.planetiler.geo.TileOrder;
 import com.onthegomap.planetiler.render.RenderedFeature;
 import com.onthegomap.planetiler.stats.Stats;
 import com.onthegomap.planetiler.util.CloseableConsumer;
 import com.onthegomap.planetiler.util.CommonStringEncoder;
 import com.onthegomap.planetiler.util.DiskBacked;
-import com.onthegomap.planetiler.util.Hilbert;
 import com.onthegomap.planetiler.util.LayerStats;
 import com.onthegomap.planetiler.worker.Worker;
 import java.io.Closeable;
@@ -63,16 +63,6 @@ public final class FeatureGroup implements Iterable<FeatureGroup.TileFeatures>, 
   private volatile boolean prepared = false;
   private final TileOrder tileOrder;
 
-  public enum TileOrder {
-    TMS(0),
-    HILBERT(1);
-
-    private final int number;
-
-    private TileOrder(int number) {
-      this.number = number;
-    }
-  }
 
   FeatureGroup(FeatureSort sorter, TileOrder tileOrder, Profile profile, Stats stats) {
     this.sorter = sorter;
@@ -216,17 +206,9 @@ public final class FeatureGroup implements Iterable<FeatureGroup.TileFeatures>, 
     var vectorTileFeature = feature.vectorTileFeature();
     byte encodedLayer = commonLayerStrings.encode(vectorTileFeature.layer());
 
-    int encoded;
-    TileCoord tileCoord = feature.tile();
-    if (this.tileOrder == TileOrder.HILBERT) {
-      encoded = TileCoord.startIndexForZoom(tileCoord.z()) +
-        Hilbert.hilbertXYToIndex(tileCoord.z(), tileCoord.x(), tileCoord.y());
-    } else {
-      encoded = tileCoord.encoded();
-    }
 
     return encodeKey(
-      encoded,
+      this.tileOrder.encode(feature.tile()),
       encodedLayer,
       feature.sortKey(),
       feature.group().isPresent()
@@ -377,13 +359,7 @@ public final class FeatureGroup implements Iterable<FeatureGroup.TileFeatures>, 
     private byte lastLayer = Byte.MAX_VALUE;
 
     private TileFeatures(int lastTileId) {
-      if (tileOrder == TileOrder.HILBERT) {
-        int z = TileCoord.zoomForIndex(lastTileId);
-        long xy = Hilbert.hilbertPositionToXY(z, lastTileId - TileCoord.startIndexForZoom(z));
-        this.tileCoord = TileCoord.ofXYZ(Hilbert.extractX(xy), Hilbert.extractY(xy), z);
-      } else {
-        this.tileCoord = TileCoord.decode(lastTileId);
-      }
+      this.tileCoord = tileOrder.decode(lastTileId);
     }
 
     private static void unscale(List<VectorTile.Feature> features) {
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
@@ -12,6 +12,7 @@ import com.onthegomap.planetiler.stats.Stats;
 import com.onthegomap.planetiler.util.CloseableConsumer;
 import com.onthegomap.planetiler.util.CommonStringEncoder;
 import com.onthegomap.planetiler.util.DiskBacked;
+import com.onthegomap.planetiler.util.Hilbert;
 import com.onthegomap.planetiler.util.LayerStats;
 import com.onthegomap.planetiler.worker.Worker;
 import java.io.Closeable;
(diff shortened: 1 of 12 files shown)


bdon:
Realistically speaking maybe TileOrder should just be a boolean Hilbert or TMS for now? This way is extensible for other archive formats to use column-major, Z-order ordering, but those would also require an implementation inside FeatureGroup.

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
```diff
@@ -60,27 +61,55 @@ public final class FeatureGroup implements Iterable<FeatureGroup.TileFeatures>,
...
+  public enum TileOrder {
+    TMS(0),
+    HILBERT(1);
+
+    private final int number;
+
+    private TileOrder(int number) {
+      this.number = number;
+    }
+  }
```
msbarry, line 75: You don't need to include the number in TileOrder definition - that gets added automatically and exposed as `TileOrder.TMS.ordinal()`.

One useful thing you can do with enums that I learned from [effective java](https://www.amazon.com/Effective-Java-Joshua-Bloch/dp/0134685997) is you can include a method definition in each enum value, so you could do something like:

```java
public enum TileOrder {
  TMS(TileCoord::encoded, TileCoord::decode),
  HILBERT(TileCoord::hilbertEncoded, TileCoord::hilbertDecode);

  private final ToIntFunction<TileCoord> encode;
  private final IntFunction<TileCoord> decode;

  private TileOrder(ToIntFunction<TileCoord> encode, IntFunction<TileCoord> decode) {
    this.encode = encode;
    this.decode = decode;
  }

  public int encode(TileCoord coord) {
    return encode.applyAsInt(coord);
  }

  public TileCoord decode(int encoded) {
    return decode.apply(encoded);
  }
}
```

Then you can use it later, so the details stay hidden from the consumer:

```java
int encoded = tileOrder.encode(feature.tile());
```

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
```diff
@@ -60,27 +61,55 @@ public final class FeatureGroup implements Iterable<FeatureGroup.TileFeatures>,
   private final Stats stats;
   private final LayerStats layerStats = new LayerStats();
   private volatile boolean prepared = false;
+  private final TileOrder tileOrder;
 
-  FeatureGroup(FeatureSort sorter, Profile profile, Stats stats) {
+  public enum TileOrder {
```
msbarry, line 66: What do you think about moving TileOrder into/alongside TileCoord? It seems closely bound to that even though it primarily gets used here.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java
```diff
@@ -38,11 +38,11 @@ public record TileCoord(int encoded, int x, int y, int z) implements Comparable<
     }
   }
 
-  private static int startIndexForZoom(int z) {
+  public static int startIndexForZoom(int z) {
     return ZOOM_START_INDEX[z];
   }
 
-  private static int zoomForIndex(int idx) {
+  public static int zoomForIndex(int idx) {
```
msbarry, line 45: I think we should just expose `hilbertEncode` and `hilbertDecode` here, that seems like more of a match for what consumers might need rather than exposing this internal api to get there.

msbarry:
> Realistically speaking maybe TileOrder should just be a boolean Hilbert or TMS for now? This way is extensible for other archive formats to use column-major, Z-order ordering, but those would also require an implementation inside FeatureGroup.

Yeah it seems to make sense for TileArchive to require implementations to override `TileOrder tileOrder();` to keep it closest to the implementations.  That will be slightly annoying thought because `Planetiler` will need to construct the tile archive before starting all of the previous stages to get the instance to pull a tile order off of. 

https://github.com/onthegomap/planetiler/blob/b97d9d0e795feefb811fb461e4bb18da45c6a36f/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java#L618C2-L645

Maybe that's good though to make planetiler fails fast if if it can't write to the output location before spending time processing the input sources 🤷‍♂️

bdon:
> Maybe that's good though to make planetiler fails fast if if it can't write to the output location before spending time processing the input sources 🤷‍♂️

A possible use case that isn't supported by any archive format yet is streaming to an object storage bucket; in that case we would definitely want to ensure the output HTTP stream is writable as soon as possible

bdon:
TODO: fix `progressOnLevel` and check for other places in `writer` where the TMS assumptions of TileCoord leak through.

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/FeatureGroup.java
bdon: moved to `geo.TileOrder` file

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchive.java
```diff
@@ -24,6 +25,11 @@ default void printStats() {}
   }
 
 
+  /**
+   * specify the preferred insertion order for this archive, e.g. TMS or HILBERT. See {@link TileOrder}.
```
msbarry, line 29: One neat thing you can do in javadoc too is link directly to class members:

```suggestion
   * Specify the preferred insertion order for this archive, e.g. {@link TileOrder#TMS} or {@link TileOrder#HILBERT}.
```

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java
```diff
@@ -93,6 +94,12 @@ public final class Mbtiles implements TileArchive {
   private PreparedStatement getTileStatement = null;
   private final boolean compactDb;
 
+  /** Inserts will be ordered to match the MBTiles index (TMS) */
```
msbarry, line 97: In general I don't put javadoc on inhereted methods since they inherit javadoc from the parent.

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
```diff
@@ -293,6 +296,45 @@ public List<VectorTile.Feature> postProcessLayerFeatures(String layer, int zoom,
       )), getFeatures());
   }
 
+  @Test
+  void testHilbertOrdering() {
```
msbarry, line 300: Can we do a matching TMS one?

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java
```diff
@@ -101,7 +108,7 @@
     return "{x=" + x + " y=" + y + " z=" + z + '}';
   }
 
-  public double progressOnLevel(TileExtents extents) {
+  public double progressOnLevel(TileOrder order, TileExtents extents) {
```
msbarry, line 111: For hilbert I think we want this to be an approximation like `encoded - startForZoom(z)` with a TODO about how we could make it exact later if desired.

msbarry (commented):
Looks good! Some minor comments, and an idea on progressForLevel...

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveWriter.java
```diff
@@ -194,7 +194,8 @@ private String getLastTileLogDetails() {
     } else {
       blurb = "%d/%d/%d (z%d %s) %s".formatted(
         lastTile.z(), lastTile.x(), lastTile.y(),
-        lastTile.z(), Format.defaultInstance().percent(lastTile.progressOnLevel(config.bounds().tileExtents())),
+        lastTile.z(),
+        Format.defaultInstance().percent(lastTile.progressOnLevel(archive.tileOrder(), config.bounds().tileExtents())),
```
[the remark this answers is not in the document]
bdon: How would I define a function type that takes `TileCoord,TileExtents`? `BiFunction<Double,TileCoord,TileExtents>` or similar?

bdon:
Is there a reason to keep `config.emitTilesInOrder())` as an Argument, instead of just making the default ordered and removing the non-ordered code path?

msbarry:
> Is there a reason to keep `config.emitTilesInOrder())` as an Argument, instead of just making the default ordered and removing the non-ordered code path?

It requires more memory and reduces parallelism to emit them in order,  but everyone's been taking that hit for a while and is OK with it so maybe there's no reason to keep.  I'd probably want to try building a planet mbtiles on a high-end (64+ cpu) machine with that set to true and false and see what the difference is.  There are also some outputs where tile ordering really won't matter, like raw files.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/TileCoord.java
```diff
@@ -101,11 +109,16 @@ public String toString() {
...
-    return 1d * (x - zoomBounds.minX()) / (zoomBounds.maxX() - zoomBounds.minX());
+  public double progressOnLevel(TileOrder order, TileExtents extents) {
+    if (order == TileOrder.TMS) {
+      // approximate percent complete within a bounding box by computing what % of the way through the columns we are
+      var zoomBounds = extents.getForZoom(z);
+      return 1d * (x - zoomBounds.minX()) / (zoomBounds.maxX() - zoomBounds.minX());
+    } else {
+      // this assumes extents is the whole world, so will be wrong when tiling a limited bbox
+      // we could improve this by making TileExtents store an exact set of all encoded coordinates.
+      return 1d * Hilbert.hilbertXYToIndex(this.z, this.x, this.y) / (2 << this.z);
```
msbarry, line 120: Shouldn't this be `(1 << (2 * this.z))` to get 1, 4, 16, 64, ... for the levels?

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
```diff
@@ -293,6 +296,84 @@ public List<VectorTile.Feature> postProcessLayerFeatures(String layer, int zoom,
...
+    features = new FeatureGroup(sorter, TileOrder.HILBERT, new Profile.NullProfile() {}, Stats.inMemory());
+    featureWriter = features.writerForThread();
+
+    // TMS tile IDs at zoom level 1:
+    // 2 4
+    // 1 3
```
msbarry, line 306:
```suggestion
    // 1 2
    // 4 3
```

planetiler-core/src/main/java/com/onthegomap/planetiler/writer/TileArchiveWriter.java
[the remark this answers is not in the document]
msbarry: `ToDoubleBiFunction<TileOrder, TileExtents>`

planetiler-core/src/test/java/com/onthegomap/planetiler/geo/TileCoordTest.java
```diff
@@ -68,7 +101,8 @@ void testTileSortOrderRespectZ() {
   })
   void testTileProgressOnLevel(int x, int y, int z, double p) {
```
msbarry: and maybe a progress on level for hilbert? that should catch the bug I pointed out

bdon: Added with z=2 test which fails correctly when bug above is present

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
bdon: Don't all X=0 come before X=1? unless I am misreading the TileCoord code...

msbarry: For hilbert I think it should be:

```java
    // Hilbert tile IDs at zoom level 1:
    // 1 4
    // 2 3
```

PR #456, opened 2023-01-24
refactor VectorLayers from mbtiles into util.LayerStats [#98]

bdon:
This logic is going to be reused by #98, and other archive formats in general may want to store metadata about vector layers in an MBTiles-like way, so I moved this into LayerStats itself. 

The serialization/JSON/Jackson bits are still left to the individual TileArchive packages.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/LayerStats.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/LayerStats.java
@@ -3,6 +3,7 @@ package com.onthegomap.planetiler.util;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.onthegomap.planetiler.mbtiles.Mbtiles;
 import com.onthegomap.planetiler.render.RenderedFeature;
+import com.onthegomap.planetiler.writer.TileArchive;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/util/LayerStats.java
```diff
@@ -1,19 +1,24 @@
...
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * Tracks the type and zoom range of vector tile attributes that have been emitted by layer to populate the {@code json}
- * attribute in the mbtiles output metadata.
+ * Tracks the type and zoom range of vector tile attributes that have been emitted by layer to populate the
+ * {@code vector_layers} attribute in the archive output metadata. Matches the MBTiles spec, but can be reused by other
```
msbarry, line 20: For javadoc, the first line is special and gets used as a short description where the full javadoc is too much, so the convention is to do:

```java
/**
 * First sentence with short description of the method.
 * <p>
 * More details...
 */
```

For example BufferedReader.read's full javadoc generates this:
image


But that first line shows up in a summary table at the top:

image

msbarry (commented):
Looks good, one minor javadoc comment. Are you going to want the same json-serialization of the layers? The jackson annotations on the record you extracted will enforce that, but if you want a different serialization then it will need to be mapped to a new json serialization object.

bdon:
> Are you going to want the same json-serialization of the layers? The jackson annotations on the record you extracted will enforce that, but if you want a different serialization then it will need to be mapped to a new json serialization object.

No, it's a slightly different serialization since the `name`, `attribution` etc are siblings of `vector_layers` instead of having a confusing `json` top-level key inside the JSON.

here is how I plan to define it in a PMTiles specific way:

```java
  public record JsonMetadata(
    @JsonProperty("vector_layers") List<LayerStats.VectorLayer> vectorLayers,
    @JsonAnyGetter Map<String, String> otherMetadata
  ) {
```

and have `TileArchiveMetadata` return a Map<String,String> with all metadata...

Issue #237, opened 2022-05-29
[BUG] housenumber wrongly placed.


farfromrefug:
This is osm
<img width="253" alt="Screenshot 2022-05-29 at 17 28 53" src="https://user-images.githubusercontent.com/655344/170877510-de199b9b-705b-435d-8246-714d93bab5ec.png">

This is omt
<img width="529" alt="Screenshot 2022-05-29 at 17 25 32" src="https://user-images.githubusercontent.com/655344/170877359-9e8f0f4e-04ea-4549-b4d1-d73b91db4977.png">

This is PT
<img width="531" alt="Screenshot 2022-05-29 at 17 25 48" src="https://user-images.githubusercontent.com/655344/170877372-b06a1a16-1387-4f3b-a332-51ce88326798.png">

As you can see the housenumber is wrongly place in PT . And you actually end up not knowing on which side of the building it should be. I understand why it is there because of `centroidIfConvex`.

msbarry:
Thanks for pointing this out - looks like `isConvex` is actually returning a result for `isConcave`.  I'll put out a quick fix for it.

PR #458, opened 2023-01-26
Revert "fix is convex negated (#241)"

erik:
This reverts commit f93e5221f8a576bbce108385a7c0387216f145ba.

The original name of this function was correct, it was returning `true` when given a convex polygon and `false` given a concave polygon. This is confirmed by the test cases in `GeoUtilsTest`. 

The issue described in #237 is likely still present, perhaps due to failing when given self intersecting polygons.

Some more context in [this slack thread](https://osmus.slack.com/archives/C031V9E9RMG/p1674677909284199)

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
@@ -403,20 +403,11 @@ public class GeoUtils {
     return JTS_FACTORY.createPolygon(exteriorRing, rings.toArray(LinearRing[]::new));
   }
 
-
-  /**
-   * Returns {@code false} if the signed area of the triangle formed by 3 sequential points changes sign anywhere along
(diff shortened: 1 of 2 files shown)


msbarry:
OK so reverting this is correct, but the root cause of #237 is this line that omits any triangles where the area < 1e-10:

https://github.com/onthegomap/planetiler/blob/a0f8c67c7828bfce9316535a02d18e64a27ef79e/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java#L451-L455

Since we fit the entire planet into 0->1 coordinates, the minimum threshold should be more like 1e-20 to cut of anything less than ~1 square inch. Want to make that in this PR or separate?

erik:
IMO let's keep the changes separate.

I think ideally it's at least somewhat zoom dependent, since you might care a lot about a polygon being concave by a meter for a building, but for a country it doesn't matter at all

PR #459, opened 2023-01-26
Improve is convex logic

msbarry:
Better fix for #237 by adapting the threshold at which we consider a polygon concave (and throw out noise due to rounding errors) based on the size of the polygon. Instead of returning false for isConvex as soon as the sign of sequential triangles change, track the minimum negative sign and maximum positive sign and compare the magnitudes once we have enough points.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
@@ -408,8 +408,11 @@ public class GeoUtils {
    * {@code ring}, ignoring repeated and collinear points.
    */
   public static boolean isConvex(LinearRing ring) {
+    double threshold = 1e-3;
+    double minPointsToCheck = 10;
     CoordinateSequence seq = ring.getCoordinateSequence();
-    if (seq.size() <= 3) {
+    int size = seq.size();
+    if (size <= 3) {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


msbarry:
image

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
```diff
@@ -429,37 +432,47 @@ public static boolean isConvex(LinearRing ring) {
...
+      // look for sign changes in the triangles formed by sequential points
+      // but, we want to allow for rounding errors and small concavities relative to the overall shape
+      // so track the largest positive and negative threshold for triangle area and compare them once we
+      // have enough points
+      boolean check = i >= minPointsToCheck;
+      if (z < 0 && absZ > negZ) {
+        negZ = absZ;
+      } else if (z > 0 && absZ > posZ) {
+        posZ = absZ;
+      } else if (i > minPointsToCheck) { // always check at i=minPointsToCheck
```
erik, line 459: Since check isn't being used before being reassigned here, couldn't the declaration be simplified to `boolean check = i == minPointsToCheck`? Doesn't seem like intended behavior

msbarry: Once we've seen over the minimum number of points, we do want to check any time we extend `negZ` or `posZ` - in that case we'll hit the first or second case and not this one.

erik: Update makes this much clearer 👍

PR #463, opened 2023-01-28
Convert get all tile coords to iterator

msbarry:
Pull mbtiles access methods up to `TileArchive` and change signature for listing tile coordinates to return a `CloseableIterator` so you can iterate through all coordinates in the archive without pulling the whole list into memory.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
@@ -29,7 +29,7 @@ public interface ReadableTileArchive extends Closeable {
    *
    * <pre>
    * {@code
-   * try (var iter = archive.getAllTileCoord()) {
+   * try (var iter = archive.getAllTileCoords()) {
    *   while (iter.hasNext()) {
    *     var coord = iter.next();
    *     ...
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
@@ -1,0 +1,44 @@
+package com.onthegomap.planetiler.archive;
+
+import com.onthegomap.planetiler.geo.TileCoord;
+import com.onthegomap.planetiler.util.CloseableIterator;
+import java.io.Closeable;
+
+/**
+ * Read API for on-disk representation of a tileset in a portable format. Example: MBTiles, a sqlite-based archive
+ * format.
+ * <p>
+ * See {@link WriteableTileArchive} for the write API.
+ */
+public interface ReadableTileArchive extends Closeable {
+
+  /** Returns the raw tile data associated with the tile at {@code coord}. */
+  default byte[] getTile(TileCoord coord) {
+    return getTile(coord.x(), coord.y(), coord.z());
+  }
+
+  /** Returns the raw tile data associated with the tile at coordinate {@code x, y, z}. */
+  byte[] getTile(int x, int y, int z);
+
+  /**
+   * Returns an iterator over the coordinates of tiles in this archive.
+   * <p>
+   * The order should respect {@link WriteableTileArchive#tileOrder()} of the corresponding writer.
+   * <p>
+   * Clients should be sure to close the iterator after iterating through it, for example:
+   *
+   * <pre>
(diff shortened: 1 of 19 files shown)


bdon:
A couple thoughts:

* Is it possible to make the reading interfaces a separate interface from the writing interfaces? Combined makes sense for an archive format backed by a transactional DB, but for #98 I'm considering implementing reader and writer as wholly separate classes because there isn't much code they would share in common.
* How should a `getAllCoords` respect tile order? Does it assume iteration through the set is via `tileOrder()` ? In Tippecanoe we have the use case of using `tile-join` to merge two tilesets; thus it's required that all iterators over archives are in TMS order to do this efficiently. 
* I think there are memory concerns with materializing the entire TileCoord + Tile `offset, length` in memory for PMTiles, but this isn't necessarily relevant for MBTiles. Also, if we only care about materializing a set of TileCoords and don't care about `offset, length`, we could store this much more compactly using a `Roaring32<TileCoord>` instead of a `List<TileCoord>` and not really have a need for a memory-efficient iterator. We need to consider if Planetiler will ever deal with zoom >=16 tilesets possibly as input sources, though.

msbarry:
> * Is it possible to make the reading interfaces a separate interface from the writing interfaces? Combined makes sense for an archive format backed by a transactional DB, but for [[FEATURE]  PMTiles output format #98](https://github.com/onthegomap/planetiler/issues/98) I'm considering implementing reader and writer as wholly separate classes because there isn't much code they would share in common.

That could possibly work - where would something like tileOrder go though?

> * How should a `getAllCoords` respect tile order? Does it assume iteration through the set is via `tileOrder()` ? In Tippecanoe we have the use case of using `tile-join` to merge two tilesets; thus it's required that all iterators over archives are in TMS order to do this efficiently.

I don't see a need for it to return tiles in a specific order, I'd expect it to respect `tileOrder` for this archive type.

> * I think there are memory concerns with materializing the entire TileCoord + Tile `offset, length` in memory for PMTiles, but this isn't necessarily relevant for MBTiles. Also, if we only care about materializing a set of TileCoords and don't care about `offset, length`, we could store this much more compactly using a `Roaring32<TileCoord>` instead of a `List<TileCoord>` and not really have a need for a memory-efficient iterator. We need to consider if Planetiler will ever deal with zoom >=16 tilesets possibly as input sources, though.

That seems interesting if we needed to check tile IDs in random order, but I think a memory-efficient iterator would be simplest to implement. We could always layer something like that on top of an efficient iterator if the need arises.

bdon:
Overall LGTM, if the main use case is for unit/integration tests then getting a total Set of all TileCoords is easier for pmtiles reading than the memory-efficient iterator. 

Iterating over coordinates and tile data can be awkward because of deduplication in both mbtiles/pmtiles as well as RLE in pmtiles, potentially we're feeding a consumer of that iterator a lot of extra work, unless we exposed not just `byte[]` but also an opaque 64(128?)-bit content hash and made loading the `byte[]` lazy?

msbarry:
OK let's start with this for now, at the very least any tile archive should be able to serve the getTile API and most likely listTiles as well, but if it ends up being problematic we could push the list API down into mbtiles and have a separate list API for different implementations. I could maybe see some key/value store supporting random access but not the ability to list tiles.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface ReadableTileArchive extends Closeable {
+
+  /** Returns the raw tile data associated with the tile at {@code coord}. */
+  default byte[] getTile(TileCoord coord) {
+    return getTile(coord.x(), coord.y(), coord.z());
+  }
+
+  /** Returns the raw tile data associated with the tile at coordinate {@code x, y, z}. */
+  byte[] getTile(int x, int y, int z);
```
bdon, line 21: How should this handle the case of the tile not existing? should it be an option return type? It doesn't seem exceptional.

msbarry: Ah sorry the javadoc should indicate this, but it should return `null` if not found.

msbarry: Opened #486 to fix the javadoc.

PR #466, opened 2023-01-30
Pmtiles package low level header and directory serialization/deserialization [#98]

bdon:
Some utility functions but they need to be efficient since the typical `entries` size will be in the tens or hundreds of millions

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/Pmtiles.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/Pmtiles.java
@@ -1,0 +1,259 @@
+package com.onthegomap.planetiler.pmtiles;
+
+import com.carrotsearch.hppc.ByteArrayList;
+import com.carrotsearch.hppc.ObjectArrayList;
+import com.onthegomap.planetiler.util.VarInt;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+
+/**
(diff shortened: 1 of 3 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/Pmtiles.java
```diff
@@ -0,0 +1,332 @@
...
+  public static byte[] serializeDirectory(List<Entry> slice, int start, int end) {
+    ByteArrayList dir = new ByteArrayList();
+
+    VarInt.putVarLong((long) end - start, dir);
+
+    long lastId = 0;
+    for (int i = start; i < end; i++) {
+      VarInt.putVarLong(slice.get(i).tileId - lastId, dir);
+      lastId = slice.get(i).tileId;
+    }
```
msbarry, line 281: Since `slice` could be any kind of list here (i.e. linked list), the only other possible tweak here would be to extract a `subList` if start/end aren't the full size of slice, then iterate through it for each of these loops, something like:

```suggestion
    List<Entry> range = start == 0 && end == slice.size() ? slice : slice.subList(start, end);
    long lastId = 0;
    for (var entry : range) {
      VarInt.putVarLong(range.tileId - lastId, dir);
      lastId = range.tileId;
    }
```

That should make it efficient for any list implementation - although a non-arraylist is unlikely.

msbarry: Also, should there be a `serializeDirectory(slice)` convenience method that does the entire list? the range one could construct the sublist and pass it in to avoid the conditional check on start/end size?

PR #499, opened 2023-02-23
Detect certain kinds of invalid polygons when slicing into tiles

### LINKED ISSUE #496: [BUG] setting pixel tolerance too high can break multipolygons
**Describe the bug**

Using a combination of `Feature.setPixelTolerance(0.5)` and `Feature.setMinPixelSize(0.1)` broke the geometry for [this relation](https://www.openstreetmap.org/relation/10041968), which caused the entire tile to be filled at z11.

Screen Shot 2023-02-22 at 10 21 17

If you look at the relation, the sand and grass polygons are super close to each other (but not touching), so it seems the combination of calls resulted in an invalid simplification.

grafik

In my case, I tested that I can fix the tile by modifying the pixel size and tolerance values, but it seems possible for other multipolygons to have the same issue. The OpenMapTiles profile for example uses [the same sequence of calls](https://github.com/openmaptiles/planetiler-openmaptiles/blob/1cae8c4bdf99f66b616bdcd7d49342cb991a8d48/src/main/java/org/openmaptiles/layers/Building.java#L172-L173), and could conceivably run into this.

These changes all independently fixed the issue for me:

1. Remove the call to `setMinPixelSize`
2. Remove the call to `setPixelTolerance`
3. Change the pixel tolerance to `0.25`

**To Reproduce**

- GeoFabrik's `new-jersey` extract.
- Create a layer which selects `natural=sand` polygons.
- Use the following profile:

```java
    features
      .polygon("test")
      .setPixelTolerance(0.5)
      .setMinPixelSize(0.1)
      .setAttr("id", source.id())
```

- Relevant area: `11/40.9405/-74.0947`

Commenting out either of `setPixelTolerance` or `setMinPixelSize` will fix the issue

**Environment (please complete the following information):**

- Hardware: 2019 Macbook Pro
- OS: MacOS 12.13.1
- Java version and distribution: openjdk-17
- Maven version: 3.8.2

msbarry:
Fix #496 by detecting the case where a inner polygon ring (hole) intersects a tile not contained by or touching an outer ring.  Without this check, feature renderer assumes the hole is taking a chunk out of a larger filled area so fills the entire tile first.  `FeatureRenderer` detects this case, fixes the polygon, and tries again.


planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1626,20 +1644,42 @@ void testBadRelation() throws Exception {
         }, member.ref(), member.role()));
       }
     }
+    return elements;
+  }
+
+  @Test
+  void testIssue496BaseballMultipolygon() throws Exception {
+    // this generated a polygon that covered an entire z11 tile where the buffer interested the baseball field
```
erik, line 1652:
```suggestion
    // this generated a polygon that covered an entire z11 tile where the buffer intersected the baseball field
```

erik (approved):
Approach makes sense to me, seems like the fix should prevent this kind of issue. I can't immediately think of another way the inner ring could break that wouldn't be handled here

PR #502, opened 2023-02-28
Finish PMTiles write implementation and basic reader

bdon:
* Reader is memory-inefficient, it loads entire dir into memory. Only really practical for unit testing. It's also limited to z15 because of the planetiler 32-bit IDs.
* follow up to this needs to expose multiple output formats via API


planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,161 @@
...
+            var buf2 = ByteBuffer.allocate(entry.length());
+            this.channel.position(header.tileDataOffset() + entry.offset());
+            this.channel.read(buf2);
+            return buf2.array();
+          } else {
+            dirOffset = header.leafDirectoriesOffset() + entry.offset();
+            dirLength = entry.length();
+          }
+        } else {
+          return null;
```
bdon, line 80: Not sure how to make Sonar accept that these should return null if not found

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,161 @@
...
+
+      }
+    } catch (IOException e) {
+      throw new IllegalStateException("Could not get tile", e);
+    }
+
+    return null;
+  }
+
+  public WriteablePmtiles.Header getHeader() throws IOException {
```
bdon, line 91: I left most of the general purpose classes like `Header` in `WriteablePmtiles`, do we need to extract all those out into a Pmtiles.Util` package?

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public TileCoord next() {
+      return iter.next();
+    }
+
+    private void collectTileCoords(List<TileCoord> l, WriteablePmtiles.Header header,
+      long dirOffset, int dirLength) throws IOException {
+
+      channel.position(dirOffset);
```
bdon, line 119: pretty sure this is both inefficient and thread-unsafe, but the scope of `ReadablePmtiles` is just for tests right now

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -96,44 +110,168 @@ void testRoundtripHeader() {
...
+
+  }
+
+  @Test
+  void testWritePmtilesSingleEntry() throws IOException {
+    var bytes = new SeekableInMemoryByteChannel(0);
+    var in = WriteablePmtiles.newWriteToMemory(bytes);
+
+    var config = PlanetilerConfig.defaults();
+    in.initialize(config, new TileArchiveMetadata(new Profile.NullProfile()), new LayerStats());
```
bdon, line 171: This is verbose to set up a test, since PMTiles depends on an entire profile and metadata to write out the JSON metadata and full header; this isn't true of MBTiles since you can read/write to the `tiles` tables in isolation

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,616 @@
...
+  /**
+   * Serialize all entries into bytes, adaptively choosing the # of leaf directories to ensure the header+root fits in
+   * 16 KB.
+   *
+   * @param entries a sorted ObjectArrayList of all entries in the tileset.
+   * @return byte arrays of the root and all leaf directories, and the # of leaves.
+   * @throws IOException
+   */
+  public static Directories makeRootLeaves(List<Entry> entries) throws IOException {
+    if (entries.size() < 16384) { // coarse heuristic to short circuit
```
bdon, line 427: There are a lot of magic numbers here

`16384`, `3500`, `4096`, `1.2`

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,616 @@
...
+      out.position(0);
+      out.write(ByteBuffer.wrap(header.toBytes()));
+      out.write(ByteBuffer.wrap(archiveDirs.root));
+
+      Format format = Format.defaultInstance();
+
+      LOGGER.info("# addressed tiles: {}", numAddressedTiles);
+      LOGGER.info("# of tile entries: {}", entries.size());
+      LOGGER.info("# of tile contents: {}", (hashToOffset.size() + numUnhashedTiles));
+      LOGGER.info("Root directory: {}B", format.storage(archiveDirs.root.length, false));
```
bdon, line 524: Not sure how to make sonar happy... it wants `format.storage` to not be evaluated if the log level doesn't reach `info`

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,619 @@
...
+        end = subEntries.size();
+      }
+      byte[] leafBytes = serializeDirectory(subEntries, i, end);
+      leafBytes = Gzip.gzip(leafBytes);
+      rootEntries.add(new Entry(subEntries.get(i).tileId, leavesLength, leafBytes.length, 0));
+      leavesOutputStream.add(leafBytes);
+      leavesLength += leafBytes.length;
+    }
+
+    byte[] rootBytes = serializeDirectory(rootEntries, 0, rootEntries.size());
```
msbarry, line 416: nit: just `serializeDirectory(rootEntries)` ?

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,619 @@
...
+      if (temp.root.length < INIT_SECTION - HEADER_LEN) {
+        return temp;
+      }
+      leafSize *= 1.2;
+    }
+  }
+
+  private WriteablePmtiles(SeekableByteChannel channel) throws IOException {
+    this.out = channel;
+    out.write(ByteBuffer.wrap(new byte[INIT_SECTION]));
```
msbarry, line 452: You can simplify this a bit to `out.write(ByteBuffer.allocate(INIT_SECTION))`

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -96,44 +113,178 @@ void testRoundtripHeader() {
...
+
+  @Test
+  void testBuildDirectoriesRootOnly() throws IOException {
+
+  }
+
+  @Test
+  void testBuildDirectoriesLeaves() throws IOException {
+
+  }
```
msbarry, line 166: empty tests?

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
msbarry: You could use the constructor that takes a name, description, attribution, version, strings, then make sure the values you read out of the readable archive match the ones you passed in?

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -96,44 +113,178 @@ void testRoundtripHeader() {
...
+        OptionalLong.empty()));
+    }
+
+    in.finish(config);
+    var reader = new ReadablePmtiles(bytes);
+    var header = reader.getHeader();
+    assertEquals(ENTRIES, header.numAddressedTiles());
+    assertEquals(ENTRIES, header.numTileContents());
+    assertEquals(ENTRIES, header.numTileEntries());
+    assert (header.leafDirectoriesLength() > 0);
```
msbarry, line 279: probably want to use `assertTrue(header.leafDirectoriesLength() > 0);` just in case the tests are ever run without `-ea`

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -96,44 +113,178 @@ void testRoundtripHeader() {
...
+
+    for (int i = 0; i < ENTRIES; i++) {
+      var coord = TileCoord.hilbertDecode(i);
+      assertArrayEquals(ByteBuffer.allocate(4).putInt(i).array(), reader.getTile(coord.x(), coord.y(), coord.z()));
```
msbarry, line 283: might want to add a message as the third argument to this with `coord.toString()` so if this fails we'll be able to tell what tile caused the issue.

msbarry (commented):
Looking good!

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,619 @@
...
+  }
+
+  @Override
+  public void initialize(PlanetilerConfig config, TileArchiveMetadata tileArchiveMetadata, LayerStats layerStats) {
+    this.layerStats = layerStats;
+    this.tileArchiveMetadata = tileArchiveMetadata;
+  }
+
+  @Override
+  public void finish(PlanetilerConfig config) {
```
msbarry, line 476: I might be interested in seeing some logs for some of these steps, like sorting, each round of the adaptive leaf sizing, then writing to the end/beginning.  We don't need super detailed timing - just a log saying the step has started will capture the timestamp to give us a sense for where the time goes.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
bdon: created new `Pmtiles` common class

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,163 @@
...
+    this.channel = channel;
+  }
+
+  /**
+   * Finds the relevant entry for a tileId in a list of entries.
+   * <p>
+   * If there is an exact match for tileId, return that. Else if the tileId matches an entry's tileId + runLength,
+   * return that. Else if the preceding entry is a directory (runLength = 0), return that. Else return null.
+   */
+  public static WriteablePmtiles.Entry findTile(List<WriteablePmtiles.Entry> entries, long tileId) {
```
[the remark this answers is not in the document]
bdon: I prefer `List<Entry>` because it makes it clear that a Directory holds no other extra information apart from its entry contents. Refactoring `WriteablePmtiles.Entry` to `Pmtiles.Entry` makes it a little less verbose.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
bdon: added more logging to `finish`

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
bdon: Yeah, trying to read the JsonMetadata back out for tests, but this throws a runtime exception:

```java
  public record JsonMetadata(
    @JsonProperty("vector_layers") List<LayerStats.VectorLayer> vectorLayers,
    @JsonAnyGetter @JsonAnySetter Map<String, String> otherMetadata
  ) {

    public byte[] toBytes() {

      try {
        return objectMapper.writeValueAsBytes(this);
      } catch (JsonProcessingException e) {
        throw new IllegalArgumentException("Unable to encode as string: " + this, e);
      }
    }

    public static JsonMetadata fromBytes(byte[] bytes) {
      try {
        return objectMapper.readValue(bytes, JsonMetadata.class);
      } catch (IOException e) {
        throw new IllegalStateException("Invalid metadata json: " + bytes, e);
      }
    }
  }
```

I'm letting Jackson serialize arbitrary string->string, and need to get it to behave the same way in the opposite direction

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    return null;
+  }
+
+  public WriteablePmtiles.Header getHeader() throws IOException {
+    this.channel.position(0);
+    var buf = ByteBuffer.allocate(127);
+    this.channel.read(buf);
+    return WriteablePmtiles.Header.fromBytes(buf.array());
```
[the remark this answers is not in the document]
bdon: moved to ctor

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -1,12 +1,29 @@
...
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.OptionalLong;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 class PmtilesTest {
```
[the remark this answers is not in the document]
bdon: added a test archive created by tippecanoe

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
bdon: Intending to add some more edge cases here around the adaptive directory sizing

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,619 @@
...
+    int leafSize = (int) Math.max(entries.size() / 3_500d, 4096);
+
+    while (true) {
+      Directories temp = buildRootLeaves(entries, leafSize);
+      if (temp.root.length < INIT_SECTION - HEADER_LEN) {
+        return temp;
+      }
+      leafSize *= 1.2;
+    }
+  }
```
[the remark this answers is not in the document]
bdon: I don't think so, the `leafSize` will grow towards infinity, possibly slowly. If we used a numerical-ish method to attempt to converge on exactly 16KB header, there would be some risk of that. The current strategy is to use some rough numbers based on typical real data to get around 16KB on first attempt, and then refine via `1.2` till we get under. 

The overall purpose here is to avoid situations where we have undersized roots like 8kB; we could ~double the # of leaves and make each individual leaf smaller and faster to fetch, while not affecting the overall latency since we blindly fetch 16KB in most clients anyway.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
bdon: does just making the method `getBytes` `synchronized` work for now?

Would changing to AsynchronousFileChannel require changing the signatures of the `ReadableTileArchive` interface?

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
msbarry: Looks like this is because of https://github.com/FasterXML/jackson-databind/issues/3439 - there is a workaround described there where you just add an extra constructor with annotations to make it visible to jackson:

```java
    @JsonCreator
    JsonMetadata(@JsonProperty("vector_layers") List<LayerStats.VectorLayer> vectorLayers) {
      this(vectorLayers, new HashMap<>());
    }
```

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
msbarry: Yes, making `getBytes` synchronized works for now.  It might make sense to add a `getTileFuture` API at some point but that could be done in a separate change. I was playing around with an HTTP readable archive which makes the future-based API more important.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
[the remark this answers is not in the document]
msbarry: Ok sounds fine - it might just be nice to have the utilities for working with/parsing/serializing directories collected to a common place.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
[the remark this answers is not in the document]
bdon: A bit difficult to unit test the adaptive sizing because those specific parameters are cherry-picked out of real-world data = how compressible the directories are; we can refine those parameters later based on real datasets

planetiler-core/src/main/java/com/onthegomap/planetiler/util/SeekableInMemoryByteChannel.java
```diff
@@ -0,0 +1,221 @@
...
+ * {@link #truncate truncate} to a value bigger than that. Internal buffer can be accessed via
+ * {@link SeekableInMemoryByteChannel#array()}. Adapted from <a href=
+ * "https://commons.apache.org/proper/commons-compress/jacoco/org.apache.commons.compress.utils/SeekableInMemoryByteChannel.java.html">Apache
+ * Commons Compress</a>.
+ * </p>
+ *
+ * @since 1.13
+ * @NotThreadSafe
+ */
+public class SeekableInMemoryByteChannel implements SeekableByteChannel {
```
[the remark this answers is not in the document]
bdon: I think this is the lowest code footprint for now, any more sophisticated interface should probably work around doing multipart uploads like we discussed in #98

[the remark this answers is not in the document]
bdon: added to `NOTICE.md`

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/WriteablePmtiles.java
```diff
@@ -0,0 +1,320 @@
...
+        ", numLeaves=" + numLeaves +
+        ", leafSize=" + leafSize +
+        ", numAttempts=" + numAttempts +
+        '}';
+    }
+  }
+
+  private static Directories makeDirectoriesWithLeaves(List<Pmtiles.Entry> subEntries, int leafSize, int attemptNum)
+    throws IOException {
+    LOGGER.info("Building directories with {} entries per leaf...", leafSize);
```
msbarry, line 89: minor - include attempt # ?

planetiler-core/src/test/java/com/onthegomap/planetiler/pmtiles/PmtilesTest.java
```diff
@@ -133,7 +149,175 @@ void testRoundtripDirectorySlice() {
...
+    assertNull(reader.getTile(0, 0, 0));
+    assertNull(reader.getTile(0, 0, 2));
+
+    Set<TileCoord> coordset = reader.getAllTileCoords().stream().collect(Collectors.toSet());
+    assertEquals(1, coordset.size());
+  }
+
+  @Test
+  void testWritePmtilesToFileWithMetadata(@TempDir Path tempDir) throws IOException {
+    var in = WriteablePmtiles.newWriteToFile(tempDir.resolve("tmp.pmtiles"));
```
msbarry, line 203: Even though this is a test, might still be good to put this in a try-with-resources block to close in when done

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/Pmtiles.java
```diff
@@ -324,4 +333,41 @@ public static List<Entry> deserializeDirectory(byte[] bytes) {
...
+    @JsonProperty("vector_layers") List<LayerStats.VectorLayer> vectorLayers,
+    @JsonAnyGetter @JsonAnySetter Map<String, String> otherMetadata
+  ) {
+
+    @JsonCreator
+    JsonMetadata(@JsonProperty("vector_layers") List<LayerStats.VectorLayer> vectorLayers) {
+      this(vectorLayers, new HashMap<>());
+    }
+
+    public byte[] toBytes() {
```
msbarry, line 356: The tests will catch that we can deserialize what we serialize, but will they confirm this actually hoists the other metadata attributes to top-level? If now you could try deserializing the bytes as a map and confirm it has keys from otherMetadata.

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -0,0 +1,167 @@
...
+      return this.iterator.next();
+    }
+  }
+
+  private List<Pmtiles.Entry> readDir(long offset, int length) throws IOException {
+    var buf = getBytes(offset, length);
+    if (header.internalCompression() == Pmtiles.Compression.GZIP) {
+      buf = Gzip.gunzip(buf);
+    }
+    return Pmtiles.directoryFromBytes(buf);
```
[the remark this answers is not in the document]
bdon: I don't forsee needing to read/write beyond `gzip` until brotli or zstd are supported directly via `DecompressionStream` in browsers

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/Pmtiles.java
bdon: https://github.com/onthegomap/planetiler/pull/502/commits/f03bc180aa5b3a873de59c19980a162413aef0e7

msbarry:
Looks great! 🎉

Issue #377, opened 2022-10-30
[FEATURE] Include node/way/relation in the feature ID


jameswestman:
**Is your feature request related to a problem? Please describe.**

As far as I can tell, for OpenStreetMap features, Planetiler uses the OSM ID for the MVT feature ID. However, it doesn't say whether it's a node, way, or relation, which is a problem because the IDs are not unique between those, i.e. there can be both a node #1000 and a way #1000.

**Describe the solution you'd like**
I'd like to have some kind of encoding in the feature ID to describe the source of the ID--OSM node, OSM way, Natural Earth, etc. This could be done, for example, using the lower 3 bits of the ID.

**Describe alternatives you've considered**

I tried patching the OpenMapTiles schema repo to include the OSM ID with type in a tag, but this incurs a heavy tile size increase (around 10% on the small extract I tried) because they're all mostly different.

phanecak-maptiler:
I'll write what I've learned so far while troubleshooting #215 in regards to IDs , @msbarry will hopefully correct my mistakes and give some further feedback:

- all IDs in the result (i.e. .mbtiles file) are allocated with `AtomicLong idGenerator` (calling `incrementAndGet()`)
- how that relates to your feature request: AFAIK source IDs usually do not make it into the result
  - one exception I've stumbled upon so far: Natural Earth features should have `ne_id` tag thanks to https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/NaturalEarthReader.java#L172
- how that relates to #215: `idGenerator` is (also) called in multi-threaded stages, hence we're getting different IDs if generating same desired result twice from same source data
- TL;DR: if we solve this in a certain way (maintaining source IDs, etc.), it should also help with #215

bdon:
We can expose a `setId` method on `Feature`, alongside the existing `getSourceId`. That would provide a Profile-facing API to set the MVT id, but to make it useful we would also need to:

* Expose whether something is an OSM node, way or relation through `SourceFeature`. We could assume points are nodes, and have a binary `isRelation` to distinguish polygons coming from ways or relations, but that would exclude cases where we might would to generate a Point geometry for a polygon label, or generate a centerline Line for a polygon label.

[...]

PR #514, opened 2023-03-14
Expose setId in FeatureCollector [#377]

bdon:
To aid in setting OSM IDs in #377

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
@@ -201,7 +201,7 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
     private final Geometry geom;
     private final Map<String, Object> attrs = new TreeMap<>();
     private final GeometryType geometryType;
-    private long sourceId;
+    private long id;
 
     private int sortKey = 0;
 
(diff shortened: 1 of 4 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
```diff
@@ -240,6 +240,11 @@ public long getSourceId() {
       return sourceId;
     }
 
+    public Feature setSourceId(long sourceId) {
+      this.sourceId = sourceId;
+      return this;
+    }
+
```
msbarry, line 247: let's change this to `setId`

bdon: should we deprecate `getSourceId` and add `getId` ?

msbarry: We _could_ but I don't see any references to `getSourceId` in the code, so probably safe to remove.

bdon:
ptal

planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
bdon: where you usually interact with `FeatureCollector` you also have access to the `sourceFeature`, right? so might be redundant to have a `getSourceId` that delegate to the `FeatureCollector`'s `source` when you can call `source.id()`

PR #520, opened 2023-03-16
Expose pmtiles writer

msbarry:
Add support for using new pmtiles writer functionality by changing `--output` argument to point at a file ending in `.pmtiles`.

Major changes:

- Standardize on using the `--output=path.mbtiles` or `--output=path.pmtiles` command-line argument to control where the output goes.  `--mbtiles=output.mbtiles` will also continue to work until you upgrade from deprecated `setOutput("mbtiles", path)` to`setOutput(path)`.
- Output can accept a URI instead of just a path string.
- Instead of global command-line arguments controlling mbtiles output, you can specify them as query parameters on the output string:
   - `--compact-db=false` becomes `--output=out.mbtiles?compact=false`
   - `--skip-mbtiles-index-creation` becomes `--output=out.mbtiles?no_index`
   - `--optimize-db` becomes `--output=out.mbtiles?vacuum_analyze`

Also cleaned up the code since there were quite a few places that assumed mbtiles output:

- Created common `TileArchiveMetadata` that describes metadata that pmtiles and mbtiles metadata can hold and pulled metadata interface up into `WriteableTileArchive` and `ReadableTileArchive` interfaces.
- Created `TileArchiveConfig` that describes the location, format, and options of a tile archive to write to/read from
- Created `TileArchives` utility with static factory methods for creating readable and writable archives


planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
```diff
@@ -303,8 +301,9 @@ private void tileWriter(Iterable<TileBatch> tileBatches) throws ExecutionExcepti
         TileEncodingResult encodedTile;
         while ((encodedTile = encodedTiles.poll()) != null) {
           TileCoord tileCoord = encodedTile.coord();
-          assert lastTile == null || lastTile.compareTo(tileCoord) < 0 : "Tiles out of order %s before %s"
-            .formatted(lastTile, tileCoord);
+          assert lastTile == null ||
+            order.encode(tileCoord) > order.encode(lastTile) : "Tiles out of order %s before %s"
```
bdon, line 305: Technically the PMTiles writer can take out-of-order writes and has some logic to set `clustered=false` and re-sort the entries. If we are enforcing this via assertion then we could take that logic out.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchives.java
```diff
@@ -0,0 +1,75 @@
+package com.onthegomap.planetiler.archive;
```
msbarry, line 1: Not sure if `TileArchives` utilities for working with tile archives makes the most sense, or collapse `TileArchives` and `TileArchiveMetadata` into one `record TileArchive` that also has static factory methods?

bdon:
LGTM! We should also update the README/other doc files about the new output format but that can go in a follow-up

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchives.java
msbarry: oops sorry I meant collapse `TileArchives` and `TileArchiveConfig`

PR #522, opened 2023-03-19
Add --keep-unzipped option to avoid unzipping sources each time

msbarry:


bdon:
Should we make a similar enhancement for Shapefile inputs?

msbarry:
> Should we make a similar enhancement for Shapefile inputs?

I don't think we need to since the shapefile reader can read out of a zip file without extracting the contents first.

PR #566, opened 2023-04-25
Fix invalid polygons before snapping to tile coordinates

### LINKED ISSUE #546: [BUG] clipping creates incorrectly filled polygon geometry
Discovered a variant of #496 while investigating why a Dutch island was mysteriously disappearing at z6, but rendered fine on other levels.

https://user-images.githubusercontent.com/188935/229980662-5956801a-a83c-42d9-a743-24365ec86fc6.mov

The video is of @wipfli's map (it was the first public Planetiler source I could think of 😄), but the profile uses the same OSM water polygon source as my own, and exhibits exactly the same issue. [public map link](https://wipfli.github.io/swiss-map/#map=5.88/52.58/5.4)

The relevant portion of my profile is simple. There's no polygon merging or any other post processing:

```kotlin
    features.polygon("ocean")
      .setBufferPixels(4.0)
      .setMinZoom(6)
```

[swiss-map's implementation](https://github.com/wipfli/swiss-map/blob/main/planetiler/src/main/java/com/onthegomap/planetiler/examples/SwissMap.java#L268) is similarly simple.

I haven't dug into what exactly is happening, but looking at a tile inspector view, we can see the island is filled in (twice apparently, given the shading)

image

Some quick changes I tried:

- Set `bufferPixels` to 0: no effect
- Set `bufferPixels` to 100: no effect
- Set `minPixelSize` to 0: no effect
- Set `pixelTolerance` to 0.5: **fixed**

[Here's the relevant WKT geometry](https://github.com/onthegomap/planetiler/files/11155150/problematic-ocean.wkt.zip).

msbarry:
Fix #546 by fixing broken geometries before calling `GeometryPrecisionReducer.reduce`


planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1710,6 +1711,55 @@ void testIssue509LenaDelta() throws Exception {
     ), counts);
   }
 
+  @Test
+  void testIssue546Terschelling() throws Exception {
+    Geometry geometry = new WKBReader()
+      .read(
+        new InputStreamInStream(Files.newInputStream(TestUtils.pathToResource("issue_546_terschelling.wkb"))));
+    geometry = GeoUtils.worldToLatLonCoords(geometry);
+    Files.writeString(Path.of("test.wkt"), new WKTWriter2().write(geometry));
```
erik, line 1720: Was it intentional to check in `test.wkt`?

msbarry: good catch! that was for testing... fixed

Issue #552, opened 2023-04-12
[FEATURE] overture maps download shortcut


msbarry:
The first overture maps release is out. We should add an `overture:latest` and `overture:<version>` shortcut to download it.

PR #599, opened 2023-06-06
Add support for Overture maps downloads.

erik:
Conveniently we can reuse the existing `AwsOsm` class here, just pointing to a different S3 bucket.

There's only one release at the moment, so it's unclear if we'll need more filtering in the future (to exclude pre-release versions, for example).

Implements #552.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/AwsOsm.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/AwsOsm.java
@@ -14,18 +14,25 @@ import java.util.List;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A utility to download {@code planet.osm.pbf} files from <a href="https://registry.opendata.aws/osm/">AWS Open Data
- * Registry</a>.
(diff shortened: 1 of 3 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/util/AwsOsm.java
```diff
@@ -34,14 +41,14 @@ public class AwsOsm {
    * @return the URL of a {@code .osm.pbf} file with name or snapshot ID matching {@code searchQuery}
    * @throws IllegalArgumentException if no matches, or more than one match is found.
    */
-  public static String getDownloadUrl(String searchQuery, PlanetilerConfig config) {
+  public String getDownloadUrl(String searchQuery, PlanetilerConfig config) {
     IndexXml index = getAndCacheIndex(config);
```
msbarry, line 45: let's clean up this sonar warning while we're updating this

msbarry:
Nice! Thanks for the fix. Looks like we lucked out that the osm-pds filename patterns are very similar to the overture ones, and sorting the names will give you the latest one but we could parameterize those in the future if we need to for another source.

PR #622, opened 2023-07-11
Improve docs for generating custom vector tiles with a schema yaml

zhibek:
I followed the documentation at `planetiler-custommap/README.md` to generate custom vector tiles. However, until I found the `generate-custom` command referenced in PR #160, the standard vector tiles were being generated.

Contributing this PR to avoid others missing this. Or if I've missed something and `generate-custom` command need not be explicitly supplied in some situations?

### DIFF
--- a/planetiler-custommap/README.md
+++ b/planetiler-custommap/README.md
@@ -5,9 +5,9 @@ file as the first argument:
 
 ```bash
 # from a java build
-java -jar planetiler.jar generate-custom schema.yml
+java -jar planetiler.jar generate-custom --schema=schema.yml
 # or with docker (put the schema in data/schema.yml to include in the attached volume)
-docker run -v "$(pwd)/data":/data ghcr.io/onthegomap/planetiler:latest generate-custom /data/schema.yml
+docker run -v "$(pwd)/data":/data ghcr.io/onthegomap/planetiler:latest generate-custom --schema=/data/schema.yml
 ```
 
 Schema files are in [YAML 1.2](https://yaml.org) format and support [anchors and aliases](#anchors-and-aliases) for
--- a/planetiler-custommap/README.md
+++ b/planetiler-custommap/README.md
@@ -5,7 +5,7 @@ file as the first argument:
 
 ```bash
 # from a java build
-java -jar planetiler.jar schema.yml
+java -jar planetiler.jar generate-custom schema.yml
 # or with docker (put the schema in data/schema.yml to include in the attached volume)
 docker run -v "$(pwd)/data":/data ghcr.io/onthegomap/planetiler:latest generate-custom /data/schema.yml
 ```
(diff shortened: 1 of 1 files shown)


msbarry:
Thanks for the change! I'm a little confused though because I thought that this line in `Main.java` would have it treat those commands as a custommap yaml config run:

https://github.com/onthegomap/planetiler/blob/db6d1755076afef571f7227d707069babae81cb0/planetiler-dist/src/main/java/com/onthegomap/planetiler/Main.java#L66-L71

Can you share exactly what command you were running?

zhibek:
@msbarry I'm sorry for the confusion. When my first attempt at using the docker command in `planetiler-custommap/README.md` failed I tried different command options. I shifted to a named schema parameter (`--schema=<SCHEMA>`) rather than just `<SCHEMA>`. I didn't think to revert back to just `<SCHEMA>` as a direct argument, but confirming it does work for me now that I've tried. So this PR is redundant.

To explain what tripped me up...

Using this command results in the default schema being used (the schema argument is ignored as far as I can tell) and a 62mb file output:
```
docker run -v "$(pwd)":/data ghcr.io/onthegomap/planetiler:latest --area=greater-london --schema=/data/samples/power.yml
```

While using this command (including `generate-custom`) results in the schema argument being respected and a 140kb file output:
```
docker run -v "$(pwd)":/data ghcr.io/onthegomap/planetiler:latest generate-custom --area=greater-london --schema=/data/samples/power.yml
```

msbarry:
OK, got it - as someone who just went through figuring out how to use this from the docs do you think including the `generate-custom` subcommand makes more sense? I was originally thinking that syntax would be like treating the yaml file name as the subcommand, but sometimes shortcuts like that can make things less clear 🤷‍♂️

zhibek:
I appreciate I'm a sample-size of one(!), but I think including the `generate-custom` sub-command explicitly would be clearer.

Looking at the code you referenced in `planetiler/Main.java` I see how the logic matches the yaml file in the argument and switches the task to `generate-custom` based on this. But as someone trying to get this work by experimenting with different arguments this did confuse things.

Equally, possibly I missed something, but it was difficult to retrieve a list of sub-commands or find which parameters could be used with each sub-command. I did find a list parameters for the main command using `docker run -v "$(pwd)":/data ghcr.io/onthegomap/planetiler:latest --help`, but this doesn't include `--schema` (used by `generate-custom`). I also found a list of sub-commands with `docker run -v "$(pwd)":/data ghcr.io/onthegomap/planetiler:latest help` (given `help` isn't a valid sub-command). But trying to run `docker run -v "$(pwd)":/data ghcr.io/onthegomap/planetiler:latest generate-custom --help` returns an error stating the `schema` parameter is missing, but doesn't state what other parameters could be used.

FYI Now I've got this working, I'm really impressed with the functionality and performance. And there is good documentation and examples - it's just some entrypoints to jump in that weren't clear.

msbarry:
Got it, that makes sense to me.  Can you add `--schema=/data/schema.yml` to those commands?

zhibek:
> Got it, that makes sense to me. Can you add `--schema=/data/schema.yml` to those commands?

Sure - done.

PR #627, opened 2023-07-16
Fix regex escaping in polygon file parsing

mikaelhg:
Escape the character once for the regex, and a second time for Java.

Without the fix, polygon files delimited by tab characters fail to parse.


mikaelhg:
An example of such a file:   
https://raw.githubusercontent.com/osmandapp/OsmAnd-misc/master/osm-planet/polygons/europe/north-europe/finland/uusimaa.poly

msbarry:
Thanks! Good catch.

Issue #478, opened 2023-02-06
Improve runtime for small areas


msbarry:
Planetiler takes ~30 seconds to run even for the smallest areas (like andorra from geofabrik). Let's see if there is any way to improve that.  Here's a summary of runtime over andorra:

```
0:00:33 INF -   overall          33s cpu:1m12s gc:3s avg:2.2
0:00:33 INF -   lake_centerlines 3s cpu:12s gc:1s avg:4.4
0:00:33 INF -     read     1x(35% 0.9s done:2s)
0:00:33 INF -     process  9x(1% 0s wait:1s done:2s)
0:00:33 INF -     write    1x(0% 0s wait:1s done:2s)
0:00:33 INF -   water_polygons   12s cpu:17s avg:1.4
0:00:33 INF -     read     1x(94% 12s)
0:00:33 INF -     process  9x(0% 0s wait:12s)
0:00:33 INF -     write    1x(0% 0s wait:12s)
0:00:33 INF -   natural_earth    11s cpu:14s avg:1.3
0:00:33 INF -     read     1x(66% 7s done:4s)
0:00:33 INF -     process  9x(2% 0.2s wait:7s done:4s)
0:00:33 INF -     write    1x(0% 0s wait:8s done:4s)
0:00:33 INF -   osm_pass1        0.4s cpu:2s avg:3.7
0:00:33 INF -   osm_pass2        1s cpu:5s avg:5
0:00:33 INF -     read     1x(0% 0s)
0:00:33 INF -     process  9x(31% 0.3s)
0:00:33 INF -     write    1x(3% 0s wait:1s)
0:00:33 INF -   boundaries       0s cpu:0.1s avg:2.9
0:00:33 INF -   sort             0.1s cpu:0.7s avg:7.2
0:00:33 INF -   archive          0.5s cpu:3s avg:5.6
```

msbarry:
The biggest issues are natural earth and water polygons since planetiler has to deserialize every feature for the whole planet.  One idea would be to switch natural earth to read the geopackage source, and use the built-in spatial index to limit what we read to only what's inside the bounding box.  I'm not sure if we could do something similar with water polygons though?

cc/ @bdon

bdon:
[...]

PR #635, opened 2023-07-27
Use rtree spatial index from PlanetilerConfig bounds by default for Geopackages

bdon:
This is a first attempt at addressing #478 .

The existing issue is that all Shapefile or Geopackage features are read by every run, even if the `--bounds` argument is passed to limit the tile output or the bounds are determined implicitly based on the `osm.pbf`. To reduce the runtime we can:

* Use a spatial index on the dataset
  *  for Shapefile, we need `.fbn`, `.fbx` and/or `qix` sidecars depending on what software created the Shapefile. Maybe GeoTools can read all of these, but I'm not familiar with its internals. 
  * for Geopackage we can read from one of the `rtree_` indexes (A vanilla SQLite feature, no Spatialite required)
* Only read from a passed list of tables, for example instead of iterating every Natural Earth table we pass only `ne_50m_admin_0_boundary_lines_land` as tables we care about. IMO, this doesn't reduce the runtime by much relative to spatial indexing, so it may be an unworthy optimization. 

My proposal here is to not change any APIs, but change the default behavior of Geopackage to use the determined `PlanetilerConfig` bounds and apply the spatial index. Remaining todos:

- [x] determine if `rtree_` indexes will always exist on gpkg or are optional
- [x] determine the correct behavior with GPKGs in different projections. Bounds in PlanetilerConfig are always LatLong, so if the GPKG data is LatLong, Web Mercator or another conformal projection we can use the index. Otherwise we must ignore the index. 
- [ ] Ultimately we need antimeridian handling. We can accomplish this by detecting a crossing and splitting it into two bboxes, but support for this would need to be pushed up to PlanetilerConfig itself. 

Another relevant format would be FlatGeobuf, which has an optional spatial index as part of the single file, but I don't see key advantages of FGB over GPKG in most Planetiler use cases where data lives on disk.


planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -138,7 +149,20 @@ public void readFeatures(Consumer<SimpleFeature> next) throws Exception {
       MathTransform transform = (coordinateTransform != null) ? coordinateTransform :
         CRS.findMathTransform(CRS.decode("EPSG:" + srsId), latLonCRS);
 
-      for (var feature : features.queryForAll()) {
+      FeatureIndexManager indexer = new FeatureIndexManager(geoPackage,
+        features);
+      indexer.setIndexLocation(FeatureIndexType.RTREE);
```
erik, line 154: It's been a while, but if I remember correctly, this will actually build the index if it doesn't already exist in the DB before running the query.

Looking at the PR where I initially added this, should be possible to check for the presence of the index before trying to use it: https://github.com/onthegomap/planetiler/pull/413#discussion_r1051006021

bdon:
If I use the 4326 land/water polygons this drops the protomaps basemap build from ~30 seconds to 10 seconds for Monaco. Once we can port Natural Earth to use GPKG it should be even faster.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
bdon: I limited using the index to when the source is in EPSG:4326.

bdon:
@msbarry @erik take a look when you can

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -138,7 +148,21 @@ public void readFeatures(Consumer<SimpleFeature> next) throws Exception {
       MathTransform transform = (coordinateTransform != null) ? coordinateTransform :
         CRS.findMathTransform(CRS.decode("EPSG:" + srsId), latLonCRS);
 
-      for (var feature : features.queryForAll()) {
+      FeatureIndexManager indexer = new FeatureIndexManager(geoPackage,
+        features);
+
+      Iterable<FeatureRow> results;
+
+      if (this.bounds != null && indexer.isIndexed() && srsId == 4326) {
```
msbarry, line 156: Another option is to transform the bounds from 4326 to whatever CRS the geopackage source uses. See ShapefileReader as an example

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
```diff
@@ -39,9 +44,14 @@ public class GeoPackageReader extends SimpleReader<SimpleFeature> {
   private final GeoPackage geoPackage;
   private final MathTransform coordinateTransform;
 
-  GeoPackageReader(String sourceProjection, String sourceName, Path input, Path tmpDir, boolean keepUnzipped) {
+  private final Envelope bounds;
+
+  GeoPackageReader(String sourceProjection, String sourceName, Path input, Path tmpDir, boolean keepUnzipped,
+    Envelope bounds) {
```
msbarry, line 50: For consistency with ShapefileReader should this take `Bounds` class?

planetiler-core/src/test/java/com/onthegomap/planetiler/reader/GeoPackageReaderTest.java
```diff
@@ -66,13 +68,44 @@ void testReadGeoPackage(boolean keepUnzipped) throws IOException {
...
+        String id = "path=" + path + " proj=" + proj + " iter=" + iter;
+        assertEquals(86, reader.getFeatureCount(), id);
+        List<Geometry> points = new ArrayList<>();
+        WorkerPipeline.start("test", Stats.inMemory())
+          .fromGenerator("geopackage", reader::readFeatures, 1)
+          .addBuffer("reader_queue", 100, 1)
+          .sinkToConsumer("counter", 1, elem -> {
+            points.add(elem.latLonGeometry());
+          }).await();
+        assertEquals(expectedCount, points.size(), id);
```
msbarry, line 97: Do we need WorkerPipeline here? I think we could potentially simplify this to just:

```java
reader.readFeatures(feature -> {
  points.add(elem.latLonGeometry());
});
```

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
bdon: I wasn't sure if that handled non-conformal translations correctly, but if so we can use that. How were you able to use the Shapefile filter without `.sbx` sidecar files?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
bdon: Changed to `Bounds`

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
msbarry: Geotools `ReferencedEnvelope#transform` drops a bunch of points along the edge of the bounding box in the old coordinate system, then computes the bounding box of those points in the new coordinate system. Also if it throws a TransformException then it just skips the filter.

I'm not 100% sure how the geotools filter predicate works internally. My guess/hope would be that it uses the spatial index if available, but then falls back to just applying it against each shape if not. I added it based off the description here: https://docs.geotools.org/latest/userguide/library/main/filter.html

msbarry:
Thanks! This looks good to me.

Issue #536, opened 2023-03-27
[FEATURE] New output archive type for individual files


msbarry:
Now that we have the ability to switch between output formats easily with string URIs from #520, we should add a new output type that just writes vector tiles directly to disk.

See https://github.com/onthegomap/planetiler/discussions/533#discussioncomment-5433488

PR #639, opened 2023-07-30
add support for simple file output streams: CSV, JSON, protobuf

bbilger:
The primary use case for writing tile data to files using simple formats, is to provide a simple plugin mechanism. By using named pipes data can be transformed and processed directly - e.g. to upload to s3 or to store in a DB, without any intermediate file (mbtiles/pmtiles).

This - depending on the use-case - also allows concurrent writing to separate files. I left a few questions in `TileArchiveWriter` about what needs to be done to handle that there.

#### JSON
```json
{"type":"initialization","metadata":{"name":"0", ...}}
{"type":"tile","x":0,"y":0,"z":0,"encodedData":"..."}
{"type":"tile","x":1,"y":2,"z":3,"encodedData":"..."}
{"type":"finish","metadata":{"name":"1", ...}}
```
entries are separated by newline (can be changed via config), and the output of non-tile entries can be suppressed

#### CSV
```csv
0,0,0,encodedDataBase64
1,2,3,encodedDataBase64
```
the value separator can be changed, and no meta data is output in that format

#### protobuf
protobuf-message are length-delimited

in Java the messages can be read from the file/pipe as follows:

```java
// note: do not use nio (Files.newInputStream) for pipes
try (var in = new FileInputStream(...)) {
  StreamArchiveProto.Entry entry;
  while ((entry = StreamArchiveProto.Entry.parseDelimitedFrom(in)) != null) {
    ...
  }
}
```

#### s3

A very simple example to push tiles directly to s3 could look likes this:

```bash
mkfifo /tmp/data/output.csv
./s3-upload < /tmp/data/output.csv
java -Xmx1g -jar planetiler.jar --download --area=monaco --append --output=/tmp/data/output.csv
```
with `s3-upload` being something like this:
```bash
#! /bin/bash
while IFS="," read -r x y z encoded
do
  echo "pushing tile z=$z x=$x y=$y"
  echo $encoded | base64 -d | aws s3 cp - s3://BUCKET/map/$z/$x/$y.pbf --content-type=application/x-protobuf --content-encoding=gzip
  # echo $encoded | base64 -d | gzip -d | aws s3 cp - s3://BUCKET/map/$z/$x/$y.pbf --content-type=application/x-protobuf
done < "${1:-/dev/stdin}"
```

Parallel:
```bash
mkfifo /tmp/data/output.csv
mkfifo /tmp/data/output.csv1
./s3-upload < /tmp/data/output.csv
./s3-upload < /tmp/data/output.csv1
java -Xmx1g -jar planetiler.jar --download --area=monaco --append --output=/tmp/data/output.csv --tile_write_threads=2
```

##### MySQL

```
mkfifo /tmp/data/output.csv
java -Xmx1g -jar planetiler.jar --download --area=monaco --append --output=/tmp/data/output.csv

mysql>...create tile(s) table
mysql> LOAD DATA INFILE '/tmp/data/output.csv'
  -> INTO TABLE tiles
  -> FIELDS TERMINATED BY ','
  -> LINES TERMINATED BY '\n'
  -> (tile_column, tile_row, zoom_level, @var1)
  -> SET tile_data = FROM_BASE64(@var1);
```

##### PostgreSQL

```
mkfifo /tmp/data/output_raw.csv
mkfifo /tmp/data/output_transformed.csv
# prefix hex-data with '\x' for the postgres import
cat /tmp/data/output_raw.csv | sed -r 's/^([0-9]+,)([0-9]+,)([0-9]+,)(.*)$/\1\2\3\\x\4/' > /tmp/data/output_transformed.csv

java -Xmx1g -jar planetiler.jar --download --area=monaco --append --output=/tmp/data/output_raw.csv --csv_binary_encoding=hex

...create tile(s) table

postgres=# \copy tiles(tile_column, tile_row, zoom_level, tile_data) from /tmp/data/output_transformed.csv DELIMITER ',' CSV;
```

This is related to:
- https://github.com/onthegomap/planetiler/discussions/261#discussioncomment-2928291
- https://github.com/onthegomap/planetiler/discussions/533
- https://github.com/onthegomap/planetiler/issues/536


planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableJsonStreamArchive.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  private static class JsonTileWriter implements TileWriter {
+
+    private final Writer writer;
+    private final SequenceWriter jsonWriter;
+    private final String rootValueSeparator;
+
+    JsonTileWriter(OutputStream outputStream, String rootValueSeparator) {
+      this.writer = new BufferedWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8.newEncoder()));
```
msbarry, line 114: I'm wondering would it be more efficient here to just make a `BufferedOutputStream(outputStrem)` and have jackson `writeValues` go to that? It seems like the curent one does an extra round trip from bytes to utf8 string and back.

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableProtoStreamArchive.java
```diff
@@ -0,0 +1,184 @@
...
+  // for quick testing
+  // note: do not use nio (Files.newInputStream) for pipes
+  public static void main(String[] args) throws IOException {
+    try (var in = new FileInputStream(args[0])) {
+      StreamArchiveProto.Entry entry;
+      while ((entry = StreamArchiveProto.Entry.parseDelimitedFrom(in)) != null) {
+        System.out.println(entry);
+      }
+    }
+  }
```
msbarry, line 183: Remove this testing code?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -652,8 +654,37 @@ public void run() throws Exception {
...
+      throw new IllegalArgumentException("require tile_write_threads >= 1");
+    }
+    if (config.tileWriteThreads() > 1) {
+      if (!output.format().supportsConcurrentWrites()) {
+        throw new IllegalArgumentException(output.format() + " doesn't support concurrent writes");
+      }
+      IntStream.range(1, config.tileWriteThreads())
+        .mapToObj(index -> StreamArchiveUtils.constructIndexedPath(output.getLocalPath(), index))
+        .forEach(p -> {
+          if (overwrite || config.force()) {
```
msbarry, line 679: also probably want to add a check here for if we're not appending

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableProtoStreamArchive.java
```diff
@@ -0,0 +1,184 @@
...
+import java.nio.file.Path;
+import java.util.function.Consumer;
+import org.locationtech.jts.geom.CoordinateXY;
+import org.locationtech.jts.geom.Envelope;
+
+/**
+ * Writes protobuf-serialized tile data as well as meta data into file(s). The messages are of type
+ * {@link StreamArchiveProto.Entry} and are length-delimited.
+ */
+public final class WriteableProtoStreamArchive extends WritableStreamArchive {
```
msbarry, line 22: The proto output stream is neat, I'm just wondering how/when you were thinking we'd make use of it?

msbarry (commented):
Very cool! I think that we'd still eventually want to add native support for more popular output formats (s3/file/postgres) as that will most likely end up having the best performance,  but I do like having the ability to support any kind of output format using this technique!

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/ReadableTileArchive.java
```diff
@@ -45,5 +45,9 @@ default byte[] getTile(TileCoord coord) {
    */
   TileArchiveMetadata metadata();
 
+  default boolean supportsMetadata() {
+    return true;
+  }
```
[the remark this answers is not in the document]
bbilger: going forward I would say that it's a property of the actual archive instance
json can be written w/o; theoretically an mbtiles (or any DB) could be generated with just tiles and without any meta data

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveConfig.java
```diff
@@ -166,18 +166,33 @@ public Arguments applyFallbacks(Arguments arguments) {
   }
 
   public enum Format {
-    MBTILES("mbtiles"),
-    PMTILES("pmtiles");
+    MBTILES("mbtiles", false, false),
```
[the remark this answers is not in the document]
bbilger: mbtiles **could** support append - in the sense of overwrite probably - , as well but that would require some additional work: 
- check if tables exist before creating them
- upsert rather then insert when writing tile data

not sure if it's worth the effort but it's definitely doable

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableJsonStreamArchive.java
bbilger: yes I agree, this makes total sense

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableProtoStreamArchive.java
bbilger: I tend to leave such code for testing if it's simple and handy, but okay, I'll remove

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WritableStreamArchive.java
```diff
@@ -0,0 +1,119 @@
...
+    final int tileWriterIndex = tileWriterCounter.getAndIncrement();
+    if (tileWriterIndex == 0) {
+      return newTileWriter(getPrimaryOutputStream());
+    } else {
+      return newTileWriter(outputStreamFactory.newOutputStream(tileWriterIndex));
+    }
+
+  }
+
+  @Pure
```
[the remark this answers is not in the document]
bbilger: it's coming with guava
```
[INFO] |  +- com.google.guava:guava:jar:32.1.1-jre:compile
[INFO] |  |  +- com.google.guava:failureaccess:jar:1.0.1:compile
[INFO] |  |  +- com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava:compile
[INFO] |  |  +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |  |  +- org.checkerframework:checker-qual:jar:3.33.0:compile
[INFO] |  |  +- com.google.errorprone:error_prone_annotations:jar:2.18.0:compile
[INFO] |  |  \- com.google.j2objc:j2objc-annotations:jar:2.8:compile
```
and https://checkerframework.org/api/org/checkerframework/dataflow/qual/Pure.html

I was initially using the method from the constructor... but since I changed that, I'll remove the annotation now

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableProtoStreamArchive.java
bbilger: Out of all formats the proto one should be the fastest to read/write given it's binary. If protobuf code generation is setup in a project, then it should also be the simplest to process given the model+parsing is generated then.
It should also be the least error-prone to use; parsing JSON even with jackson can be tricky if not done correctly: fail on unknown (i.e. new) properties, fail on unknown (i.e. new) enum values, ....

So for writing anything more serious/stables (let's say some pipeline that updates tiles in postgres - batched - with the tile ID or quadkey as primary key), I would probably choose the proto one over JSON.

I understand if you think that's too niche but given protobuf codegeneration is already setup, I think it's rather simple to add support for it. But other than it's faster+easier (depends case by case), most cases can be realized with JSON as well.

- CSV vs proto/JSON - probably boils down to whether meta data is needed
- proto vs CSV/JSON - probably boils down to binary vs text and "simplicity"/"speed" of parsing

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
```diff
@@ -137,15 +138,16 @@ public static void writeOutput(FeatureGroup features, WriteableTileArchive outpu
           writerEnqueuer.accept(batch); // also send immediately to writer
         });
         writerQueue.close();
+        // TODO can this restriction be lifted when tileWriteThreads>1 ?
```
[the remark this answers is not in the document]
bbilger: to start with: leaving the changes here aside not all "formats" need tiles to be emitted in order. An S3 upload for example does not need/benefit from any order.

currently Planetiler behaves like this (tileWriteThreads=1, and tiles being emitted in order)
```
 ____________        ___________________________________________________
| Planetiler | ---> () ... 10, 09, ,08, 07, 06, 05, 04, 03, 02, 01, 00 () --> output.x
 ‾‾‾‾‾‾‾‾‾‾‾‾        ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```

with the changes here, tileWriteThreads>1 and tiles being emitted in order, it behaves like this
```
                     ______________________
               +--> () ... 09, 08, 01, 00 () --> output.xyz
               |     ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 ____________  |     ______________________ 
| Planetiler | +--> ()     ... 10, 05, 04 () --> output.xyz1
 ‾‾‾‾‾‾‾‾‾‾‾‾  |     ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
               |     ______________________
               +--> () ... 07, 06, 03, 02 () --> output.xyz2
                     ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```
so when the output is spread over multiple files, tiles are still ordered within a file - but spread. So I don't think there's any benefit of enforcing the order when tileWriteThreads>1 (and also for tileWriteThreads=1 there are cases where the order is not needed).

Now if it's worth to NOT emit tiles in order, is something I don't know. For now it's probably simplest/best to just keep emitting in order. So I would at least keep it out of this PR, and if there's a strong need then the option to not emit tiles in order again could be re-added.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
```diff
@@ -251,7 +253,10 @@ private void tileEncoder(Iterable<TileBatch> prev, Consumer<TileBatch> next) thr
             bytes = null;
           } else {
             encoded = en.encode();
-            bytes = gzip(encoded);
+            bytes = switch (config.tileCompression()) {
+              case GZIP -> gzip(encoded);
+              case NONE -> encoded;
+            };
```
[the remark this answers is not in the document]
bbilger: https://github.com/onthegomap/planetiler/pull/639/commits/4e051631fef27f294c92373a59f8f27303147624
added tileCompression to TileArchiveMetadata, and handle it accordingly for pmtiles, as well as mbtiles
note mbtiles spec is mentioning it: https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#future-directions

let me know if you had something different in mind, and/or if you want to handle tileCompression as string in TileArchiveMetadata rather as an enum

planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java
```diff
@@ -929,6 +930,16 @@ public TileArchiveMetadata get() {
       String[] center = map.containsKey(TileArchiveMetadata.CENTER_KEY) ?
         map.remove(TileArchiveMetadata.CENTER_KEY).split(",") : null;
       var metadataJson = MetadataJson.fromJson(map.remove("json"));
+
```
msbarry, line 933: I since the spec requires gzip compression, I wonder if we should log a warning in the set method if tile compression != gzip?

planetiler-core/src/test/java/com/onthegomap/planetiler/stream/StreamArchiveUtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+class StreamArchiveUtilsTest {
+
+  @ParameterizedTest
+  @ArgumentsSource(GetEscpacedStringTestArgsProvider.class)
```
msbarry, line 21: Minor: Could we use `@CsvSource` for this?

planetiler-core/src/test/java/com/onthegomap/planetiler/stream/WriteableJsonStreamArchiveTest.java
```diff
@@ -0,0 +1,225 @@
...
+          ImmutableMap.of("1", LayerStats.FieldType.BOOLEAN, "2", LayerStats.FieldType.NUMBER, "3",
+            LayerStats.FieldType.STRING),
+          Optional.of("description"), OptionalInt.of(1), OptionalInt.of(2)),
+        new LayerStats.VectorLayer("vl1",
+          Map.of(),
+          Optional.empty(), OptionalInt.empty(), OptionalInt.empty())
+      ),
+      ImmutableMap.of("a", "b", "c", "d"),
+      TileCompression.GZIP);
+  private static final String maxMetadataOut = """
```
msbarry, line 44: Can we always expect the metadata to be serialized in this exact order? I usually do something like parse both to an object and ensure the objects are equal just in case key ordering changes in the future.

msbarry (commented):
This is looking good! A few mostly minor comments.

planetiler-core/src/test/java/com/onthegomap/planetiler/stream/WriteableJsonStreamArchiveTest.java
msbarry: ok got it, thanks! Good to know about ImmutableMap preserving order.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchives.java
msbarry, line 1: While running this locally, I'm realizing there are some format "aliases" that would be nice if we handled... For example `output.tsv` would be equivalent to `--output='output.tsv?format=csv' --csv-column-separator='\t'` and `output.pbf` would be equivalent to the `output.proto`.

Would it be possible to have those mappings? I think `TileArchiveConfig#getFormat` could map additional aliases specified in `TileArchiveConfig.Format` enum, but I'm not exactly sure the best way to change the default value for the arguments?

msbarry (commented):
Looks good! I ran this locally and noticed 2 minor things, after that looks good to merge.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchives.java
bbilger: good idea.
I, however, think the cleanest is to add separate formats for them. Otherwise things get a bit strange, for example for format specific parameters 
```
--output='output.tsv' --csv-column-separator='\t'
# vs
--output='output.tsv' --tsv-column-separator='\t'
```
the latter looks better to me
it also simplifies the handling of format specifics

so I implemented it accordingly

wdyt?

bbilger:
I think I addressed everything now
Let me know if I missed something or you want me to change/adjust anything else.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchives.java
msbarry: That looks good to me. Thanks!!

msbarry:
Looks great! Thanks for adding this.

PR #642, opened 2023-08-03
remove the old TileEncodingResult

bbilger:
...only to be merged once planetiler-openmaptile has been updated, so the failing build is expected until then

-> https://github.com/openmaptiles/planetiler-openmaptiles/pull/93

just some code cleanup
let me know if missed any reason for this still being used


msbarry:
Thank you!!

PR #656, opened 2023-09-06
Tile stats

### LINKED ISSUE #391: [FEATURE] Expose more detailed tile size statistics
When developing a new tile set, there isn't a great way to dig into why a particular tile is too large. We get a warning that a specific tile is over the configured max size, but to investigate why we need to switch to another tool that can inspect the tile and print out details about the geometry size, which layers are included, etc.

I'd like to be able to see a breakdown of sizes (basic summary stats) by layer and zoom level at the end of a run. This could be printed to stdout (maybe formatted like [st]?), written to a separate file as JSON, or included in the MBTiles `metadata` table.

The [MBTiles spec] references another mapbox project called [geostats], which looks like a good starting point, but seems more interested in counts of features and peeking at their attribute values than on-disk size.

**Additional context**
Briefly discussed [in Slack](https://osmus.slack.com/archives/C031V9E9RMG/p1669137766947899)

[st]: https://github.com/nferraz/st
[MBTiles spec]: https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#tilestats
[geostats]: https://github.com/mapbox/mapbox-geostats#output-the-stats

msbarry:
Expose more detailed tile size statistics from planetiler:

- Run planetiler with `--output-layerstats` to output an extra `<output file>.layerstats.tsv.gz` file (~5% of the original archive size) with a row per layer per tile that can be analyzed using duckdb (see `layerstats/README.md`)
- Run `java -jar planetiler.jar stats --input=<pmtiles or mbtiles file> --output=layerstats.tsv.gz` on an existing archive to compute stats for it.
- By default planetiler now emits top 10 largest tiles, max layer size by zoom, and weighted average tile size at the end of every run.
- For either command, add `--tile-weights=weights.tsv.gz` to point planetiler to a file with z, x, y, loads columns to customize the weights used for weighted-average tile sizes, or use `--download-osm-tile-weights` to download pre-computed top-1m tiles from opemnstreetmap.org traffic.
- Generate a custom top-tiles file from openstreetmap.org traffic by running `java -jar planetiler.jar top-osm-tiles --days=<# days to fetch> --top=<# tiles to include> --output=weights.tsv.gz`

Fixes #391


msbarry:
Sounds good to me. Any thoughts on the formats?  Most recently I've been generated a single tsv.gz with columns:

- z
- x
- y
- hilbert
- tile_bytes
- gzipped_tile_bytes
- deduped_tile_id
- layer
- features
- layer_bytes
- layer_attr_bytes
- layer_attr_values
- min_lon (these duplicate z,x,y data but make it easier to generate debug URLs to look at a tile, or filter within a bounding box)
- max_lon
- min_lat
- max_lat

You can either split them out into 2 tables (z,x,y->id, and id,layer->stats) or duplicate all of the z,x,y tile IDs into one denormalized table.  Split makes it easy to query tile-level stats but duckdb struggles with joining to use tile coordinates with layer-level stats (too big to add an index), and denormalized makes it easier to work with layer-level stats, but duckdb struggles grouping by tile ID to get tile level stats.

msbarry:
And in terms of outputting a high-level summary, the most useful ones have been:

- a single "weighted average tile size" (raw and gzipped)  number based on weighting tile sizes by top n most-visited osm tiles
- the total archive size
- a table of max layer size by layer and zoom
- a list of the top N biggest tiles, along with which layers cause them to be big and z/lat/lon to generate a link to visually inspect them (there's probably overlap between biggest tiles and tiles with biggest layers - might want to merge the lists or something)

bdon:
> min_lon (these duplicate z,x,y data but make it easier to generate debug URLs to look at a tile, or filter within a bounding box)

Can we solve this via a DuckDB function definition instead of repeating information in the stats?

msbarry:
> > min_lon (these duplicate z,x,y data but make it easier to generate debug URLs to look at a tile, or filter within a bounding box)
> 
> Can we solve this via a DuckDB function definition instead of repeating information in the stats?

Yeah, it looks like that could work - would just need to put it in documentation somewhere for people to copy/paste:

```sql
D create or replace macro lat(z,y) as round((180/pi())*atan(0.5*(exp(pi()-2*pi()*(y/(2**z)))-exp(-(pi()-2*pi()*(y/(2**z)))))), 5);
D create or replace macro lon(z,x) as (180/pi())*atan(0.5*(exp(pi()-2*pi()*(y/(2**z)))-exp(-(pi()-2*pi()*(y/(2**z))))));
D create or replace macro url(z,x,y) as concat('https://onthegomap.github.io/planetiler-demo/#', z+0.5, '/', lat(z,y+0.5), '/', lon(z,x+0.5));
D select z,x,y,layer,round(layer_bytes/1000) as kbs, url(z,x,y) from stats order by layer_bytes desc limit 10;
```
returns:
```
┌───────┬───────┬───────┬─────────────┬────────┬───────────────────────────────────────────────────────────────────────┐
│   z   │   x   │   y   │    layer    │  kbs   │                             url(z, x, y)                              │
│ int64 │ int64 │ int64 │   varchar   │ double │                                varchar                                │
├───────┼───────┼───────┼─────────────┼────────┼───────────────────────────────────────────────────────────────────────┤
│    14 │ 13722 │  7013 │ housenumber │ 2412.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/25.05575/121.51978 │
│    14 │ 13723 │  7014 │ housenumber │ 1850.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/25.03584/121.54175 │
│    14 │ 13723 │  7013 │ housenumber │ 1827.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/25.05575/121.54175 │
│    14 │  6435 │  8361 │ building    │ 1761.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/-3.72175/-38.59497 │
│    14 │ 13724 │  7014 │ housenumber │ 1626.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/25.03584/121.56372 │
│    14 │ 13667 │  7134 │ housenumber │ 1597.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/22.62415/120.31128 │
│    14 │  6435 │  8364 │ building    │ 1550.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/-3.78752/-38.59497 │
│    14 │  6435 │  8363 │ building    │ 1510.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/-3.7656/-38.59497  │
│    14 │ 13683 │  7058 │ housenumber │ 1472.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/24.15678/120.66284 │
│    14 │ 13724 │  7013 │ housenumber │ 1460.0 │ https://onthegomap.github.io/planetiler-demo/#14.5/25.05575/121.56372 │
├───────┴───────┴───────┴─────────────┴────────┴───────────────────────────────────────────────────────────────────────┤
│ 10 rows                                                                                                    6 columns │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

There's a benefit to having it persisted in parquet file to allow efficient bbox filtering, but if you're going to go through duckdb to create a parquet file you could just add it there.

msbarry:
@bdon what do you think about the different format options?

> - tile-level (z,x,y->tile size)
> - deduped-tile-level (z,x,y->id, id->tile sizes)
> - layer-level (z,x,y,layer->layer sizes)
> - or layer/attribute level (z,x,y,layer,attr->size that attribute contributes to layer)

bdon:
* I don't think we care about deduplicated data, since deduped data is uninteresting for stats 99.9% of the time
* layer/attribute level is the most interesting, and the most useful for situations like we're doing in protomaps/basemaps where many types of things are combined into the `landuse`, `pois` and `roads` layer

so #4 would be great (though unclear to me what the TSV looks like),  then #3, #1, #2

msbarry:
OK that makes sense. Would you omit "deduped_tile_id" column as well? Or "hilbert"?

I was thinking 4 would be something like: `z, x, y, layer, attribute/geom, size` so you could see the size contributed by geometry vs. individual attributes. It would be expensive to compute and large, but might make sense if you're looking at a small number of tiles or low-zoom area. But it sound like maybe you're looking for total size of features that have certain attributes set?  

Unfortunately it's hard to get tile-level stats from 3 since duckdb runs out of memory on the huge "group by" unless you limit to a small set of tiles first.

One other route to go might be to have a row per tile, and included nested data in it... for example if you import a file:

```json
{"id":1,"layers":[{"id":"layer1","size":100}, {"id":"layer2","size":200}]}
{"id":2,"layers":[{"id":"layer2","size":300}, {"id":"layer3","size":400}]}
```

Then you can do:

```
D select * from stats;
┌───────┬────────────────────────────────────────────────────────────┐
│  id   │                           layers                           │
│ int64 │             struct(id varchar, size bigint)[]              │
├───────┼────────────────────────────────────────────────────────────┤
│     1 │ [{'id': layer1, 'size': 100}, {'id': layer2, 'size': 200}] │
│     2 │ [{'id': layer2, 'size': 300}, {'id': layer3, 'size': 400}] │
└───────┴────────────────────────────────────────────────────────────┘
select id, unnest(layers).id, unnest(layers).size from stats;
┌───────┬─────────────────────┬───────────────────────┐
│  id   │ (unnest(layers)).id │ (unnest(layers)).size │
│ int64 │       varchar       │         int64         │
├───────┼─────────────────────┼───────────────────────┤
│     1 │ layer1              │                   100 │
│     1 │ layer2              │                   200 │
│     2 │ layer2              │                   300 │
│     2 │ layer3              │                   400 │
└───────┴─────────────────────┴───────────────────────┘
```

I'm not sure how performant querying that would be though, might be worth a test.

msbarry:
I tried outputting newline-delimited json that looks like:

```json
{"z":0,"x":0,"y":0,"hilbert":0,"total_bytes":142374,"gzipped_bytes":82957,"layers":[{"name":"water","features":2,"total_bytes":8435,"attr_bytes":18,"attr_values":2},{"name":"landcover","features":11,"total_bytes":1578,"attr_bytes":27,"attr_values":2},{"name":"place","features":42,"total_bytes":115614,"attr_bytes":82249,"attr_values":4363},{"name":"water_name","features":6,"total_bytes":7042,"attr_bytes":5541,"attr_values":238},{"name":"boundary","features":32,"total_bytes":9689,"attr_bytes":807,"attr_values":37}]}
{"z":1,"x":0,"y":0,"hilbert":1,"total_bytes":103276,"gzipped_bytes":64171,"layers":[{"name":"water","features":2,"total_bytes":4139,"attr_bytes":18,"attr_values":2},{"name":"landcover","features":3,"total_bytes":468,"attr_bytes":27,"attr_values":2},{"name":"place","features":29,"total_bytes":68351,"attr_bytes":47302,"attr_values":2493},{"name":"water_name","features":7,"total_bytes":7136,"attr_bytes":5562,"attr_values":245},{"name":"boundary","features":13,"total_bytes":23165,"attr_bytes":277,"attr_values":14}]}
{"z":1,"x":0,"y":1,"hilbert":2,"total_bytes":38635,"gzipped_bytes":24721,"layers":[{"name":"water","features":2,"total_bytes":1340,"attr_bytes":18,"attr_values":2},{"name":"landcover","features":8,"total_bytes":982,"attr_bytes":27,"attr_values":2},{"name":"place","features":10,"total_bytes":25423,"attr_bytes":17164,"attr_values":915},{"name":"water_name","features":3,"total_bytes":3102,"attr_bytes":2431,"attr_values":97},{"name":"boundary","features":5,"total_bytes":7772,"attr_bytes":37,"attr_values":5}]}
{"z":1,"x":1,"y":1,"hilbert":3,"total_bytes":91365,"gzipped_bytes":52859,"layers":[{"name":"water","features":2,"total_bytes":2206,"attr_bytes":18,"attr_values":2},{"name":"landcover","features":1,"total_bytes":521,"attr_bytes":27,"attr_values":2},{"name":"place","features":23,"total_bytes":66179,"attr_bytes":48121,"attr_values":2174},{"name":"water_name","features":4,"total_bytes":6010,"attr_bytes":4716,"attr_values":196},{"name":"boundary","features":6,"total_bytes":16432,"attr_bytes":66,"attr_values":6}]}
{"z":1,"x":1,"y":0,"hilbert":4,"total_bytes":275506,"gzipped_bytes":166904,"layers":[{"name":"water","features":2,"total_bytes":3647,"attr_bytes":18,"attr_values":2},{"name":"place","features":76,"total_bytes":191026,"attr_bytes":129687,"attr_values":7421},{"name":"water_name","features":5,"total_bytes":5869,"attr_bytes":4640,"attr_values":193},{"name":"boundary","features":55,"total_bytes":74950,"attr_bytes":1212,"attr_values":55}]}
```

The raw json.gz file was 3.3GB, duckdb file was 1.8GB and exported parquet file was 1.4GB.  You can query tile-level stats instantly:

```sql
select z,x,y,gzipped_bytes from 'nested_stats.parquet' order by gzipped_bytes desc limit 10;
```

But a layer-level query takes 40 seconds:

```sql
select z,x,y,unnest(layers).name,unnest(layers).total_bytes from stats order by unnest(layers).total_bytes desc limit 10;
```

vs. this query which is instant with option 3:

```sql
select z,x,y,layer,layer_bytes from flat order by layer_bytes desc limit 10;
```

But if you know you are going to be doing layer-level  analysis you could create an unnested table once (~60 seconds), then use it for instant layer-level queries afterwards:

```sql
create table unnested as (select z,x,y,unnest(layers) layer from stats); -- ~60 seconds
select z,x,y,layer.name,layer.total_bytes from unnested order by layer.total_bytes desc limit 10; -- instant
```

bdon:
@msbarry I would omit both of those, and simply not include any tile that occurs more than once in the statistics (anything I'm missing here?)

My PoC is here: https://github.com/protomaps/go-pmtiles/pull/75

for now it just outputs a `.tsv.gz` with z,x,y,compressed_length contents

next steps:

* should we decide on a naming convention for these files? if I run `pmtiles stats foo.pmtiles` it will create the sidecar `foo.pmtiles.stats.tsv.gz`
* How do you determine the bytes by each layer, by looking at the protobuf message length? I'm using https://github.com/paulmach/orb which unmarshals MVT straight to GeoJSON-like Simple Features, so I may need to use the raw protobuf-generated functions
* would it make sense for the duckdb utility functions and rough spec to live in the planetiler repo at a dir like `/stats`? if so I can make go-pmtiles just follow that.

msbarry:
> * should we decide on a naming convention for these files? if I run `pmtiles stats foo.pmtiles` it will create the sidecar `foo.pmtiles.stats.tsv.gz`

That sounds good to me, I could foresee wanting different flavors of stats so maybe `foo.pmtiles.tilestats.tsv.gz` or `foo.pmtiles.layerstats.tsv.gz` ?

> * How do you determine the bytes by each layer, by looking at the protobuf message length? I'm using https://github.com/paulmach/orb which unmarshals MVT straight to GeoJSON-like Simple Features, so I may need to use the raw protobuf-generated functions

Yeah you need to parse raw protobuf.  I'm pretty sure the tile is just a series of concatenated layers, and the java library at least lets you get the serialized byte length of a parsed proto struct.  You could probably also decompress with a very simple schema to only separate the layers (it just passes through all the unrecognized fields)

```proto
message Tile {
  message Layer {
    required string name = 1;
  }
  repeated Layer layers = 3;
}
```

> * would it make sense for the duckdb utility functions and rough spec to live in the planetiler repo at a dir like `/stats`? if so I can make go-pmtiles just follow that.

Yeah, anywhere common would work but would definitely be useful to have some common queries listed out (especially the z/x/y -> lat/lon macro)

bdon:
* I realized the term "tilestats" is taken by https://github.com/mapbox/mapbox-geostats and well used by default output from tippecanoe so we should go with "layerstats"
* I prefer we have one omnibus stats file instead of multiple flavors, as long is it's fast in DuckDB
* maybe we could have `.layerstats` imply `.tsv.gz`? I don't think the `.gz` bit is useful if we're not expanding this in Finder (it would be really big) - so just `planet.pmtiles.layerstats`
* If gzipping is the bottleneck maybe we should use zstd, lzw or lzma? prefer something that's in the go stdlib (lzw, gzip) but open to others

msbarry:
> * I realized the term "tilestats" is taken by https://github.com/mapbox/mapbox-geostats and well used by default output from tippecanoe so we should go with "layerstats"

👍

> * I prefer we have one omnibus stats file instead of multiple flavors, as long is it's fast in DuckDB

Sounds good, after more testing I think layer to tile grouping should be reasonably fast enough. Just thinking if we went down to attribute stats that would be a lot bigger?

> * maybe we could have `.layerstats` imply `.tsv.gz`? I don't think the `.gz` bit is useful if we're not expanding this in Finder (it would be really big) - so just `planet.pmtiles.layerstats`

The benefit of naming `.tsv.gz` is that it tells duckdb how to parse it without any explicit instructions (`select * from 'planet.pmtiles.layerstats.tsv.gz'`)

> 
> * If gzipping is the bottleneck maybe we should use zstd, lzw or lzma? prefer something that's in the go stdlib (lzw, gzip) but open to others

Gzipping takes ~3m for the planet, maybe less with a smaller format so probably not a big enough deal to worry about...

msbarry:
For the actual stats format I think we definitely need:

- z
- x
- y
- gzipped_tile_bytes
- layer
- layer_bytes

then maybe:

- hilbert ID
- layer_attr_bytes
- features
- tile_bytes (can be derived from `sum(layer_bytes)`)
- layer_attr_values
- deduped_tile_id (or tile hash, something to indicate deduplication)

then I think we can skip the `{min,max}_{lon,lat}` values

Are there any "maybes" you think would also be useful to include @bdon ?

bdon:
* If we have `deduped_tile_id` it should be an opaque type not not necessarily a hash (I will use offset uint64)

bdon:
more notes while implementing:

* instead of `gzipped_tile_bytes` why not `archived_bytes`,  that makes it clear it is the size of the tile at-rest in the archive, not the result of gzipping the raw tile data which might be different depending on gzip params and implementation
* should we sort the layers alphabetically? it would be ideal if the output of planetiler and go-pmtiles was identical

layerstats/README.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+The output is a gzipped tsv with a row per layer on each tile and the following columns:
+
+|       column        |                                                                   description                                                                   |
+|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
+| z                   | tile zoom                                                                                                                                       |
+| x                   | tile x                                                                                                                                          |
+| y                   | tile y                                                                                                                                          |
+| hilbert             | tile hilbert ID (defines [pmtiles](https://protomaps.com/docs/pmtiles) order)                                                                   |
```
bdon, line 24: should this be the first column, indicating that everything is sorted by this? (is that true for mbtiles?)

layerstats/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+To get a table of biggest layers by zoom:
+
+```sql
+pivot (
+  select z, layer, (max(layer_bytes)/1000)::int size from layerstats group by z, layer order by z asc
+) on z using sum(size);
+```
+
+|        layer        |  0  |  1  | 10  | 11  | 12  | 13  | 14  |  2  |  3  |  4  |  5  | 6  |  7  |  8  |  9  |
```
bdon, line 60: can we make this sorted numerically instead of lexicographically?

layerstats/README.md
```diff
@@ -0,0 +1,175 @@
...
+create table tilestats as select
+z, x, y,
+any_value(archived_tile_bytes) gzipped,
+sum(layer_bytes) raw
+from layerstats group by z, x, y;
+select * from tilestats order by gzipped desc limit 2;
+```
+
+NOTE: this group by uses a lot of memory so you need to be running in file-backed
+mode `duckdb analysis.duckdb` (not in-memory mode)
```
bdon, line 91: is there a one-liner to change the `.tsv.gz` into a `.duckdb`?

layerstats/README.md
msbarry: If you run this over an mbtiles file, it will be sorted by z/x/y. I was kind of leaning towards z/x/y first since those are the more human-readable tile coordinates?

bdon: Right, that makes sense then.

layerstats/README.md
msbarry:
```
PIVOT ... ON printf('%2d', z)
```

🙃

layerstats/README.md
msbarry: I think it would be

```
duckdb analysis.duckdb -cmd "CREATE TABLE layerstats AS SELECT * FROM 'output.pmtiles.layerstats.tsv.gz';"
```

to drop you into a shell after importing the file, of `-c "create...` to just create the file - given the shortcut's not much shorter than the individual steps I'm inclined to just leave them as separate steps for clarity.

msbarry:
Ran on the planet with these settings used by [openstreetmap-americana](https://github.com/ZeLonewolf/openstreetmap-americana)

https://github.com/ZeLonewolf/planetiler-scripts/blob/2c2c7e3a9f045b1e44e31226b1b2e5d6b4412510/local/render_once.sh#L30-L37

image

The output tsv.gz file was 2.8GB

PR #725, opened 2023-11-15
Add `setAttrWithMinSize` to feature API

msbarry:
Add a new `features.getMinZoomForPixelSize(zoom)` method to feature API that returns the minimum zoom level at which a feature is a certain size in pixels, and a convenience `feature.setAttrWithMinSize(key, value, minPixelSize)` method that uses it to set the min zoom based on feature size.

For more flexibility you can do math on the result of `features.getMinZoomForPixelSize(zoom)` or create a `ZoomFunction` that calculates `feature.getPixelSizeAtZoom(zoom)` and applies a custom threshold based on the zoom level.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
@@ -704,6 +704,27 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


wipfli:
Nice. Do you have an example use case?

msbarry:
This openmaptiles PR does something similar for the `brunnel` tag on road segments at lower zooms to  help avoid breaking them up when they get merged https://github.com/openmaptiles/planetiler-openmaptiles/pull/126 - I figure that's a common use case so makes sense to be a reusable utility in planetiler

PR #752, opened 2023-12-14
Compute layer attr stats from actual vector tile features

### LINKED ISSUE #749: Tilejson attributes missing attributes added or removed during post-processing
Tilejson attributes are currently computed from attributes during geometry pre-processing, which includes some attributes removed during post-processing and misses attributes added during post-processing. Should instead compute tilejson from the actual attributes present in the tiles written to the archive.

_Originally posted by @mdoakes42 in https://github.com/onthegomap/planetiler/discussions/746#discussioncomment-7841296_

msbarry:
Tilejson stats in the archive metadata are currently computed while processing source features, but this misses vector tile attributes added and removed during post-processing. This PR moves computing the attr stats to tile-write time to capture the actual attributes written on features to the output archive.

A side effect of this is that tile archive `initialize` method can no longer accept metadata since the metadata isn't fully-known until the tiles are done being written so any archive implementation needs write metadata from the `finish` method.

Fixes #749, see #746

### DIFF
--- a/planetiler-core/src/main/proto/stream_archive_proto.proto
+++ b/planetiler-core/src/main/proto/stream_archive_proto.proto
@@ -1,4 +1,3 @@
-
 syntax = "proto3";
 
 package com.onthegomap.planetiler.proto;
@@ -19,7 +18,6 @@ message TileEntry {
 }
 
 message InitializationEntry {
-  Metadata metadata = 1;
 }
 
 message FinishEntry {
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
@@ -81,7 +81,7 @@ public class VectorTile {
   private static final int EXTENT = 4096;
   private static final double SIZE = 256d;
   private final Map<String, Layer> layers = new LinkedHashMap<>();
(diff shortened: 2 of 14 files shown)


planetiler-core/src/main/proto/stream_archive_proto.proto
```diff
@@ -19,7 +18,6 @@ message TileEntry {
 }
 
 message InitializationEntry {
-  Metadata metadata = 1;
```
msbarry, line 22: @bbilger I need to remove metadata from the archive init call since it's not fully known until done writing all tiles. Any concerns about this?

msbarry:
Before:
image

After:
image

Also Poi and Park layers do not have `rank` field in metadata before change, but they do after the change

planetiler-core/src/main/proto/stream_archive_proto.proto
bbilger: no concerns :+1: 

just a note for the future: once 1.0 gets out, we might want to consider to ensure that changes in protobuf are backwards compatible - which in that case would mean to reserve the field number+name
but absolutely not needed now (pre 1.0), and I don't think anybody is using it, anyways

bdon:
Looks like a useful fix to me!

PR #761, opened 2023-12-20
Add support for "files"-archive

### LINKED ISSUE #536: [FEATURE] New output archive type for individual files

bbilger:
i.e. write individual pbf-files to disk in the format BASE/z/x/y.pbf

in order to use that format it must be passed as "--ouput=/path/to/tiles?format=files"

Fixes #536

note: it also supports concurrent writing - just have to pass --tile_write_threads=5


planetiler-core/src/main/java/com/onthegomap/planetiler/files/WriteableFilesArchive.java
```diff
@@ -0,0 +1,92 @@
...
+    // nothing to do here
+  }
+
+  private static void mkdirs(Path p) {
+    try {
+      Files.createDirectories(p);
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
```
msbarry, line 57: You could use `FileUtils.createDirectory` as well, which does the same catch/throw.

planetiler-core/src/test/java/com/onthegomap/planetiler/archive/TileArchiveConfigTest.java
```diff
@@ -33,4 +42,51 @@ void testPmtiles() {
...
+        case FILES -> assertEquals(p, config.getLocalPath());
+        default -> assertEquals(config.getLocalPath().getParent().resolve(Paths.get("output1")), p);
+      }
+    }
+  }
+
+  @Test
+  void testExistsForFilesArchive(@TempDir Path tempDir) throws IOException {
+    final Path out = tempDir.resolve("outdir");
+    final var config = TileArchiveConfig.from(out + "?format=files");
```
msbarry, line 76: Currently when `format=` is missing `TileArchiveConfig` defaults to mbtiles. I wonder if when the output path has no extension or ends with a slash if it should default to `files` as a small ease of use improvement?

planetiler-core/src/main/java/com/onthegomap/planetiler/files/WriteableFilesArchive.java
```diff
@@ -0,0 +1,92 @@
...
+      final Path file = FilesArchiveUtils.absolutePathFromTileCoord(basePath, encodingResult.coord());
+      final Path folder = file.getParent();
+
+      // tiny optimization in order to avoid too many unnecessary "folder-exists-checks" (I/O)
+      if (!lastCheckedFolder.equals(folder) && !Files.exists(folder)) {
+        mkdirs(folder);
+      }
+      lastCheckedFolder = folder;
+      try {
+        Files.write(file, encodingResult.tileData());
```
msbarry, line 81: `encodingResult.tileData()` is gzipped here. We should probably make it configurable whether or not the output files should be gzipped (I think default to false?). If they aren't gzipped we can skip the gzip step while encoding, and if they are gzipped the output extension should probably be `.pbf.gz`

bbilger: The output is not necessarily gzipped. This is already configurable at a higher level: tile_compression. So I don't think anything needs to be handled here specifically at the archive level.

msbarry:
Regarding performance, this is definitely going to be slower but I wonder if there's any performance difference to `Files.write` from multiple OS threads vs. `Files.write` from virtual threads vs. something like `AsynchronousFileChannel` (or others?).

These could also be separate followup issues, but I noticed:
- when there are multiple writer threads with `--tile_write_threads=5` they each print `Finished z11 in 0.4s cpu:0.6s avg:1.7, now starting z12`, call `tileWriter.printStats()` and call `archive.finish()` - maybe there should be a different way to handle those? Or only do the logging from thread 0 or something?
- `FileUtils.size` looks like a bottleneck because the `ProgressLoggers` calls it while printing status during the archive write phase. Might make sense to keep a counter of bytes written to the output directory instead of having to scan it each time?

planetiler-core/src/main/java/com/onthegomap/planetiler/files/FilesArchiveUtils.java
```diff
@@ -0,0 +1,21 @@
...
+final class FilesArchiveUtils {
+
+  static final String PBF_FILE_ENDING = ".pbf";
+
+  private FilesArchiveUtils() {}
+
+  static Path relativePathFromTileCoord(TileCoord tc) {
+    return Paths.get(Integer.toString(tc.z()), Integer.toString(tc.x()),
+      tc.y() + PBF_FILE_ENDING);
+  }
```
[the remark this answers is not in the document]
bbilger: added support for custom "tile_scheme" supported templates are `{x}`, `{y}`, `{z}` as well as `{xs}` and `{ys}` (s=safe). `{xs}` and `{ys}` are split into 2 levels.
```
{z}/{x}/{y}.pbf     => 3/1/2.pbf
{x}/{y}/{z}.pbf     => 1/2/3.pbf
{x}-{y}-{z}.pbf     => 1-2-3.pbf
{x}/a/{y}/b{z}.pbf  => 1/a/2/b3.pbf
{z}/{x}/{y}.pbf.gz  => 3/1/2.pbf.gz
{z}/{xs}/{ys}.pbf   => 3/000/001/000/002.pbf
{z}/{x}/{ys}.pbf    => 3/1/000/002.pbf
{z}/{xs}/{y}.pbf    => 3/000/001/2.pbf
```

planetiler-core/src/main/java/com/onthegomap/planetiler/files/ReadableFilesArchive.java
```diff
@@ -0,0 +1,99 @@
...
+      return CloseableIterator.of(it);
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  public TileArchiveMetadata metadata() {
+    return null;
+  }
```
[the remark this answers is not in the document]
bbilger: done

did a bunch of refactorings to the (de-)serialization of TileArchiveMetadata

https://github.com/onthegomap/planetiler/pull/761/commits/d238a4f30ca63abc450a4c7361a11705a13f501c

planetiler-core/src/test/java/com/onthegomap/planetiler/archive/TileArchiveConfigTest.java
bbilger: done https://github.com/onthegomap/planetiler/pull/761/commits/1faf56611a11f9e92ec02f631262689c8a13e417

planetiler-core/src/main/java/com/onthegomap/planetiler/files/WriteableFilesArchive.java
bbilger: I personally prefer the tiles to be at a known location by default. With the introduction of the tile_scheme param users can adjust as needed.
https://github.com/onthegomap/planetiler/commit/2e4652ef39e501b472265852c261d7685cf179ee

bbilger:
> These could also be separate followup issues, but I noticed:
> 
> * when there are multiple writer threads with `--tile_write_threads=5` they each print `Finished z11 ... now starting z12`, call `tileWriter.printStats()` and call `archive.finish()` - maybe there should be a different way to handle those? Or only do the logging from thread 0 or something?
> * `FileUtils.size` looks like a bottleneck because the `ProgressLoggers` calls it while printing status during the archive write phase, and `Stats` calls it at the end to print a file size summary. Might make sense to either disable (have the method give up a certain number of files or for a certain amount of time) or keep a counter of bytes written to the output directory instead of having to scan it each time?

Addressed those points here: https://github.com/onthegomap/planetiler/pull/761/commits/5371520256376a3b61114d54b0e56be75e19bf5d

1. `archive.finish()` being called for each writer was definitely a bug - it's now called after ALL writers finished
2. `Finished z11 ... now starting z12`, call `tileWriter.printStats()` - logging for the first writer only - not 100% accurate but should be good enough for some "zoom-progress logging"
3. `tileWriter.printStats()` think that's ok that we print stats for each writer as long as they are writer specific. I introduced printStats on WriteableTilesArchive now for archive related logging.
4. `FileUtils.size` in `ProgressLoggers`  was definitely an issue. I introduced WriteableTilesArchive#bytesWritten in order to handle that correctly (was also an issue for the stream writers).

bbilger:
> Regarding performance, this is definitely going to be slower than a single-file archive but I wonder if there's any performance difference to `Files.write` from multiple OS threads vs. `Files.write` from virtual threads vs. something like `AsynchronousFileChannel` (or others?).

Good question/point. Need to investigate a bit on this.

bbilger:
regarding performance: I wrote some benchmark (BenchmarkTileFilesWrites) that loads a mbtiles file into memory and then writes the tiles to files in different modes.

See the benchmark results below from my machine for US (0<=z<=13). If I interpret things correctly, then there's no real benefit for virtual threads. At the same time async seems to be good compromise between absolute and CPU time.

So I impletemented:
1. async file writing for the files archive with the option to write sync.
2. a way to use virtual threads for the tile writers (`--tile_write_threads=100 --tile_write_executor=virtual`) - in case you want to experiment with it

If you agree, then I'd remove the benchmark as well as the option to use virtual threads for writing, again. 

| mode | run | duration |
| --- | --- | --- |
| fixed1 | 0 | 4m34s cpu:4m25s gc:1s avg:1 |
| fixed1 | 1 | 4m20s cpu:4m21s gc:2s avg:1 |
| fixed2 | 0 | 3m12s cpu:5m56s avg:1.9 |
| fixed2 | 1 | 2m54s cpu:5m43s avg:2 |
| fixed4 | 0 | 2m55s cpu:11m2s avg:3.8 |
| fixed4 | 1 | 2m56s cpu:11m2s avg:3.8 |
| fixed10Virtual | 0 | 3m18s cpu:22m21s avg:6.8 |
| fixed10Virtual | 1 | 3m19s cpu:22m30s avg:6.8 |
| fixed100Virtual | 0 | 3m30s cpu:30m14s gc:1s avg:8.6 |
| fixed100Virtual | 1 | 3m33s cpu:30m32s avg:8.6 |
| fixed1000Virtual | 0 | 3m42s cpu:33m47s gc:1s avg:9.1 |
| fixed1000Virtual | 1 | 3m46s cpu:33m56s avg:9 |
| async | 0 | 3m53s cpu:6m56s avg:1.8 |
| async | 1 | 3m54s cpu:7m13s avg:1.8 |
| unboundVirtual | 0 | 4m29s cpu:19m42s gc:3s avg:4.4 |
| unboundVirtual | 1 | 4m37s cpu:20m34s gc:3s avg:4.5 |

changes are here: https://github.com/onthegomap/planetiler/pull/761/commits/b8cfa56977d98520aa8b62252c3a2726d440afe0

msbarry:
> regarding performance: I wrote some benchmark (BenchmarkTileFilesWrites) that loads a mbtiles file into memory and then writes the tiles to files in different modes.
> 
> See the benchmark results below from my machine for US (0<=z<=13). If I interpret things correctly, then there's no real benefit for virtual threads. At the same time async seems to be good compromise between absolute and CPU time.
> 
> So I impletemented:
> 
> 1. async file writing for the files archive with the option to write sync.
> 2. a way to use virtual threads (bound only) for the tile writers (`--tile_write_threads=100 --tile_write_executor=virtual`) - in case you want to experiment with it
> 
> If you agree, then I'd remove the benchmark as well as the option to use virtual threads for writing, again.
> 
> mode	run	duration
> fixed1	0	4m34s cpu:4m25s gc:1s avg:1
> fixed1	1	4m20s cpu:4m21s gc:2s avg:1
> fixed2	0	3m12s cpu:5m56s avg:1.9
> fixed2	1	2m54s cpu:5m43s avg:2
> fixed4	0	2m55s cpu:11m2s avg:3.8
> fixed4	1	2m56s cpu:11m2s avg:3.8
> fixed10Virtual	0	3m18s cpu:22m21s avg:6.8
> fixed10Virtual	1	3m19s cpu:22m30s avg:6.8
> fixed100Virtual	0	3m30s cpu:30m14s gc:1s avg:8.6
> fixed100Virtual	1	3m33s cpu:30m32s avg:8.6
> fixed1000Virtual	0	3m42s cpu:33m47s gc:1s avg:9.1
> fixed1000Virtual	1	3m46s cpu:33m56s avg:9
> async	0	3m53s cpu:6m56s avg:1.8
> async	1	3m54s cpu:7m13s avg:1.8
> unboundVirtual	0	4m29s cpu:19m42s gc:3s avg:4.4
> unboundVirtual	1	4m37s cpu:20m34s gc:3s avg:4.5
> changes are here: [b8cfa56](https://github.com/onthegomap/planetiler/commit/b8cfa56977d98520aa8b62252c3a2726d440afe0)

Thanks for doing those tests! I saw similar results running this on a linux box with fast SSD - fastest seemed to be with 4 dedicated OS threads.

Given these results, can we just remove the extra code paths and keep the configurable number of fixed threads you had initially since it's the fastest and simplest?

planetiler-core/src/main/java/com/onthegomap/planetiler/files/WriteableFilesArchive.java
```diff
@@ -0,0 +1,250 @@
...
+      }
+    }
+
+    @Override
+    public void close() {
+      // nothing to do here
+    }
+  }
+
+  private static class AsyncTileFilesWriter extends BaseTileFilesWriter {
```
msbarry, line 175: Let's remove the async tile writer and collapse into a single sync writer class here for simplicity

planetiler-core/src/main/java/com/onthegomap/planetiler/stats/ProcessInfo.java
```diff
@@ -234,6 +234,7 @@ public static Map<Long, ThreadState> getThreadStats() {
   public static ThreadState getCurrentThreadState() {
     ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
     ThreadInfo thread = threadMXBean.getThreadInfo(Thread.currentThread().getId());
-    return new ThreadState(threadMXBean, thread);
+    // null for virtual - tmp workaround
+    return thread == null ? null : new ThreadState(threadMXBean, thread);
```
msbarry, line 238: We can remove this too.

planetiler-core/src/main/java/com/onthegomap/planetiler/stats/Timers.java
```diff
@@ -115,7 +115,11 @@ public Finishable startTimer(String name, boolean logStart) {
   public void finishedWorker(String prefix, Duration elapsed) {
     Stage stage = currentStage.get();
     if (stage != null) {
-      stage.threadStats.add(new ThreadInfo(ProcessInfo.getCurrentThreadState(), prefix, elapsed));
+      final ProcessInfo.ThreadState threadState = ProcessInfo.getCurrentThreadState();
+      // null for virtual - tmp workaround
+      if (threadState != null) {
+        stage.threadStats.add(new ThreadInfo(threadState, prefix, elapsed));
+      }
```
msbarry, line 122: We can remove this

planetiler-core/src/main/java/com/onthegomap/planetiler/stream/WriteableStreamArchive.java
```diff
@@ -34,6 +36,8 @@
  */
 abstract class WriteableStreamArchive implements WriteableTileArchive {
 
+  private final LongAdder bytesWritten = new LongAdder();
```
msbarry, line 39: Can we use `Counter.newMultiThreadCounter` here and get a new `counterForThread` from it in the factory?

planetiler-core/src/main/java/com/onthegomap/planetiler/worker/Worker.java
```diff
@@ -41,6 +41,20 @@ public Worker(String prefix, Stats stats, int threads, RunnableThatThrows task)
...
+   * @param prefix       string ID to add to logs and stats
+   * @param stats        stats collector for this thread pool
+   * @param threads      number of parallel threads to run {@code task} in
+   * @param executorType the type of executor to use
+   * @param task         the work to do in each thread
+   */
+  @SuppressWarnings("java:S1181")
+  public Worker(String prefix, Stats stats, int threads, ExecutorType executorType, RunnableThatThrows task) {
+    this(prefix, stats, threads, executorType, i -> task.run());
+  }
```
msbarry, line 56: Let's remove this as well since we're just doing the fixed threads. I think things using virtual threads going forward will probably use a slightly different paradigm (new virtual thread per task, with a semaphore to limit concurrency) so probably don't need to be supported in the worker framework.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/WriteableTileArchive.java
```diff
@@ -44,6 +44,10 @@ default void initialize() {}
    */
   default void finish(TileArchiveMetadata tileArchiveMetadata) {}
 
+  default void printStats() {}
+
+  long bytesWritten();
```
msbarry, line 49: Thanks for adding this, I could do this in a separate PR as well but there's one remaining place where it gets the size of the archive which could take a very long time if you just wrote tiles for the planet: `Stats.printSummary` - ideally it either uses the same output from this method, or we add a failsafe to `FileUtils.size` that gives up after a certain number of files/time since that output duplicates what got printed earlier we won't miss much.

msbarry (commented):
Thanks for making these changes! I'm trying to test it with the custom schemes but getting some errors.

My first attempt was to use `--output=tiles/{z}/{x}-{y}.pbf.gz` - it would be cool if that worked but looks like it's not currently handled.

Then I tried `--output=tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz` but I get `Illegal character in query at index 31: tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz`.

To make it work I had to do `--output=tiles&format=files&tile_scheme=%7Bz%7D%2F%7Bx%7D%2F%7By%7D.pbf.gz` which isn't super easy to use.

Is there another way that I'm missing?

bbilger:
> Thanks for making these changes! I'm trying to test it with the custom schemes but getting some errors:
> 
> My first attempt was to use `--output=tiles/{z}/{x}-{y}.pbf.gz` - it would be cool if that worked but looks like it's not currently handled.
> 
> Then I tried `--output=tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz` but I get `Illegal character in query at index 31: tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz`.
> 
> To make it work I had to do `--output=tiles&format=files&tile_scheme=%7Bz%7D%2F%7Bx%7D%2F%7By%7D.pbf.gz` which isn't super user-friendly.
> 
> Is there another way that I'm missing?

The other option - which I always use - is to use format pre-fixed args => `--output=tiles/ --files_tile_scheme={x}/{y}/{z}.pbf`

I'll check if we can support your other approaches.

msbarry:
> > Thanks for making these changes! I'm trying to test it with the custom schemes but getting some errors:
> > My first attempt was to use `--output=tiles/{z}/{x}-{y}.pbf.gz` - it would be cool if that worked but looks like it's not currently handled.
> > Then I tried `--output=tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz` but I get `Illegal character in query at index 31: tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz`.
> > To make it work I had to do `--output=tiles&format=files&tile_scheme=%7Bz%7D%2F%7Bx%7D%2F%7By%7D.pbf.gz` which isn't super user-friendly.
> > Is there another way that I'm missing?
> 
> The other option - which I always use - is to use format pre-fixed args => `--output=tiles/ --files_tile_scheme={x}/{y}/{z}.pbf`
> 
> I'll check if we can support your other approaches.

Ah got it, that makes sense.  Could we add a short javadoc to the major new classes and maybe include that cli arg snippet in the javadoc for readable/writeable files archive classes?

The shorthand one would be a nice to have, but might be hard to disambiguate between that and writing to a file archive on disk?

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveMetadataDeSer.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.Map;
+import java.util.Optional;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.CoordinateXY;
+import org.locationtech.jts.geom.Envelope;
+
+/**
+ * Container for everything related to (de-)serialization of {@link TileArchiveMetadata}
+ */
+public final class TileArchiveMetadataDeSer {
```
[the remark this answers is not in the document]
bbilger: np - I think overall the (de-) serialization is now more centralized and less duplicate code

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
```diff
@@ -371,8 +387,6 @@ private void tileWriter(Iterable<TileBatch> tileBatches) throws ExecutionExcepti
     if (time != null) {
```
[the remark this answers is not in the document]
bbilger: I think there's still a point in having printStats for a writer. And there's one usage https://github.com/onthegomap/planetiler/blob/cbb092a1f76925e0642cc4c8a954783f59c9f3aa/planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java#L804
Moving that to the archive's printStats feels strange. But if we for sure could if you want, or log from #close in that specific case.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/WriteableTileArchive.java
```diff
@@ -44,6 +44,10 @@ default void initialize() {}
    */
   default void finish(TileArchiveMetadata tileArchiveMetadata) {}
 
+  default void printStats() {}
```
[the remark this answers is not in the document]
bbilger: I think there's still a point in having printStats for a writer. And there's one usage https://github.com/onthegomap/planetiler/blob/cbb092a1f76925e0642cc4c8a954783f59c9f3aa/planetiler-core/src/main/java/com/onthegomap/planetiler/mbtiles/Mbtiles.java#L804
Moving that to the archive's printStats feels strange. But if we for sure could if you want, or log from #close in that specific case.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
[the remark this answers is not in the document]
bbilger: I am not quite sure what you mean here, sorry

planetiler-core/src/main/java/com/onthegomap/planetiler/files/FilesArchiveUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+  static WriteableFilesArchive.WriteMode writeMode(Arguments options) {
+    final String mode = options.getString(
+      OPTION_WRITE_MODE,
+      "the write mode - either sync or async",
+      WriteableFilesArchive.WriteMode.ASYNC.toString().toLowerCase()
+    );
+    return WriteableFilesArchive.WriteMode.valueOf(mode.toUpperCase());
+  }
+}
```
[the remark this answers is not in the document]
bbilger: there's not a lot code for the async mode, and it might be beneficial to have it but okay, I removed it again

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java
```diff
@@ -330,11 +334,18 @@ private void tileEncoderSink(Iterable<TileBatch> prev) throws IOException {
...
   }
 
+  private final AtomicBoolean firstTileWriterTracker = new AtomicBoolean(true);
+
   private void tileWriter(Iterable<TileBatch> tileBatches) throws ExecutionException, InterruptedException {
+
+    final boolean firstTileWriter = firstTileWriterTracker.compareAndExchange(true, false);
+    if (firstTileWriter) {
+      archive.initialize();
+    }
```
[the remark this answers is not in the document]
bbilger: Good point, was thinking about it as well but thought not very likely.
Didn't find any good spot to move it to. I moved it here now: https://github.com/onthegomap/planetiler/blob/e39551859bffc08d6b2f769f7cdf47bb20ccd614/planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveWriter.java#L154
Let me know if you have a better idea.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/WriteableTileArchive.java
bbilger: done: https://github.com/onthegomap/planetiler/pull/761/commits/c41b95354f224a112cba09e380ccac9de8aa235b

bbilger:
> > > Thanks for making these changes! I'm trying to test it with the custom schemes but getting some errors:
> > > My first attempt was to use `--output=tiles/{z}/{x}-{y}.pbf.gz` - it would be cool if that worked but looks like it's not currently handled.
> > > Then I tried `--output=tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz` but I get `Illegal character in query at index 31: tiles?format=files&tile_scheme={z}/{x}/{y}.pbf.gz`.
> > > To make it work I had to do `--output=tiles&format=files&tile_scheme=%7Bz%7D%2F%7Bx%7D%2F%7By%7D.pbf.gz` which isn't super user-friendly.
> > > Is there another way that I'm missing?
> > 
> > 
> > The other option - which I always use - is to use format pre-fixed args => `--output=tiles/ --files_tile_scheme={x}/{y}/{z}.pbf`
> > I'll check if we can support your other approaches.
> 
> Ah got it, that makes sense. Could we add a short javadoc to the major new classes (don't worry about all of them, the helper classes should be self-explanatory without any javadoc) and maybe include that cli arg snippet in the javadoc for readable/writeable files archive classes?
> 
> The shorthand one would be a nice to have, but might be hard to disambiguate between that and writing to a file archive on disk?

I added support for those shortcuts here: https://github.com/onthegomap/planetiler/pull/761/commits/7d0bc26b72d1ee7b95c85d8a5196cbbecf7040d4
Detecting it is rather simple by the templates ({) - which otherwise should not really be used, and even if so users can pass the format-query param to enforce some format.

I also added javadoc where I thought it makes sense but let me know if you want more.

msbarry (commented):
Thanks for making those changes, it's looking good, a couple of minor comments...

planetiler-core/src/main/java/com/onthegomap/planetiler/files/FilesArchiveUtils.java
```diff
@@ -0,0 +1,73 @@
...
+        " - instead of {x}/{y} {xs}/{ys} can be used which splits the x/y into 2 directories each" +
+        " which ensures <1000 files per directory",
+      defaultTileScheme
+    );
+    return new TileSchemeEncoding(tileScheme, basePath);
+  }
+
+  static BasePathWithTileSchemeEncoding basePathWithTileSchemeEncoding(Arguments options, Path basePath) {
+    final String basePathStr = basePath.toString();
+    final int curlyIndex = basePathStr.indexOf('{');
```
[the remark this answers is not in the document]
bbilger: good catch! adjusted the implementation accordingly

bbilger:
Should be all set now. Let me know if there's anything else.

msbarry (approved):
Looks great, thank you!!

PR #769, opened 2023-12-28
fix center coordinates of pmtiles writer

bdon:
Simple mistake not caught by tests. Tests adjusted to be more realistic

Fixes https://github.com/protomaps/basemaps/issues/193


msbarry:
Good catch! Thanks for fixing.

PR #772, opened 2024-01-06
add tile-copy utility

bbilger:
to copy tiles from one archive into another e.g. mbtiles to files, ...


msbarry:
Awesome, thank you! This has been on my todo list for months.

wipfli:
Can you share an example?

bbilger:
@wipfli sure you just need to pass `--input=/path/to/archive.ext` and `--output=/path/to/output/archive.ext`
e.g. `--input=/path/to/archive.mbtiles --output=/path/to/output/archive.pmtiles`

You can either run TileCopy from your IDE if you have things setup, or if not.
1. Checkout the branch
2. Build `./mvnw clean package`
3. Run `java -jar planetiler-dist/target/planetiler-dist-0.7-SNAPSHOT-with-deps.jar tile-copy --input=/path/to/input/monco.mbtiles --output=/path/to/output/monco.mbtiles`

All archives (mbtiles, pmtiles, csv, json, proto) are supported as input as well as output.

More examples here: https://github.com/onthegomap/planetiler/pull/772/files#diff-54f69fe27546dacbdffb609ef73b1b8044fa9b06f96018fc4980d59bdb1983d5R30

If you need to pass any input/output archive specific params then either use query params, or use args prefixed with `in_` or rather `out_`

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileCopy.java
```diff
@@ -0,0 +1,236 @@
...
+      var writer = TileArchives.newWriter(config.outArchive(), config.outArguments())
+    ) {
+
+      final TileArchiveMetadata inMetadata = getInMetadata(reader);
+      final TileArchiveMetadata outMetadata = getOutMetadata(inMetadata);
+
+      writer.initialize();
+      try (
+        var rawTiles = reader.getAllTiles();
+        var it = config.skipEmpty() ? rawTiles.filter(t -> t.bytes() != null && t.bytes().length > 0) : rawTiles
```
msbarry, line 74: What if the "native order" of the archive is different between the archives? If they are the same I think the approach of iterating and pushing them through output makes sense, but if they are different we can choose the source or target tile order.  So far, all planetiler written in order to the target archive.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileCopy.java
```diff
@@ -0,0 +1,236 @@
...
+      ) {
+
+        final var tileConverter = tileConverter(inMetadata.tileCompression(), outMetadata.tileCompression(), writer);
```
msbarry, line 77: The tile converter seems like it could get expensive if compressions are different. Given we are only using a read and write thread so far, might make sense to add a worker per core when the converter has something to do?

I could also see that converter doing the work to deserialize tiles and update tile stats eventually, or filter-out layers, apply some geometry transform, merge multiple input archives, etc. which would make sense in a worker-per-core model.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileCopy.java
msbarry: Also, there's a lot of things we could add to this, we don't need to add all of them in this PR but one interesting one might be setting a bounding box/zoom range. In that case we'd likely want a different enumeration mechanism that lists all the tiles in the order expected by the dest archive, and queries the source archive for it.  Here's an initial version I was playing with a few months ago that iterates in dest archive order, does work in parallel and joins the ordered tiles at the end: 

https://github.com/onthegomap/planetiler/blob/38ca9b4edbdc9bf194ae028233e873ad2e122dbf/planetiler-core/src/main/java/com/onthegomap/planetiler/util/TileArchiveCopier.java#L122

I guess though we'll never be able to query tiles from a json/protobuf/csv input?

bbilger: > What if the "native order" of the archive is different between the archives? ...
Okay, I see what I can do here.

> I guess though we'll never be able to query tiles from a json/protobuf/csv input?
Well yes, it's implemented here as well. Loading all tiles (coord+data) in order is simple and fast - just read the file from top to bottom. Reading tile data by a coord is also implemented but highly inefficient - so that is not super useful but it's implemented for debugging and simple/small tile sets.

> ...enumeration...
I see. That code is helpful. Think I'll tackle it a bit different - since for json/protobuf/csv loading all corrds and then load data by coord is not efficient. Thinking more in the direction of an iterator - also in order to avoid loading all tileCoords/-IDs into memory. For "in-order"-iteration we can - at least for json/protobuf/csv - load the coord and the tile data eagerly. For anything else we would load the tile data lazy.

msbarry: Got it, seems like we might want to have 2 iteration modes:

- if input/output preferred modes are different AND input format supports random access (and user opts-in?) then enumerate the tiles in preferred output order and query input archive for them
- otherwise use this simple iterator approach

In any case, seems like it might be a good idea to log a warning when the output order is not going to be the preferred order of the output format just so the user is aware. Some formats care more than others, for example pmtiles sets a "clustered=true" header field when output tiles are ordered that enables some client-side query optimizations.

I also think it's fine to start with the simple iterator approach then add the random-access-output-order approach in a later PR as an optimization.

msbarry: Or another option besides random access queries would be to use `ExternalMergeSort` class to write tiles to temporary storage blocks, sort them in parallel, then iterate through them in order to write to the output format. That would work for any input format.

Another complication will be an HTTP tile archive that only supports random access but not iteration (s3 would be similar but does support iteration), but we could cross that bridge when we get there.

I think this just builds the case for doing the output-order change in a separate PR since there are a few options to eva

msbarry: Good timing! @bdon just did a blog post on the benefits of clustered pmtiles archives too: https://protomaps.com/blog/pmtiles-cluster/

nyurik:
Note that you can copy `.mbtiles` to `.mbtiles` using Martin's mbtiles utility - including filtering (zoom and multiple bbox-es)

PR #774, opened 2024-01-07
initial s3 support (draft)

bbilger:
early draft => needs tests and some other adjustments but wanted to share early in case you have any concerns

### DIFF
--- a/planetiler-core/pom.xml
+++ b/planetiler-core/pom.xml
@@ -154,6 +154,11 @@
       <artifactId>geopackage</artifactId>
       <version>${geopackage.version}</version>
     </dependency>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
+      <artifactId>s3</artifactId>
+      <version>2.22.12</version>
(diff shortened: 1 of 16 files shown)


msbarry:
Nice! I wonder if S3 should be a destination that we vary independently from the archive format...? For example an archive in mbtiles, pmtiles or individual files format can all end up in S3, as a file on disk, or some other destination. A naive pipeline using that approach would write to disk first then upload to s3 second, but a specialized s3-files implementation can skip writing to disk, and pmtiles could possibly upload chunks as its writing them so maybe they'd need to be separate anyway.

msbarry:
Also I see the amazon s3 dependency adds ~20% to the bundle size. My understanding is it introduces its own threading model as well. Is there any lower level dependency we can use instead that can just help constructing the upload API requests?

bbilger:
not sure why I forgot/overlooked that but https://github.com/minio/minio-java would probably the best compromise, to replace the aws sdk with - if we want to go such a route

PR #776, opened 2024-01-08
Remove log4j usage

cmahnke:
See https://github.com/onthegomap/planetiler/discussions/773#discussioncomment-8037396

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
@@ -2,6 +2,7 @@ package com.onthegomap.planetiler.expression;
 
 import static com.onthegomap.planetiler.expression.DataType.GET_TAG;
 
+import com.google.common.base.Joiner;
 import com.onthegomap.planetiler.reader.WithGeometryType;
 import com.onthegomap.planetiler.reader.WithTags;
 import com.onthegomap.planetiler.util.Format;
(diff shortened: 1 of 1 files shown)


cmahnke:
Please don't merge yet.

cmahnke:
I've checked locally and downloading there works, can you please rerun the failing checks?

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/Expression.java
```diff
@@ -362,7 +362,7 @@ static MatchAny from(String field, BiFunction<WithTags, String, Object> valueGet
 
       return new MatchAny(field, values,
         Set.copyOf(exactMatches),
-        patterns.isEmpty() ? null : Pattern.compile("(" + Strings.join(patterns, '|') + ")"),
+        patterns.isEmpty() ? null : Pattern.compile("(" + Joiner.on('|').join(patterns) + ")"),
```
bbilger, line 365: not sure if I miss anything but there's IMO no need for guava
would personally either do
`Pattern.compile("(" + String.join("|", patterns) + ")")`
or even "better"
`Pattern.compile(patterns.stream().collect(Collectors.joining("|", "(", ")")))`

msbarry: +1, slight preference to using built-in utilities.

cmahnke: Thanks for your feedback! I've included the proposed change.

PR #777, opened 2024-01-08
Replaced log4j usage with commons-io

cmahnke:
See https://github.com/onthegomap/planetiler/discussions/773#discussioncomment-8037396

This is a separate PR since it's introducing a new dependency which is certainly more controversial then #776  - especially since it's only used for one method.

The same goal is probably also reachable by returning an anonymous inline extension of some sort OutputStream wich just overwrites the `close` method...

### DIFF
--- a/planetiler-core/pom.xml
+++ b/planetiler-core/pom.xml
@@ -154,6 +154,11 @@
       <artifactId>geopackage</artifactId>
       <version>${geopackage.version}</version>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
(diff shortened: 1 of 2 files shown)


bbilger:
> The same goal is probably also reachable by returning an anonymous inline extension of some sort OutputStream wich just overwrites the `close` method...

ya, I would definitely do that and not pull in another dependency just for that. I just made use of `org.apache.logging.log4j.core.util.CloseShieldOutputStream` since it was there anyways. I'll work on getting rid of that later  this week.

cmahnke:
Closing due to https://github.com/onthegomap/planetiler/pull/777#issuecomment-1881972852

PR #780, opened 2024-01-10
Archive compare

msbarry:
Start to address #382 by adding a utility that compares the tiles in 2 archives. To run: `java -jar planetiler.jar compare archive1 archive2` To start with when the encoded tiles are not the same, it diffs layer IDs then feature IDs within those layers. Once those are consistent, it will diff more details on each matching feature.

PR #793, opened 2024-01-16
Fix comparison of multipolygons bounds check [#780]

bdon:
Fix crash when comparing multi polygons that have different # of member polygons


phanecak-maptiler (approved):

planetiler-core/src/main/java/com/onthegomap/planetiler/util/CompareArchives.java
```diff
@@ -366,7 +366,7 @@ private void compareGeometry(TileCoord coord, String layer, Geometry geom1, Geom
     compareValues(coord, layer, "feature JTS geometry type", geom1.getGeometryType(), geom2.getGeometryType());
     compareValues(coord, layer, "feature num geometries", geom1.getNumGeometries(), geom2.getNumGeometries());
     if (geom1 instanceof MultiPolygon) {
-      for (int i = 0; i < geom1.getNumGeometries(); i++) {
+      for (int i = 0; i < geom1.getNumGeometries() && i < geom2.getNumGeometries(); i++) {
```
msbarry, line 369: Could you add the same check on line 381?

Issue #794, opened 2024-01-16
[FEATURE] Add arbitrary extra metadata using the Profile interface


bdon:
**Is your feature request related to a problem? Please describe.**
The profile has a few interface methods like `name`, `attribution`, `version`, `isOverlay` but does not provide a way to add arbitrary other string:string pairs. 

**Describe the solution you'd like**
Another interface method called `extraMetadata` that is a `Map<String,String>`

**Describe alternatives you've considered**
* Using a lower-level interface instead of Profile might give a profile author access to `TileArchiveMetadata`, but that's more work.
* We can deprecate the individual methods `name`, `version`, `attribution` and instead make all metadata go through one `metadata` interface method that returns `Map<String,String>`

**Additional context**
We should consider how key conflicts between `extraMetadata` and `name`, `version`... as well as with build-time info `planetiler:githash` and run-time info (`osmosis:*`) choose which source gets priority.

bdon:
This was merged in https://github.com/onthegomap/planetiler/pull/795

PR #795, opened 2024-01-16
add extraMetadata to profile interface [#794]

bdon:


planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveMetadata.java
```diff
@@ -175,6 +177,14 @@ public TileArchiveMetadata withJson(TileArchiveMetadataJson json) {
...
    */
 
+  private static Map<String,String> mergeMaps(Map<String,String> m1, Map<String,String> m2) {
+    return Stream.concat(m1.entrySet().stream(), m2.entrySet().stream())
+      .collect(Collectors.toMap(
+        Map.Entry::getKey,
+        Map.Entry::getValue,
+        (value1, value2) -> value2));
+  }
+
```
msbarry, line 187: Might be slightly more straightforward to use putAll here, that makes it explicit which direction we are clobbering too:

```java
var result = new TreeMap<>(m1);
result.putAll(m2);
return result;
```

planetiler-core/src/main/java/com/onthegomap/planetiler/Profile.java
```diff
@@ -157,6 +158,8 @@ default boolean isOverlay() {
     return false;
   }
 
+  default Map<String,String> extraMetadata() { return Map.of(); }
```
msbarry, line 161: I wonder if we should include `archive` in the name here to make it clear where the metadata is going?

bdon: changed to `extraArchiveMetadata`

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveMetadata.java
```diff
@@ -93,7 +96,7 @@ public TileArchiveMetadata(Profile profile, PlanetilerConfig config, List<LayerA
       config.minzoom(),
       config.maxzoom(),
       vectorLayers == null ? null : new TileArchiveMetadataJson(vectorLayers),
-      mapWithBuildInfo(),
+      mergeMaps(mapWithBuildInfo(),profile.extraArchiveMetadata()),
```
msbarry, line 99: I think I'd expect the profile to provide defaults, then args to layer on top of that, so let's either reverse this or the behavior of mergeMaps?

bdon: shouldn't `mapWithBuildInfo` just be the `planetiler:*` metadata? If I define those in my profile (or command line args passed into my profile), they ought to override the defaults, I don't know why someone would do that though.

msbarry: Oh right! Yeah this looks good then. The command-line args get copied in just above this.

planetiler-core/src/main/java/com/onthegomap/planetiler/archive/TileArchiveMetadata.java
```diff
@@ -17,6 +17,9 @@
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
```
msbarry, line 22: remove unused imports

msbarry (commented):
looks good! let's just remove the unused imports then good to merge.

PR #802, opened 2024-01-20
Add whole-tile postprocess hook

msbarry:
Add a new `Map<String, List<VectorTile.Feature>> postProcessTileFeatures(TileCoord, Map<String, List<VectorTile.Feature>>)` hook that profiles can implement to post-process features across layers. For example:

- create a brand new layer that didn't exist before by combining features from other layers
- transform features in one layer based on features in another layer (like subtracting land polygons from water polygons, or adding tags to features that overlap water)

See #786

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Profile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Profile.java
@@ -1,6 +1,7 @@
 package com.onthegomap.planetiler;
 
 import com.onthegomap.planetiler.geo.GeometryException;
+import com.onthegomap.planetiler.geo.TileCoord;
 import com.onthegomap.planetiler.mbtiles.Mbtiles;
 import com.onthegomap.planetiler.reader.SourceFeature;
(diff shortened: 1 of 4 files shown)


wipfli:
Beautiful!

wipfli:
So if I have two layers, one called "landcover" and one called "climate zone", can I take the landcover polygons, intersect them with the climate zone polygons, and then tag the landcover polygons with the climate zone class that makes the biggest overlap?

wipfli:
Another use case I would like to try out for this: in some countries and regions there are not many road, but minzooms are usually done at a global level. So when you set minzooms that are good in Europe, the map might be quite empty in Africa. Could I use this new feature to compute road density in each tile and then drop roads accordingly?

msbarry:
You could definitely do the first one, just need to do some lower-level grouping by tags in java, then JTS operations. If it seems like a common pattern, we could pull it into a reusable utility in planetiler.

For the second one you _could_ emit more roads then you need during process step, then filter them out based on # roads in a tile in postprocess, but I'd be concerned about discontinuities at the tile seams?

PR #826, opened 2024-03-04
Encode OSM node/way/relation in vector tile feature id

msbarry:
Encode what source a feature comes from in its ID:

```
{vector tile feature id} = {original ID} * {feature_source_id_multiplier arg value} + {1 for OSM nodes, 2 for OSM ways, 3 for OSM relations, 0 for any other source}
```

By default `--feature-source-id-multiplier=10` but you can set to `--feature-source-id-multiplier=false` to disable.

See #824


zstadler:
We may consider adding a `feature_id()` method to the `OsmElement` interface and implement it differently in `Node`, `Way`, and `Relation`.
Then `Feature(... , source.feature_id())` can be used wherever `Feature(... , source.id())` is used in the current code.

PR #827, opened 2024-03-04
Encode 3x node/way/relation in vector tile feature id

msbarry:
Encode whether a vector tile feature comes from an OSM node, way, or relation in the feature ID according to the formula {vector tile feature id} = {osm id} * 3 + {0 for node, 1 for way, 2 for relation}. See https://github.com/onthegomap/planetiler/discussions/824 and compare to #826 which resulted in slightly larger tile sizes


msbarry:
closing in favor of #826

PR #844, opened 2024-03-19
Remove quoting side-bars from planetiler-custommap/README.md

zstadler:
Multiple levels of quoting side-bars were present in this README.md for no apparent reason.


msbarry:
This was to try to show that those contexts are nested inside each other so they each inherit from the parent. Maybe a nested bulleted list would work to illustrate that?

zstadler:
I wouldn't have guessed...

msbarry:
Looks good! May need to update when we can have multiple contexts that nest directly within a parent, but this works for now.

PR #874, opened 2024-04-26
exclude slf4j-nop from geopackage to avoid SLF4J warning

phanecak-maptiler:
Planetiler from current `main` is complaining a little bit:

```
SLF4J(W): Class path contains multiple SLF4J providers.
SLF4J(W): Found provider [org.apache.logging.slf4j.SLF4JServiceProvider@49097b5d]
SLF4J(W): Found provider [org.slf4j.nop.NOPServiceProvider@6e2c634b]
SLF4J(W): See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J(I): Actual provider is of type [org.apache.logging.slf4j.SLF4JServiceProvider@49097b5d]
```

This PR fixes that by excluding `org.slf4j:slf4j-nop` from `mil.nga.geopackage:geopackage`.

### DIFF
--- a/planetiler-core/pom.xml
+++ b/planetiler-core/pom.xml
@@ -153,6 +153,12 @@
       <groupId>mil.nga.geopackage</groupId>
       <artifactId>geopackage</artifactId>
       <version>${geopackage.version}</version>
+      <exclusions>
+      <exclusion>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-nop</artifactId>
+      </exclusion>
+    </exclusions>
     </dependency>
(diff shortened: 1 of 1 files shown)


phanecak-maptiler:
Note: As per SLF4J's recommendation, I've also reached out to the relevant component project and filled a bug-fix report, see:

- https://github.com/ngageoint/geopackage-java/issues/62

So either we update to fixed newer version or merge this.

msbarry:
Thanks! I just started noticing this too but wasn't sure where it was coming from.

PR #955, opened 2024-07-18
Update README.md to reflect change in demo tile hosting

quincylvania:
Vector tiles previously hosted by Americana are now hosted by OpenStreetMap US. CC: @ZeLonewolf

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,7 @@ for more of the backstory.
 
 ## Demo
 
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
Concur with these changes.

msbarry:
Thanks for the update! I'll need to update https://github.com/onthegomap/planetiler-demo as well.

PR #960, opened 2024-07-23
Expose vector tile feature ID method on `RelationInfo`

msbarry:


zstadler:
Thank you!
It works like charm!

zstadler:
Thanks!

PR #968, opened 2024-07-31
Expose caresAboutLayer() and new ForwardingProfile(config) ...

phanecak-maptiler:
... constructor so that profiles do not need to process `--only-layer` and `--exclude-layers` on their own, making sure we're working with `onlyLayers` and `excludeLayers` in `OpenMapTilesProfile` and `ForwardingProfile` in consistent manner.

This relates to https://github.com/openmaptiles/planetiler-openmaptiles/pull/182 .

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
@@ -55,7 +55,7 @@ public abstract class ForwardingProfile implements Profile {
   @SuppressWarnings("java:S3077")
   private volatile MultiExpression.Index<FeatureProcessor> indexedSourceElementProcessors = null;
 
-  public ForwardingProfile(PlanetilerConfig config) {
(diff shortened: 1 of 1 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
```diff
@@ -55,9 +55,13 @@ public abstract class ForwardingProfile implements Profile {
   @SuppressWarnings("java:S3077")
   private volatile MultiExpression.Index<FeatureProcessor> indexedSourceElementProcessors = null;
 
-  protected ForwardingProfile(PlanetilerConfig config, Handler... handlers) {
+  public ForwardingProfile(PlanetilerConfig config) {
```
msbarry, line 58: This should just be protected since only a subclass can ever call it.

phanecak-maptiler: Thanks, adjusted.

msbarry (approved):

PR #970, opened 2024-08-02
0.8.2 was released, so let's use 0.8.3 for SNAPSHOT

phanecak-maptiler:


pom.xml
```diff
@@ -28,7 +28,7 @@
     <sonar.projectKey>onthegomap_planetiler</sonar.projectKey>
     <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
     <sonar.exclusions>planetiler-benchmarks/**/*, planetiler-openmaptiles/**/*</sonar.exclusions>
-    <revision>0.8-SNAPSHOT</revision>
+    <revision>0.8.3-SNAPSHOT</revision>
```
msbarry, line 31: I'm still publishing snapshots as 0.8-SNAPSHOT. Not sure if that's the right way to do it or if I should switch to 0.8.3-SNAPSHOT but right now that doesn't exist.

phanecak-maptiler: I'm doing it sort of "out of habit" like this:

- we have 1.0.0 => version 1.0.0
- we're working on 1.0.1 => version 1.0.1-SNASPSHOT
- 1.0.1 relesed => version 1.0.1
- we're working on 1.1.0 => version 1.1.0-SNAPSHOT
- etc.

(But now I see that I did not bump planetiler-openmaptiles to 3.16-SNAPSHOT yet.)

I was not able to find some "good documentation" regarding this (e.g. some proper backing for my habit) but say looking at `maven` we can see:

- some beta release being done => "-SNAPSHOT" taken away from the version: https://github.com/apache/maven/commit/e92f645c2749eb2a4f5a8843cf01e7441e4b559f
- that beta release done, working on next beta release => version bumped up and "-SNAPSHOT" added back: https://github.com/apache/maven/commit/6e9889b1ade35f2d6e45a53b978a997e70f7742e

The main motivation here being the ability to somehow properly disambiguate between "0.8-SNASPHOT before 0.8.2" and "0.8-SNAPSHOT after 0.8.2" for PRs like:
1. https://github.com/openmaptiles/planetiler-openmaptiles/pull/182
2. https://github.com/openmaptiles/planetiler-openmaptiles/pull/184

"SNAPSHOT" implies (sort of) breakage without warning hence will not always help with the first example but will at least avoid confusion like in second example.

msbarry: Looks good, I'll bump planetiler to 0.8.3-SNAPSHOT: https://github.com/onthegomap/planetiler/pull/975

PR #971, opened 2024-08-02
Update old wikidata items

phanecak-maptiler:
When Planetiler is used with `--fetch-wikidata` it works roughly like this:

1. Planetiler run 1: `wikidata_names.json` does not exist, hence all the translations are fetched from WikiData
2. Some OSM data gets updated, e.g. some new items get added and some existing items get updated
3. Some WikiData items gets updated, e.g. some new items get added and some existing items get updated
4. Planetiler run 2: `wikidata_names.json` exists, so translations are loaded from it and only translations for new OSM elements are fetched from WikiData

Good thing is, that this lowers the load on WikiData and speeds-up tileset generation.

Problem: If some translation changed for existing OSM element, old value from `wikidata_names.json` is used. If we want to get updates, we can delete `wikidata_names.json` and fetch all translations once again.

This PR tries to partially address the problem without the need for deletion (or manual tweaking of) `wikidata_names.json`:

- It adds new option `wikidata_max_age` (with default value 0, e.g. "disabled")
- Is adds new option `wikidata_update_limit` (with default value 0, e.g. "disabled")
- It tweaks loading of translations during fetch phase to:
  - Skip up to `wikidata_update_limit` items which are older than `wikidata_max_age`
  - This then causes subsequent fetch to load those values once again

With the defaults Planetiler works as before.

When called with for example `--wikidata-max-age=P30D --wikidata-update-limit=100000`, it should then work roughly as follows:

1. Planetiler run 1: `wikidata_names.json` does not exist, hence all the translations are fetched from WikiData
2. Some OSM data gets updated, e.g. some new items get added and some existing items get updated
3. Some WikiData items gets updated, e.g. some new items get added and some existing items get updated
4. Planetiler run 2, say one month after 1st run: `wikidata_names.json` exists, so translations are loaded from it and only translations for new OSM elements are fetched from WikiData
    - Given `wikidata_max_age=P30D` all translations are now considered outdated
    - But given `wikidata_update_limit=100000` (which is roughly 5% of existing translations for full Planet) only up-to 100'000 translations are dropped and fetched from WikiData again
5. ...

Other combination might be `--wikidata-max-age=P30D` and `--wikidata-update-limit=0` which would keep using the cached translations for a month but the run after a month will drop all (now outdated) translations and fetch them again from WikiData.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -579,9 +579,8 @@ public class Planetiler {
     wikidataMaxAge =
       arguments.getDuration("wikidata_max_age",
         "Maximum age of Wikidata translations (in ISO-8601 duration format PnDTnHnMn.nS; 0S = disabled)", "P30D");
-    // limit of 100_000 is roughly 5% of total amount of entries for whole Planet
     wikidataUpdateLimit = arguments.getInteger("wikidata_update_limit",
-      "Limit on how many old translations to update during one download (0 = disabled)", 100_000);
+      "Limit on how many old translations to update during one download (0 = disabled)", 0);
     return this;
   }
 
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -40,6 +40,7 @@ import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -111,6 +112,8 @@ public class Planetiler {
   private boolean useWikidata = false;
   private boolean onlyFetchWikidata = false;
   private boolean fetchWikidata = false;
+  private Duration wikidataMaxAge = Duration.ZERO;
+  private int wikidataUpdateLimit = 0;
   private final boolean fetchOsmTileStats;
   private TileArchiveMetadata tileArchiveMetadata;
 
@@ -573,6 +576,11 @@ public class Planetiler {
         fetchWikidata);
     useWikidata = fetchWikidata || arguments.getBoolean("use_wikidata", "use wikidata translations", true);
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
+    wikidataMaxAge =
+      arguments.getDuration("wikidata_max_age", "Maximum age of Wikidata translations (0 = disabled)", "720H");
+    // limit of 100_000 is roughly 5% of total amount of entries for whole Planet
+    wikidataUpdateLimit = arguments.getInteger("wikidata_update_limit",
+      "Limit on how many old translations to update during one download (0 = disabled)", 100_000);
     return this;
   }
 
@@ -793,10 +801,15 @@ public class Planetiler {
     ensureInputFilesExist();
 
     if (fetchWikidata) {
-      Wikidata.fetch(osmInputFile(), wikidataNamesFile, config(), profile(), stats());
+      Wikidata.fetch(osmInputFile(), wikidataNamesFile, config(), profile(), stats(), wikidataMaxAge,
+        wikidataUpdateLimit);
     }
     if (useWikidata) {
-      translations().addFallbackTranslationProvider(Wikidata.load(wikidataNamesFile));
+      // update of old items handled by "fetch" above => do NOT use age checks here
+      var translationsProvider = Wikidata.load(wikidataNamesFile, Duration.ZERO, 0);
+      translationsProvider.clearUpdateTimes();
+      translations()
+        .addFallbackTranslationProvider(translationsProvider);
     }
     if (onlyDownloadSources || onlyFetchWikidata) {
       return; // exit only if just fetching wikidata or downloading sources
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -577,7 +577,8 @@ public class Planetiler {
     useWikidata = fetchWikidata || arguments.getBoolean("use_wikidata", "use wikidata translations", true);
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
     wikidataMaxAge =
-      arguments.getDuration("wikidata_max_age", "Maximum age of Wikidata translations (0 = disabled)", "720H");
+      arguments.getDuration("wikidata_max_age",
+        "Maximum age of Wikidata translations (in ISO-8601 duration format PnDTnHnMn.nS; 0S = disabled)", "P30D");
     // limit of 100_000 is roughly 5% of total amount of entries for whole Planet
     wikidataUpdateLimit = arguments.getInteger("wikidata_update_limit",
       "Limit on how many old translations to update during one download (0 = disabled)", 100_000);
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -578,7 +578,7 @@ public class Planetiler {
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
     wikidataMaxAge =
       arguments.getDuration("wikidata_max_age",
-        "Maximum age of Wikidata translations (in ISO-8601 duration format PnDTnHnMn.nS; 0S = disabled)", "P30D");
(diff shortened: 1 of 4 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/util/Wikidata.java
```diff
@@ -330,21 +352,27 @@ private LongObjectMap<Map<String, String>> queryWikidata(List<Long> qidsToFetch)
...
+    throws IOException {
+    final String updateTimeDefault = Instant.now().toString();
     for (LongObjectCursor<Map<String, String>> cursor : results) {
+      String updateTime =
+        updateTimes.containsKey(cursor.key) ? updateTimes.get(cursor.key).toString() : updateTimeDefault;
       writer.write(objectMapper.writeValueAsString(List.of(
         Long.toString(cursor.key),
-        cursor.value
+        cursor.value,
+        updateTime
```
msbarry, line 375: To make these a little more compact, can we store `updateTime.toEpochMilli()`?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -793,10 +801,15 @@ public void run() {
...
+      Wikidata.fetch(osmInputFile(), wikidataNamesFile, config(), profile(), stats(), wikidataMaxAge,
+        wikidataUpdateLimit);
     }
     if (useWikidata) {
-      translations().addFallbackTranslationProvider(Wikidata.load(wikidataNamesFile));
+      // update of old items handled by "fetch" above => do NOT use age checks here
+      var translationsProvider = Wikidata.load(wikidataNamesFile, Duration.ZERO, 0);
+      translationsProvider.clearUpdateTimes();
+      translations()
+        .addFallbackTranslationProvider(translationsProvider);
```
msbarry, line 812: Can you add another `Wikidata.load` method that retains the old behavior so this if block can just remain the same?  The normal behavior I'd expect calling "load" externally is "load everything" so maybe the "load with time limit" method should just be private/internal since it's an implementation detail of the "load for update" external API?

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Wikidata.java
```diff
@@ -110,12 +112,14 @@ private static LongObjectMap<Map<String, String>> parseResults(InputStream resul
    *
    * @throws UncheckedIOException if an error occurs
    */
-  public static void fetch(OsmInputFile infile, Path outfile, PlanetilerConfig config, Profile profile, Stats stats) {
+  public static void fetch(OsmInputFile infile, Path outfile, PlanetilerConfig config, Profile profile, Stats stats,
+    Duration maxAge, int updateLimit) {
+
```
msbarry, line 117: Can you also add a `fetch` method too that sets the new arguments to whatever they need to be set to to retain the old behavior?

planetiler-core/src/test/java/com/onthegomap/planetiler/util/WikidataTest.java
```diff
@@ -161,6 +171,43 @@ void testRetryFailedRequestOnce() throws IOException, InterruptedException {
...
+
+    var reader = new BufferedReader(new StringReader(json));
+    var translationsProvider = Wikidata.load(reader, maxAge, 0);
+    assertEquals(1, translationsProvider.getAll().size());
+  }
+
+  @Test
+  void testWikidataNamesJsonUpdateLimit() throws IOException {
+    Duration maxAge = Duration.ofSeconds(1);
+    Instant old = Instant.now().minus(maxAge).minus(maxAge);
```
msbarry, line 201: Let's not rely on current time in tests, instead can we make a test `Clock` at a fixed time and pass it into `static WikidataTranslations load(BufferedReader reader, ...` method?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -573,6 +576,11 @@ public Planetiler fetchWikidataNameTranslations(Path defaultWikidataCache) {
         fetchWikidata);
     useWikidata = fetchWikidata || arguments.getBoolean("use_wikidata", "use wikidata translations", true);
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
+    wikidataMaxAge =
+      arguments.getDuration("wikidata_max_age", "Maximum age of Wikidata translations (0 = disabled)", "720H");
+    // limit of 100_000 is roughly 5% of total amount of entries for whole Planet
+    wikidataUpdateLimit = arguments.getInteger("wikidata_update_limit",
+      "Limit on how many old translations to update during one download (0 = disabled)", 100_000);
```
msbarry, line 583: I think the default behavior should minimize load on wikidata infra, this seems like more of an advanced feature we'd want people to opt into. Can we make the default max age be 0?

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -573,6 +576,11 @@ public Planetiler fetchWikidataNameTranslations(Path defaultWikidataCache) {
         fetchWikidata);
     useWikidata = fetchWikidata || arguments.getBoolean("use_wikidata", "use wikidata translations", true);
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
+    wikidataMaxAge =
+      arguments.getDuration("wikidata_max_age", "Maximum age of Wikidata translations (0 = disabled)", "720H");
```
msbarry, line 580: This method does a naive conversion from string to duration:

```java
Duration parsed = Duration.parse("PT" + value);
```

duration an also handle "30D" except the string needs to be "P30D" or (a full example would be "P2DT3H4M"). If you want you could also update that method to do this conversion to make it a bit easier to use (and maybe call toUpperCase automatically to handle "30d") Or I could do that in a separate PR

msbarry (commented):
Looks good overall, a couple minor improvements.  At first I was a bit concerned this would just drop the first N translations by ID whereas it seems like we want it to drop them somewhat randomly - but it looks like since we write the translation file however it's ordered in a hashmap that it will be randomly sorted to begin with so dropping the first N will be random also 👍

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
phanecak-maptiler: Adjusted. `Duration.parse()` tweaked to be "backward compatible", so that values like `10s` still work.

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
phanecak-maptiler: Makes sense, adjusted.

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
phanecak-maptiler: Adjusted.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Wikidata.java
phanecak-maptiler: Adjusted.

planetiler-core/src/test/java/com/onthegomap/planetiler/util/WikidataTest.java
phanecak-maptiler: Adjusted.

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
```diff
@@ -573,6 +576,11 @@ public Planetiler fetchWikidataNameTranslations(Path defaultWikidataCache) {
         fetchWikidata);
     useWikidata = fetchWikidata || arguments.getBoolean("use_wikidata", "use wikidata translations", true);
     wikidataNamesFile = arguments.file("wikidata_cache", "wikidata cache file", defaultWikidataCache);
+    wikidataMaxAge =
+      arguments.getDuration("wikidata_max_age",
+        "Maximum age of Wikidata translations (in ISO-8601 duration format PnDTnHnMn.nS; 0S = disabled)", "P30D");
+    wikidataUpdateLimit = arguments.getInteger("wikidata_update_limit",
+      "Limit on how many old translations to update during one download (0 = disabled)", 0);
```
msbarry, line 583: It would be nice to only require one CLI parameter to enable this, it seems like the main one someone would want to control is the max age. Could we go back to the old update limit but set max age to 0 by default so you just need to set `--wikidata-max-age=P30D` to enable?

msbarry (commented):
One minor tweak to the default CLI options but otherwise looks good!

planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
phanecak-maptiler: Adjusted.

msbarry (approved):

msbarry:
Looks good! Thanks for adding this.

PR #975, opened 2024-08-08
Fix snapshot version

msbarry:
Bump snapshot version to be ahead of release 0.8.2


msbarry:
oops duplicate of #970

PR #987, opened 2024-08-19
Fix handling of string typed null values

phanecak-maptiler:
When using custom map as follows:

```
schema_name: Test Case Schema
schema_description: Test case tile schema
attribution: Test attribution
sources:
  osm:
    type: osm
    url: geofabrik:rhode-island
layers:
- id: testLayer
  features:
  - source:
    - osm
    geometry: polygon
    include_when:
      natural: water
    attributes:
    - key: non_existent
    - key: non_existent_typed
      type: string
```

then currently `non_existent` is handled differently than `non_existent_typed`:

- `non_existent` gives `null` value (since not present in source data) which gets properly handled so that resulting layer feature does NOT have a `non_existent` attribute
- `non_existent_typed` gives `null` value which gets converted into `"null"` string which is then handled so that resulting layer feature DOES have a `non_existent_typed` attribute with value `"null"`

Second case is inconsistent with the first and is considered a bug. This PR fixes that bug and adds unit test for this scenarios.

Other types seems to work as expected, e.g.:

```
...
    - key: non_existent_typed
      type: integer
```

generates features without `non_existent_typed` attribute.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/DataType.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/expression/DataType.java
@@ -10,7 +10,7 @@ import java.util.function.UnaryOperator;
  * Destination data types for an attribute that link the type to functions that can parse the value from an input object
  */
 public enum DataType implements BiFunction<WithTags, String, Object> {
-  GET_STRING("string", WithTags::getString, Objects::toString),
+  GET_STRING("string", WithTags::getString, Parse::parseStringOrNull),
   GET_BOOLEAN("boolean", WithTags::getBoolean, Parse::bool),
   GET_DIRECTION("direction", WithTags::getDirection, Parse::direction),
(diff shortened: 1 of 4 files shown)


msbarry (approved):
Good catch, thanks for the fix!

planetiler-core/src/main/java/com/onthegomap/planetiler/expression/DataType.java
```diff
@@ -10,7 +10,7 @@
  * Destination data types for an attribute that link the type to functions that can parse the value from an input object
  */
 public enum DataType implements BiFunction<WithTags, String, Object> {
-  GET_STRING("string", WithTags::getString, Objects::toString),
```
msbarry, line 13: Can you remove the unused `import java.util.Objects` line now?

phanecak-maptiler: Ah, sorry for that. Removed.

PR #991, opened 2024-08-21
added `dependsOnLayer()` ...

phanecak-maptiler:
... so that `caresAboutLayer()` can handle situations when some layer depends on another layer, e.g. for example in OpenMapTiles:

- <s>when 'transportation' is excluded, then also 'transportation_name' needs to be also treated as excluded</s>
- when 'transportation' is included (e.g. mentioned in --only-layers or not exluded) then 'transportation_name' needs to be also treated as included

This is a follow-up on #968.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
@@ -98,7 +98,7 @@ public abstract class ForwardingProfile implements Profile {
   /**
    * Indicate to {@link #caresAboutLayer(String)} what layers (dependents) rely on other layers (dependencies) so that
    * methods like {@link #registerFeatureHandler(FeatureProcessor)}, {@link #registerHandler(Handler)} or
-   * {@link #registerSourceHandler(String, FeatureProcessor)} do register also those dependencies.
+   * {@link #registerSourceHandler(String, FeatureProcessor)} do not block registration of those dependencies.
    * <p>
    * Dependencies are described as a map with dependent layer names as keys and lists containing dependency layer names
    * as map values. Example: Layer named {@code transportation_name} depends on layer named {@code transportation}:
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
@@ -85,6 +85,7 @@ public abstract class ForwardingProfile implements Profile {
       this.handlers.stream()
         .filter(HandlerForLayer.class::isInstance)
         .map(HandlerForLayer.class::cast)
+        .filter(l -> !l.name().equals(layer))
         .anyMatch(existing -> dependsOnLayer(existing.name(), layer));
   }
 
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
@@ -81,11 +81,19 @@ public abstract class ForwardingProfile implements Profile {
   }
 
   private boolean caresAboutLayer(String layer) {
-    return (onlyLayers.isEmpty() || onlyLayers.contains(layer)) && !excludeLayers.contains(layer);
+    if (excludeLayers.contains(layer) || excludeLayers.stream().anyMatch(excluded -> dependsOnLayer(layer, excluded))) {
+      return false;
+    }
+    return onlyLayers.isEmpty() || onlyLayers.contains(layer) ||
+      onlyLayers.stream().anyMatch(included -> dependsOnLayer(included, layer));
   }
 
   public boolean caresAboutLayer(Object obj) {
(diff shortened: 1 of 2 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
```diff
@@ -81,13 +81,21 @@ protected ForwardingProfile(Handler... handlers) {
   }
 
   private boolean caresAboutLayer(String layer) {
-    return (onlyLayers.isEmpty() || onlyLayers.contains(layer)) && !excludeLayers.contains(layer);
+    if (excludeLayers.contains(layer) || excludeLayers.stream().anyMatch(excluded -> dependsOnLayer(layer, excluded))) {
```
msbarry, line 84: I think this logic means exclude a layer if it is explicitly excluded, or if it depends on a layer that we exclude right? This means if you exclude transportation, it will exclude transportation_name as well.  The existing behavior is that if transportation_name is used (either via include or lack of exclude) then it forces transportation to be included. Either interpretation seems valid to me, but this is a change from the current behavior.

phanecak-maptiler: OK, adjusted to fix the bug (include `transportation` when `transportation_name` is used) but maintain old behavior (e.g. force `transportation` when needed even over exclude).

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
```diff
@@ -81,13 +81,21 @@ protected ForwardingProfile(Handler... handlers) {
   }
 
   private boolean caresAboutLayer(String layer) {
-    return (onlyLayers.isEmpty() || onlyLayers.contains(layer)) && !excludeLayers.contains(layer);
+    return ((onlyLayers.isEmpty() || onlyLayers.contains(layer)) && !excludeLayers.contains(layer)) ||
+      this.handlers.stream()
+        .filter(h -> h instanceof HandlerForLayer)
+        .map(h -> (HandlerForLayer) h)
```
msbarry, line 87: minor: filter-out this layer from the handlers?

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
```diff
@@ -81,13 +81,21 @@ protected ForwardingProfile(Handler... handlers) {
...
+        .anyMatch(existing -> dependsOnLayer(existing.name(), layer));
   }
 
   public boolean caresAboutLayer(Object obj) {
     return !(obj instanceof HandlerForLayer l) || caresAboutLayer(l.name());
   }
 
+  public boolean dependsOnLayer(String dependent, String dependency) {
```
msbarry, line 95: onlyLayers and exluceLayers returns lists, would it make sense for this one to return a `Map<String, List<String>>` of the dependencies?

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
msbarry, line 87: sonar wants this to be `.filter(HandlerForLayer.class::isInstance).map(HandlerForLayer.class::cast)` but I think the way you have it is more readable so feel free to ignore sonar here.

phanecak-maptiler: Adjusted.

phanecak-maptiler: Adjusted.

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
phanecak-maptiler: Adjusted. Should I alse rename? Say `dependsOnLayer()` -> `layerDependencies()`?

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
```diff
@@ -81,19 +81,22 @@ protected ForwardingProfile(Handler... handlers) {
...
+          existing -> dependencies.containsKey(existing.name()) && dependencies.get(existing.name()).contains(layer));
   }
 
   public boolean caresAboutLayer(Object obj) {
     return !(obj instanceof HandlerForLayer l) || caresAboutLayer(l.name());
   }
 
-  public boolean dependsOnLayer(String dependent, String dependency) {
-    return false;
+  public Map<String, List<String>> dependsOnLayer() {
```
msbarry, line 98: Minor: add javadoc to make it clear this should return a map from layer to list of layer names that it depends on (not the other way around)

msbarry (commented):
One minor comment otherwise looks good!

planetiler-core/src/main/java/com/onthegomap/planetiler/ForwardingProfile.java
phanecak-maptiler: javadoc added

msbarry (approved):

Issue #993, opened 2024-08-23
Implement a feature to remove elements that exceed the max_feature_count limit based on their spatial distribution and density.


CrazyBug-11:
It would be helpful to provide a max_feature_count parameter that can remove elements exceeding the limit based on the density within the tile, ensuring a more uniform appearance across the entire tile.

msbarry:
For points the label grid size/limit accomplishes this.  It divides each tile into a grid of a certain size then limits the number of features that can go in each cell.  Maybe if we just add the capability for this to apply to small polygons too that would accomplish what you're trying to do?

CrazyBug-11:
My dataset consists of polygon (area) data, not point data. The current issue is that setting min_size=0 results in too many features at lower zoom levels, which increases memory requirements. However, setting min_size=256/4096 causes some areas to be blank, and in such cases, the tile size at lower zoom levels inflates to 20-30MB. Therefore, I hope to have a feature that can evenly sample features within a tile, limit the number of features, and also control the tile size.

If source code modification is necessary, I would appreciate any advice you can provide. Thank you very much.

msbarry:
On each feature you can call `setPointLabelGridPixelSize` and `setPointLabelGridLimit` (there are a few aliases) and `FeatureRenderer.renderPoint` calls `getPointLabelGridPixelSizeAtZoom` and `getPointLabelGridLimitAtZoom` to get these values when generating vector tile features from the input feature.  We'd have to add similar handling to `renderLineOrPolygon` probably using either the first or center point of the polygon or line so that it passes the `groupInfo` parameter into `encodeAndEmitFeature`.  The rest of the pipeline should handle size/limit automatically, but we should add an end to end test to `PlanetilerTests` to be sure.

Once this is implemented it would be good to rename all of the `setPointLabelGrid...` methods to `setLabelGrid...` but we'd have to add the new methods, mark old as deprecated then remove later since it changes the external API.

Feel free to submit a PR, this would be useful for others as well. For example the openmaptiles profile should be limiting mountain_peak linestrings (cliffs and ridges) using label grid like this, but because it's not implemented we just end up with [very crowded tiles](https://onthegomap.github.io/planetiler-demo/#13.5/38.08269/-109.84131).

CrazyBug-11:
I have added the following code to calculate the centroid of each feature after executing sliceIntoTiles(). However, after testing, I found that the results did not change, and the label grid functionality is still not implemented as expected.

[...]

PR #996, opened 2024-08-24
feat: Polygon increase label grid limit(Beta version)

CrazyBug-11:
The current feature is incomplete, and the existing code does not achieve the expected results! [#993](https://github.com/onthegomap/planetiler/issues/993)


msbarry:
Thanks @CrazyBug-11 ! Would you mind giving me access to push to your fork?

CrazyBug-11:
Certainly, I've already invited you.

msbarry:
Thanks, but I'm pretty sure you just need to click the "allow edits from maintainers" button on the PR

CrazyBug-11:
Haha, got it, I've already checked the allow edits by maintainers

msbarry:
OK thanks! Pushed some end to end tests that should pass once this is working.

CrazyBug-11:
Thank you for your support. @msbarry 
would like to know why the first point of the feature is used. Based on my understanding, at the same zoom level, the first point of a feature should only be present in one tile. However, after executing sliceIntoTiles, the feature is split into multiple tiles. Why isn't the number of points limited by taking the vertices of each smaller feature after the split?

CrazyBug-11:
Thank you very much for your patient guidance.
I noticed that FeatureGroup.TileFeatures.add applies certain restrictions to features, but I couldn’t locate the logic that verifies if the current tile information matches the group information.For example, if a large feature is split across (1,0,0) and (1,1,0), is the group calculation performed twice within the same tile grid, or is it handled separately for each tile?
If you could point me to the specific part of the code where this is managed, that would be really helpful. I’m planning to carefully review the workflow in this section.

CrazyBug-11:
I have a dataset of over 300 million land-use polygons, the data is about 200g, and after testing, the results seem a bit strange. There are noticeable connection boundaries between tiles, and at certain zoom levels, you can clearly see the four borders of each tile. After further testing, I've realized that even if this functionality is implemented, it doesn't meet my current needs. Please forgive me for initially thinking in the wrong direction. I thought there would be a large boundary feature encompassing all the data, so even if some small features didn't hit, the boundary would serve as the background. However, in reality, there isn't such a large boundary feature, but rather many small polygons. When features are limited, it leads to many small areas being left blank.

What I actually need is to achieve a display where detailed polygon outlines are shown only at higher zoom levels, such as 11-14.
image
However, at lower zoom levels (0-10), the entire area should still be covered, achieving a similar effect to the following:
image

To ensure full coverage while keeping the tile size within approximately 1MB, my idea is to pixelate the polygons before executing encodeAndEmitFeature. For example, with a tile resolution of 256x256, I would calculate how many pixels each geom to be output occupies, and then replace the original feature with small squares for each pixel. The goal is to minimize the size of each feature while ensuring that the area is fully covered. Could you help me assess whether this approach is feasible? I really appreciate your help!

PR #1034, opened 2024-09-25
Fix comment about needing admin.osc.bz2

Zverik:
Was broken in 2dc4bc660cb525f0245874d1d3ba80eb45c9cf07


msbarry:
Thanks! Although daylight is going away so I'll need to remove these instructions soon 😞

PR #1064, opened 2024-10-11
All sources must use WGS84 coordinates

zstadler:
This seems to be assumed and not checked.

### DIFF
--- a/planetiler-custommap/README.md
+++ b/planetiler-custommap/README.md
@@ -58,7 +58,7 @@ A description that tells planetiler how to read geospatial objects with tags fro
 
 - `type` - Enum representing the file format of the data source, one
   of [`osm`](https://wiki.openstreetmap.org/wiki/PBF_Format), [`shapefile`](https://en.wikipedia.org/wiki/Shapefile),
-  or [`geopackage`](https://www.geopackage.org/).
+  or [`geopackage`](https://www.geopackage.org/). `shapefile` and `geopackage` sources must use the [WGS84 (EPSG:4326)](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84) coordinates, as `osm` already does.
 - `local_path` - Local path to the file to use, inferred from `url` if missing. Can be a string
   or [expression](#expression) that can reference [argument values](#arguments).
 - `url` - Location to download the file from if not present at `local_path`.


planetiler-custommap/README.md
```diff
@@ -58,7 +58,7 @@ A description that tells planetiler how to read geospatial objects with tags fro
 
 - `type` - Enum representing the file format of the data source, one
   of [`osm`](https://wiki.openstreetmap.org/wiki/PBF_Format), [`shapefile`](https://en.wikipedia.org/wiki/Shapefile),
-  or [`geopackage`](https://www.geopackage.org/).
+  or [`geopackage`](https://www.geopackage.org/). `shapefile` and `geopackage` sources must use the [WGS84 (EPSG:4326)](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84) coordinates, as `osm` already does.
```
msbarry, line 61: The java API does expose a projection field that you can override for shapefile and geopackage sources.  It would be pretty straightforward to add the ability to control that via a yaml config if you'd rather make that change in this PR instead?

See:

https://github.com/onthegomap/planetiler/blob/be7cec36db8d81001d674c01b1b4e712b46fd1e0/planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/Source.java#L7-L12

and:

https://github.com/onthegomap/planetiler/blob/be7cec36db8d81001d674c01b1b4e712b46fd1e0/planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredMapMain.java#L72-L78

zstadler: I think that automatically accepting other coordinate systems would be great.
Unfortunately, my JAVA coding skills are close to zero :-(

msbarry: I ended up just making projection configurable here: #1066

zstadler: Thanks!
Does it accept any projection or only the two [https://github.com/onthegomap/planetiler/blob/yaml-projection/planetiler-custommap/planetiler.schema.json#L60-L61](listed here)
```
                  "EPSG:3857",
                  "EPSG:4326"
```

msbarry: Those are examples of values so people can see the expected format. It can be any projection code.

zstadler:
This PR is superseded by https://github.com/onthegomap/planetiler/pull/1066

PR #1084, opened 2024-11-02
Add length/area units and CEL expression geometry accessors

### LINKED ISSUE #979: [FEATURE] Add basic geometry attributes for custom schema inline expressions
For OpenTrailMap I'd like to be able to be able to show the trail length. I'd also like to filter parks by size/coverage with MapLibre instead of with `min_size` in the tiles (gets around our need for #965 too).

I'd like to be able to do something like the below. I'm not sure about the best syntax or property names, or if these should be functions instead.

```
attributes: 
  - key: length
    value: '${ feature.length_km }'
  - key: area
    value: '${ feature.area_km2 }'
  - key: bbox_area
    value: '${ feature.bbox_area_km2 }'
  - key: min_lat
    value: '${ feature.min_lat }'
  - key: min_lon
    value: '${ feature.min_lon }'
  - key: max_lat
    value: '${ feature.max_lat }'
  - key: max_lon
    value: '${ feature.max_lon }'
```

For point features the area/length would return zero and the min/max lat/lon would just be the feature's lat/lon.

msbarry:
Add `Unit` class so that in java you can get the length or area of a feature in different units:

```java
feature.length(Unit.Length.METERS);
feature.length(Unit.Length.Z0_PIXELS);
feature.length(Unit.Length.NAUTICAL_MILES);
feature.area(Unit.Area.SQUARE_METERS);
feature.area(Unit.Area.SQUARE_Z0_TILES);
feature.area(Unit.Area.ACRES);
feature.area(Unit.Area.HECTARES);
```

And also expose a new "geometry" type to CEL expressions so they can access different attributes of feature geometries:

```
${feature.length("meters")}
${feature.area("km2")}
${feature.bbox.area("km2")}
${feature.centroid.lat} / ${feature.centroid.lon}
${feature.min_lat} / ${feature.max_lat} / ${feature.min_lon} / ${feature.max_lon}
${feature.centroid_if_convex.lat}
${feature.point_on_surface.lat}
${feature.innermost_point.lat}
${feature.validated_polygon.area('nm2')}
${feature.line_midpoint.lat}
${feature.point_along_line(0.25).lat}
${feature.partial_line(0, 0.25).length('m')}
```

Fixes #979

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/Unit.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/Unit.java
@@ -1,0 +1,214 @@
+package com.onthegomap.planetiler.geo;
+
+import com.onthegomap.planetiler.util.ToDoubleFunctionThatThrows;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import systems.uom.common.USCustomary;
+
+/** Units of length and area measurement based off of constants defined in {@link USCustomary}. */
+public interface Unit {
+
+  Pattern EXTRA_CHARS = Pattern.compile("[^a-z]+");
+  Pattern TRAILING_S = Pattern.compile("s$");
+
+  private static <T extends Unit> Map<String, T> index(T[] values) {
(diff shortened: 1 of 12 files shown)


quincylvania (approved):
Very exciting, thank you!

Issue #1092, opened 2024-11-08
[BUG] --simplify-tolerance-at-max-zoom=0 still does some simplification


davecraig:
We use the `--simplify-tolerance-at-max-zoom=0` option with the aim of preserving intersections at the maximum zoom level. However, we have noticed that it doesn't always work. The node I'm focussed on is https://www.openstreetmap.org/node/8869599302#map=19/55.941903/-4.316366 and when we run `planetiler` it's being dropped from Station Road.

I'm building a map for Scotland (obviously if I was a little more expert I could narrow this down!):
```
java -Xmx20g -jar planetiler-dist-0.8.3-SNAPSHOT-with-deps.jar --force --download --area=scotland --fetch-wikidata --output=scotland.pmtiles --nodemap-type=array --storage=mmap --maxzoom=15 --render_maxzoom=15 --simplify-tolerance-at-max-zoom=0
```

At zoom level 15 for tile 15991,10213 I would expect there to be 7 nodes remaining in the line representing Station Road, but instead there are only 6. The one that has been dropped is the intersection. Running `tileserver-gl-light` to inspect the results and zooming right in the gap is almost imperceptible, but there definitely is one.

If I alter `DouglasPeuckerSimplifier` so that it skips processing if the `distanceTolerance` is 0 or less:
```
  public static Geometry simplify(Geometry geom, double distanceTolerance) {
    if (geom.isEmpty() || (distanceTolerance <= 0.0)) {
      return geom.copy();
    }

    return (new DPTransformer(distanceTolerance)).transform(geom);
  }
```

then the output file changes, growing from 519230972 to 519316026 bytes. This suggests that there are a lot of other simplifications going on elsewhere in the map despite the expectation that simplification is turned off. On the upside this change also means that a value of `--simplify-tolerance-at-max-zoom=-1` also now works - previously the value was being squared before being used which negated its effect.

I'll make a PR with this change. However, it obviously doesn't address the underlying issue which I presume is precision related in some way. (Apologies for not delving further, but I haven't yet figured out the best way to run a debugger on the code to look more closely as I'm really a C++ embedded linux developer and Maven is still a bit mysterious to me).

Many Thanks.

PR #1093, opened 2024-11-08
Shortcut through simplification step if distanceTolerance is negative

davecraig:
If the simplification tolerance has been set to less than 0.0 then there's no need to run the DouglasPeuckerSimplifier algorithm. This change checks the tolerance and simply returns a copy of the Geometry if no simplification is requested.
This change is required for the case where the user wants no simplification to take place e.g. to preserve all intersection nodes. Without this change any negative value for distanceTolerance is squared inside DPTransformer and immediately loses its meaning.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/DouglasPeuckerSimplifier.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/DouglasPeuckerSimplifier.java
@@ -38,7 +38,7 @@ public class DouglasPeuckerSimplifier {
    * @return the simplified geometry
    */
   public static Geometry simplify(Geometry geom, double distanceTolerance) {
-    if (geom.isEmpty()) {
+    if (geom.isEmpty() || (distanceTolerance < 0.0)) {
       return geom.copy();
     }
 
--- a/planetiler-core/src/test/java/com/onthegomap/planetiler/FeatureMergeTest.java
+++ b/planetiler-core/src/test/java/com/onthegomap/planetiler/FeatureMergeTest.java
@@ -165,6 +165,21 @@ class FeatureMergeTest {
         true
       )
     );
+    // but doesn't resimplify if the tolerance is negative even when resimplify=true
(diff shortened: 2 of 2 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/geo/DouglasPeuckerSimplifier.java
```diff
@@ -38,7 +38,7 @@ public class DouglasPeuckerSimplifier {
    * @return the simplified geometry
    */
   public static Geometry simplify(Geometry geom, double distanceTolerance) {
-    if (geom.isEmpty()) {
+    if (geom.isEmpty() || (distanceTolerance < 0.0)) {
```
msbarry, line 41: Should this be <= 0 so distance tolerance=0 has this same effect?

msbarry: Actually simplifying with tolerance=0 to remove colinear points is different from not simplifying at all so this is probably what we want. One nice ease-of-use improvement might be to make the arg so that you can do `--simplify-tolerance-at-max-zoom=false` or `=off` or something to turn it off instead of `=-1`

davecraig: The problem with `tolerance=0` is that it does still perform some simplification. I had initially filed #1092 before deciding that it was actually doing the right thing. My understanding is that if a node is EXACTLY on a line then `tolerance=0` will still remove it. There's certainly a difference in the output between using 0 and using -1.

davecraig: No problem at all, thanks for all your work on planetiler. It's proving invaluable for our port of a navigation app for the visually impaired. We're using the protomaps output from planetiler to drive both a maplibre graphical UI and an audio descriptive UI.

PR #1104, opened 2024-11-17
Add lz4 compression to planetiler

flymarq:
- lz4 compressed files can be generated with libosmium based tools like osmium or python scripts using pyosmium like pyosmium-up-to-date
- lz4 compressed files are bigger (~15-20%) then zlib compressed files but much faster to produce and process
- this patch is in use by us for several month without any problems


planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PbfDecoder.java
```diff
@@ -74,6 +78,20 @@ private static byte[] readBlobContent(Fileformat.Blob blob) {
         throw new FileFormatException("PBF blob contains incomplete compressed data.");
       }
       inflater.end();
+    } else if (blob.hasLz4Data()) {
```
msbarry, line 81: Thanks for adding this! Would you mind adding a test case in `PlanetilerTests` that verifies end-to-end reading from an lz4 pbf file? We have an existing test file `planetiler-core/src/test/resources/monaco-latest.osm.pbf` - if you could just convert that to lz4 and update `testPlanetilerRunner` (or add a similar test) to use it then that should cover it, thanks!

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PbfDecoder.java
```diff
@@ -74,6 +78,20 @@ private static byte[] readBlobContent(Fileformat.Blob blob) {
...
+        int decompressedLength2 = decompressor.decompress(blob.getLz4Data().toByteArray(), 0, blobData, 0, decompressedLength);
+        if (decompressedLength != decompressedLength) {
+          // System.out.println("uncompressed size mismatch: "+decompressedLength2+" != "+decompressedLength);
+          throw new FileFormatException("Unable to decompress PBF blob. uncompressed size mismatch");
+        }
+      } catch (LZ4Exception e) {
+        throw new FileFormatException("Unable to decompress PBF blob.", e);
+      }
     } else {
       throw new FileFormatException("PBF blob uses unsupported compression, only raw or zlib may be used.");
```
msbarry, line 96: Add lz4 to this error message

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/PbfDecoder.java
```diff
@@ -74,6 +78,20 @@ private static byte[] readBlobContent(Fileformat.Blob blob) {
...
       }
       inflater.end();
+    } else if (blob.hasLz4Data()) {
+      final int decompressedLength = blob.getRawSize();
+      LZ4Factory factory = LZ4Factory.fastestInstance();
+      LZ4FastDecompressor decompressor = factory.fastDecompressor();
+      blobData = new byte[decompressedLength];
+      try {
+        int decompressedLength2 = decompressor.decompress(blob.getLz4Data().toByteArray(), 0, blobData, 0, decompressedLength);
+        if (decompressedLength != decompressedLength) {
```
msbarry, line 88: `decompressedLength2 != decompressedLength` ?

flymarq:
Thank you for fixing all open issues - I didn't find free time to do it so quickly.

I successfully tested the current main branch without any additional patches on a LZ4 compressed planet file.

Issue #1043, opened 2024-09-29
[FEATURE] Don't filter languages


tbodt:
**Is your feature request related to a problem? Please describe.**
The default planetiler OpenMapTiles-like config includes about 80 languages, which leaves out the long tail of hundreds of languages that have labels on Wikidata.

**Describe the solution you'd like**
Ideally, the default would be to suck in all language data without filtering. If the goal of language filtering is to save space, well it can't have saved that much space, since the uncommon languages by definition don't have a lot of labels! So would need to do an experiment to be sure, but simply pulling in all language data probably wouldn't inflate the tileset too much. And that marginal increase would be worth it, I think, for giving underrepresented languages instant access to maps in their language, with no requirements beyond typing labels into wikidata.

**Describe alternatives you've considered**
Adding languages one by one as they are requested, but that seems kind of like a pointless barrier

msbarry:
Hello! The default version should match OpenMapTiles defaults, hence this language list. Seems like it would make sense to add an `--all-languages` command line arg you can use to opt-into all languages without any filtering though.

PR #1111, opened 2024-11-24
Add more flexibility to "languages" flag

tbodt:
With this, the flag supports:

- "default" to append the default set of languages, instead of overriding it entirely, e.g. --languages=default,tlh means all default languages *plus* Klingon.
- "-lang" to exclude languages, e.g. --languages=default,-en means all default languages *except* English.
- "*" to include all languages not specified. This can be combined with "-lang". e.g. --languages=*,-jbo means all languages *except* Lojban.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Translations.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/util/Translations.java
@@ -22,15 +22,31 @@ public class Translations {
     ThreadLocal.withInitial(() -> new ThreadLocalTransliterator().getInstance("Any-Latin"));
 
   private boolean shouldTransliterate = true;
-  private final Set<String> languageSet;
   private final List<TranslationProvider> providers = new ArrayList<>();
+  private final Set<String> includeLanguages;
+  private final Set<String> excludeLanguages;
+  private boolean defaultInclude = false;
 
   private Translations(List<String> languages) {
-    this.languageSet = new HashSet<>();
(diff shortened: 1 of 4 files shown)


msbarry:
Cool! Thanks for adding. 

Fixes #1043

planetiler-core/src/main/java/com/onthegomap/planetiler/util/Translations.java
```diff
@@ -22,15 +22,31 @@ public class Translations {
...
+  private final Set<String> excludeLanguages;
+  private boolean defaultInclude = false;
 
   private Translations(List<String> languages) {
-    this.languageSet = new HashSet<>();
+    this.includeLanguages = new HashSet<>();
+    this.excludeLanguages = new HashSet<>();
+
     for (String language : languages) {
+      if (language.equals("*")) {
```
msbarry, line 35: Minor but what about supporting `"all"` instead of or in addition to `"*"` so that you don't need to worry about bash expanding it as a wildcard?

tbodt: Good idea, but `all` is a language code assigned to https://en.wikipedia.org/wiki/Allar_language. Any better ideas for a keyword?

msbarry: Ah, good catch! I can't think of any better short name than "all" so `"*"` seems good for now!

msbarry:
Thank you!

PR #1117, opened 2024-11-26
Make valid language tag regex case sensitive

msbarry:
Address https://github.com/onthegomap/planetiler/pull/1115#discussion_r1857682616


1ec5 (approved):
Thanks! This takes care of the cases I was concerned about.

PR #1118, opened 2024-11-27
GeometryPipeline

msbarry:
This PR adds a new `GeometryPipeline` interface for operations that transform a `Geometry` to another `Geometry` and also:

- add new `feature.transformScaledGeometry(GeometryPipeline)` which tells planetiler to apply that transform to the feature each time it is scaled to tile coordinates (where 1=width of tile at that zoom level) before slicing the feature into tiles at that zoom
- add new `feature.transformScaledGeometryByZoom((zoom) -> GeometryPipeline)` which lets you dynamically change the transform applied to scaled geometries at each zoom level
- update `FeatureMerge` `mergeLineStrings()` and `mergeNearbyPolygons` to accept a geometry pipeline to apply to each merged feature in tile coordinates (where 256=width of the tile)
- Adapt `DouglasPeuckerSimplifier` and `VWSimplifier` to implement `GeometryPipeline`
- Add `MidpointSmoother` to implement [midpoint smoothing](https://github.com/wipfli/midpoint-smoothing)
- Add `DualMidpointSmoother` to implement [Chaikin smoothing](https://observablehq.com/@pamacha/chaikins-algorithm)
- Add static factory methods to `GeometryPipeline` for these common transforms (simplifyDP, simplifyVW, smoothChaikin, smoothMidpoint)

You can also pass simple functions where a GeometryPipeline is expected, for example:

```java
feature.transformScaledGeometry(GeometryFixer::fix)
feature.transformScaledGeometry(Geometry::getCentroid)
feature.transformScaledGeometry(Geometry::getInteriorPoint)
```

And you can chain transforms using `andThen`:

```java
feature.transformScaledGeometry(
  GeometryPipeline.simplifyVW(1).setWeight(0.9)
    .andThen(GeometryPipeline.simplifyDP(1))
);
```

⚠️ transformScaledGeometry replaces default simplification with the provided transform, so you need to manually add simplification back if you want it with `.andThen(GeometryPipeline.defaultSimplify(feature))`
⚠️ transformScaledGeometry and FeatureMerge work with geometries at different scales (tile width=1 in transformScaledGeometry vs. tile width=256 for FeatureMerge) so you can't reuse a pipeline between feature processing and tile post-processing

Here is a joke example using Visvalingam Whyatt simplification and Chaikin smoothing on [OSM ocean polygons](https://osmdata.openstreetmap.de/data/water-polygons.html):

image

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
@@ -749,7 +749,7 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
      * features. This function gets run instead of simplification, so should include any simplification if you want
      * that.
      */
-    public Feature setGeometryPipeline(GeometryPipeline pipeline) {
+    public Feature transformScaledGeometry(GeometryPipeline pipeline) {
       this.defaultGeometryPipeline = pipeline;
       return this;
     }
@@ -759,7 +759,7 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
      * vector tile features. These functions get run instead of simplification, so should include any simplification if
      * you want that.
      */
-    public Feature setGeometryPipelineByZoom(ZoomFunction<GeometryPipeline> overrides) {
+    public Feature transformScaledGeometryByZoom(ZoomFunction<GeometryPipeline> overrides) {
       this.geometryPipelineByZoom = overrides;
       return this;
     }
@@ -768,7 +768,7 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
      * Returns the geometry transform function to apply to scaled geometries at {@code zoom}, or null to not update them
      * at all.
      */
-    public GeometryPipeline getGeometryPipelineAtZoom(int zoom) {
+    public GeometryPipeline getScaledGeometryTransformAtZoom(int zoom) {
       return ZoomFunction.applyOrElse(geometryPipelineByZoom, zoom, defaultGeometryPipeline);
     }
 
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureCollector.java
@@ -507,8 +507,8 @@ public class FeatureCollector implements Iterable<FeatureCollector.Feature> {
     private SimplifyMethod defaultSimplifyMethod = SimplifyMethod.DOUGLAS_PEUCKER;
     private ZoomFunction<SimplifyMethod> simplifyMethod = null;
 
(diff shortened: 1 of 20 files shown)


wipfli:
Amazing, thanks for introducing this! I am playing a bit around with it and post things that come to mind below...

wipfli:
First question, if I do

```java
features.polygon("forest")
    .setPixelTolerance(0.0)
    .setMinPixelSize(0.0)
    .transformScaledGeometry(GeometryPipeline.simplifyDP(2.0 / 256));
```

in the processFeature function, will setPixelTolerance have any effect?

wipfli:
Chaikin works nicely

```java
features.polygon("waterway")
    .setPixelTolerance(0.0)
    .transformScaledGeometry(GeometryPipeline.smoothChaikin(3))
    .setMinPixelSize(0.0);
```

chaikin

wipfli:
One thing that I always found a little bit disturbing about Douglas Peucker simplification was that you get these sharp corners all over the place. Would the recommended way to remove those be something like this?

```java
features.polygon("waterway")
    .setPixelTolerance(0.0)
    .transformScaledGeometry(
        GeometryPipeline.simplifyDP(1/256d)
            .andThen(GeometryPipeline.smoothChaikin(1))
    )
    .setMinPixelSize(0.0);
}
```

msbarry:
Thanks for taking a look!

> First question, if I do
> 
> ```java
> features.polygon("forest")
>     .setPixelTolerance(0.0)
>     .setMinPixelSize(0.0)
>     .transformScaledGeometry(GeometryPipeline.simplifyDP(2.0 / 256));
> ```
> 
> in the processFeature function, will setPixelTolerance have any effect?

transformScaledGeometry replaces simplification, but I added a utility so you can do `pipeline.andThen(GeometryPipeline.defaultSimplify(feature))` if you want to add back in the simplification that would have happened automatically. Do you think that makes sense?


> One thing that I always found a little bit disturbing about Douglas Peucker simplification was that you get these sharp corners all over the place. Would the recommended way to remove those be something like this?
> 
> ```java
> features.polygon("waterway")
>     .setPixelTolerance(0.0)
>     .transformScaledGeometry(
>         GeometryPipeline.simplifyDP(1/256d)
>             .andThen(GeometryPipeline.smoothChaikin(1))
>     )
>     .setMinPixelSize(0.0);
> }
> ```

Yes that could work! You end up adding points back to remove the sharp corners. You could also add points back by just using a smaller DP tolerance, but adding them back with chaikin smoothing results in a globbier shape.

You could also try `setMaxVertexArea`/`setMaxVertexOffset` which limit how far chaikin smoothing is allowed to move the smoothed shape away from the original shape, and `setMinVertexArea`/`setMinVertexOffset` to stop smoothing when the corners get too small instead of using a fixed number of iterations.

msbarry:
@wipfli I'm going to merge this now, let me know if you have any feedback using it on main branch and we can adjust if necessary!

PR #1136, opened 2024-12-25
Keep descriptions of builtin args

zstadler:
when overridden in custom map `args`


zstadler:
I could not find an object that contains the descriptions of the builtin args. They seem to be provided as parameters to `arguments.get*` calls. For example:

https://github.com/onthegomap/planetiler/blob/4c2e187ea3dc148d5c6e54c758b60969d2e78898/planetiler-core/src/main/java/com/onthegomap/planetiler/config/PlanetilerConfig.java#L122

PR #1147, opened 2025-01-14
Add support for geojson

HarelM:
- Related to: #46

This is a very similar to #413.
It uses geotools' geojson capabilities (very similar to shapefile).
 I've added some minimal tests, I think it might be enough, but I'm not sure.
I was able to run the tests I added but I'm getting an error when running `mvn clean install` on the entire project.
Let me know if anything is missing.

### DIFF
--- a/planetiler-core/pom.xml
+++ b/planetiler-core/pom.xml
@@ -66,6 +66,16 @@
       <version>${geotools.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-geojson</artifactId>
+      <version>${geotools.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-geojson-store</artifactId>
+      <version>${geotools.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.xerial</groupId>
       <artifactId>sqlite-jdbc</artifactId>
       <version>3.47.2.0</version>
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
@@ -24,7 +24,7 @@ public class GeoJsonReader extends SimpleReader<SimpleFeature> {
     GeoJsonReader(String sourceName, Path input) {
         super(sourceName);
         store = new GeoJSONDataStore(input.toFile());
-        layer = input.getFileName().toString().replaceFirst("\\.+", ""); // remove file extention.
+        layer = input.getFileName().toString().replaceFirst("\\.[^.]+$", ""); // remove file extention.
     }
 
     /**
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
@@ -78,9 +78,7 @@ public class GeoJsonReader extends SimpleReader<SimpleFeature> {
                 var properties = feature.getProperties();
                 SimpleFeature simpleFeature = SimpleFeature.create((Geometry) feature.getDefaultGeometry(), HashMap.newHashMap(properties.size()),
                 sourceName, layer, id++);
-                properties.forEach(property ->
-                    simpleFeature.setTag(property.getName().toString(), property.getValue());
-                );
+                properties.forEach(property -> simpleFeature.setTag(property.getName().toString(), property.getValue()));
                 next.accept(simpleFeature);
             }
         }
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
@@ -1,0 +1,87 @@
+package com.onthegomap.planetiler.reader;
+
+import com.onthegomap.planetiler.Profile;
+import com.onthegomap.planetiler.collection.FeatureGroup;
+import com.onthegomap.planetiler.config.PlanetilerConfig;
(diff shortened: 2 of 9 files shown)


HarelM:
I think the tests are failing due to download issue from geofabrik:
```
Error: Exception in thread "main" java.lang.IllegalStateException: java.net.ConnectException: Connection refused
	at com.onthegomap.planetiler.util.Geofabrik.getAndCacheIndex(Geofabrik.java:59)
```
I'm not sure this is related to the changes I made...

planetiler-core/pom.xml
```diff
@@ -65,6 +65,16 @@
...
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-geojson</artifactId>
+      <version>${geotools.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-geojson-store</artifactId>
+      <version>${geotools.version}</version>
+    </dependency>
```
msbarry, line 77: I see JTS already has a [GeoJsonReader](https://github.com/locationtech/jts/blob/master/modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonReader.java) class. Geotools has some licensing issues so I'm trying to avoid adding new dependencies on it.  But the JTS reader doesn't seem to handle properties at all so this might be the best one to start with...

msbarry (commented):
Thanks for adding this! Newline-delimited geojson is also a common format supported by tools like tippecanoe, do you think there's any way for this to support that as well?

planetiler-core/pom.xml
HarelM: Yes, JTS is only for geometries, it doesn't handle feature and feature collection...
Geotools uses it under the hood...

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
```diff
@@ -0,0 +1,87 @@
...
+        long id = 0;
+        String typeName = store.getTypeNames()[0];
+        SimpleFeatureCollection features = store.getFeatureSource(typeName).getFeatures();
+
+        try (var iter = features.features()) {
+            while (iter.hasNext()) {
+                var feature = iter.next();
+                SimpleFeature simpleFeature = SimpleFeature.create((Geometry) feature.getDefaultGeometry(), HashMap.newHashMap(feature.getProperties().size()),
+                sourceName, layer, id++);
+                feature.getProperties().forEach(property -> {
```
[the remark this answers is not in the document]
HarelM: Sure, I'll fix that.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+public class GeoJsonReader extends SimpleReader<SimpleFeature> {
+
+    private final GeoJSONDataStore store;
+    private final String layer;
+
+    GeoJsonReader(String sourceName, Path input) {
+        super(sourceName);
+        store = new GeoJSONDataStore(input.toFile());
+        layer = input.getFileName().toString().replaceAll("\\.geojson$", "");
```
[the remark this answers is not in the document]
HarelM: Sure, no problem.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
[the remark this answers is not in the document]
HarelM: Fixed in https://github.com/onthegomap/planetiler/pull/1147/commits/636dcd1c73ba6526b118bff715f11cdf39ceadd1

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoJsonReader.java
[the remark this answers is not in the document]
HarelM: Fixed in https://github.com/onthegomap/planetiler/pull/1147/commits/24ec3ba129f688e5bb466d72bfd555a40a62c6a9

HarelM:
For people using newline geojson it should be an easy task to add the surrounding feature collection I hope before running planetiler.
As far as I can tell geotools doesn't support newline geojson, but one could easily test this using the code I added here and find out.
We could try and create an issue there, not sure if it will help, we can also try and find a different geojson parser, I don't know, I'm not an expert in the Java ecosystem for GIS...

msbarry:
OK thanks! The main benefit of newline-delimited geojson is if the dataset is very large, it can be read line-by-line instead of loading the whole thing in memory.  I could take a stab at auto-detecting which kind it is and handling newline-delimited variant in a followup PR...

planetiler-core/pom.xml
msbarry: Does the geotools version automatically handle parsing/converting CRS of input geojson?

HarelM: No clue, probably not, it is very similar to shapefile, I belive, in terms of API.

HarelM: I have tried to play with the Crs object of the store in various way but it seems to always return WGS84 even if the file said otherwise.
I've opened a ticket here:
https://osgeo-org.atlassian.net/browse/GEOT-7705
This can be read from the file itself, but it seems hacky...

msbarry: Ok thanks! The jts version handles crs but not properties, and geotools version handles properties but not crs 🤦‍♂️ if there's just one crs defined for the entire object it wouldn't be too bad to read it out like jts does, but we could also merge this version then handle crs and newline-delimited in a follow-up pr?

HarelM: Sure, I think this is a good addition as is. We can always improve in the future.

msbarry:
Looks good now! Thanks for adding!

HarelM:
Great! Let me know when a new version with this is available.

PR #1166, opened 2025-02-04
Fix creation of CRS with longitudeFirst in GeoPackageReader

### LINKED ISSUE #1165: [BUG] Geopackage with non-geographic input, outputs geometries with inverted X/Y axis
**Describe the bug**
The GeoPackageReader projects non-geographic sources incorrectly. 

**To Reproduce**
* Convert a non-geographic source such as water polygons in web mercator https://osmdata.openstreetmap.de/data/water-polygons.html EPSG:3857 to GeoPackage: `ogr2ogr -f GPKG water.gpkg input.shp`
* Tile this as input using a profile, e.g.

```java
  @Override
  public void processFeature(SourceFeature sourceFeature, FeatureCollector features) {
    features.polygon("water").setZoomRange(0,8);
  }
```

```java
    Planetiler.create(args)
      .setProfile(new Water())
      .addGeoPackageSource("landcover", Path.of("data", "sources", "water.gpkg"), "https://example.com")
      .overwriteOutput(Path.of("data", "water.pmtiles"))
      .run();
```

**Expected behavior**
The output geometries are not tiled correctly, the X and Y axis are inverted. 

Image

**Environment (please complete the following information):**

- Hardware: M1 MacBook Pro
- OS: MacOS 15.2
- Java OpenJDK Runtime Environment Temurin-22.0.2+9
- Maven version: 3.9.9

Looks like the CRS initialization needs to look like  `CRS.decode("EPSG:4326", true);` as is done in ShapefileReader.

bdon:
Fixes #1165 

Not quite sure where we need these, but this seems to fix the initial issue

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/GeoPackageReader.java
@@ -46,7 +46,7 @@ public class GeoPackageReader extends SimpleReader<SimpleFeature> {
     if (sourceProjection != null) {
       try {
         var sourceCRS = CRS.decode(sourceProjection);
-        var latLonCRS = CRS.decode("EPSG:4326");
+        var latLonCRS = CRS.decode("EPSG:4326", true);
         coordinateTransform = CRS.findMathTransform(sourceCRS, latLonCRS);
       } catch (FactoryException e) {
         throw new FileFormatException("Bad reference system", e);
@@ -123,7 +123,7 @@ public class GeoPackageReader extends SimpleReader<SimpleFeature> {
 
   @Override
   public void readFeatures(Consumer<SimpleFeature> next) throws Exception {
-    var latLonCRS = CRS.decode("EPSG:4326");
+    var latLonCRS = CRS.decode("EPSG:4326", true);
     long id = 0;
     boolean loggedMissingGeometry = false;
 


msbarry (commented):
Thanks! looks like this makes the tests fail though... does that mean the geopackage files/tests were incorrect in the first place? I'd be curious if anyone is using the geopackage reader now and it works correctly? If so this might break things for them...

bdon:
@wipfli are the geometries in the geo package 4326 or 3857?

wipfli:
> @wipfli are the geometries in the geo package 4326 or 3857?

4326

wipfli:
[0.gpkg.txt](https://github.com/user-attachments/files/18688946/0.gpkg.txt)

```
ogrinfo -al -summary 0.gpkg

INFO: Open of `0.gpkg'
      using driver `GPKG' successful.

Layer name: out
Geometry: Polygon
Feature Count: 217

Extent: (-162.150000, -54.925000) - (178.475000, 72.025000)
Layer SRS WKT:
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
        SCOPE["unknown"],
        AREA["World"],
        BBOX[-90,-180,90,180]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
FID Column = fid
Geometry Column = geom
DN: Integer (0.0)
```

bdon:
Can you try with a different CRS? like https://osmdata.openstreetmap.de/data/land-polygons.html

msbarry:
`Data axis to CRS axis mapping: 2,1`

according to chatgpt this means that your input source is longitude-first, so without @bdon 's change it should be rotated 90 degrees?

msbarry:
I'm almost inclined to make this fix and break some existing use-cases but provide a shorthand way to set longitude first (like `EPSG:4326:lon_first` or something).  In @wipfli's case it would probably work without specifying a CRS since it would pick up the right one from the input file.

What do you think @wipfli @bdon ?

wipfli:
I think planetiler should read the crs and the order of longitude and latitude from the geopackage file.

PR #1173, opened 2025-02-08
Normalize merged fill polygons

### LINKED ISSUE #1170: [FEATURE] canonical vector tile polygon representations for deduplication
Deduplication of repetitive tiles makes all of the tile archive formats more efficient, usually this works by hashing the tile contents to determine if they are byte-for-byte identical. These tiles are almost always oceans or large areal features where there is a single square covering the whole tile.

Right now, it is possible for the tiling to emit polygons that are geometrically identical but have different vertex representations, such as these two geometries from `-128,-128` to `4224, 4224`:

```
geometry:
      RING[count=5](-128 -128,4224 -128,4224 4224,-128 4224,-128 -128)[OUTER]
```

```
geometry:
  RING[count=9](-128 4224,-128 -128,4093 -128,4099 -128,4224 -128,4224 4224,4099 4224,4093 4224,-128 4224)[OUTER]
```

 If all of the features in the tile are squares of the full extent of the tile, apply Douglas-Peucker with a tolerance of 0 to get a polygon with 5 vertices, then ensure that the start/end vertices are the smallest values.

**Additional context**
Maybe it already works this way and I have a bug in my implementation of this, but you can check the tiles `9/254/85` and `9/255/85` from a build on https://maps.protomaps.com/builds/:

```
pmtiles tile https://demo-bucket.protomaps.com/v4.pmtiles 9 254 85 > 9_254_85.pbf.gz
```

and then look at it with [vtzero](https://github.com/mapbox/vtzero)

msbarry:
Fix #1170 by merging normalized fill polygons

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
@@ -479,19 +479,23 @@ public class FeatureMerge {
    */
   private static void extractPolygons(Geometry geom, List<Polygon> result, double minArea, double minHoleArea) {
     if (geom instanceof Polygon poly) {
-      if (Area.ofRing(poly.getExteriorRing().getCoordinateSequence()) > minArea) {
+      double outerArea = Area.ofRing(poly.getExteriorRing().getCoordinateSequence());
+      if (outerArea > minArea) {
         int innerRings = poly.getNumInteriorRing();
-        if (minHoleArea > 0 && innerRings > 0) {
-          List<LinearRing> rings = new ArrayList<>(innerRings);
-          for (int i = 0; i < innerRings; i++) {
-            LinearRing innerRing = poly.getInteriorRingN(i);
-            if (Area.ofRing(innerRing.getCoordinateSequence()) > minArea) {
-              rings.add(innerRing);
-            }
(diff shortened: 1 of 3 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
```diff
@@ -502,6 +506,41 @@ private static void extractPolygons(Geometry geom, List<Polygon> result, double
...
+    }
+    double proposedBuffer = (Math.sqrt(outerArea) - 256) / 2;
+    double tol = proposedBuffer / 10;
+    double min = -proposedBuffer;
+    double max = 256 + proposedBuffer;
+    var cs = exteriorRing.getCoordinateSequence();
+    for (int i = 0; i < cs.size(); i++) {
+      double x = cs.getX(i);
+      double y = cs.getY(i);
+      if ((notClose(x, min, tol) && notClose(x, max, tol)) && (notClose(y, min, tol) && notClose(y, max, tol))) {
```
msbarry, line 521: this just checks that all vertices are on the fill edge... I _think_ that covers all possible cases here, but might want something more robust like this: https://github.com/onthegomap/planetiler/blob/de4f99c4a3156389cead6016d17c2ef12ebc80b9/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java#L854 that checks whether any segment crosses the inside?

bdon:
Working well, LGTM

PR #1176, opened 2025-02-08
Add Natural Earth support in custom maps

zstadler:


msbarry:
We initially decided not to do this since natural earth distributes in geopackage format now so we could eventually get rid of the 1-off natural earth sqlite format reader.  Do you think there's a use-case for that format that we're missing though?

zstadler:
I realized that from the [sonarqubecloud](https://github.com/apps/sonarqubecloud) message.

I was looking for an example of using Natural Earth's GeoPackage, and could not find any in both planetiler and planetiler-openmaptiles repositories.

In fact, two years after announcing the deprecation, planetiler-openmaptiles still [uses](https://github.com/openmaptiles/planetiler-openmaptiles/blob/309af63b6cbb3ceb4f5c062a781e47764a936332/src/main/java/org/openmaptiles/OpenMapTilesMain.java#L48-L52) the deprecated natural earth sqlite format reader.

PR #1179, opened 2025-02-10
YAML example of using Natural Earth GeoPackage

zstadler:


planetiler-custommap/src/main/resources/samples/natural_earth.yml
```diff
@@ -0,0 +1,67 @@
...
+- id: ne_water_bodies
+  features:
+  - source: natural_earth
+    include_when:
+      - '${ feature.source_layer == "ne_10m_ocean" }'
+      - '${ feature.source_layer == "ne_10m_lakes" }'
+      - '${ feature.source_layer == "ne_50m_ocean" }'
+      - '${ feature.source_layer == "ne_50m_lakes" }'
+      - '${ feature.source_layer == "ne_110m_ocean" }'
+      - '${ feature.source_layer == "ne_110m_lakes" }'
```
msbarry, line 24: Minor: I think with #1063 you should now also be able to do this here and for class attribute:

```yaml
include_when:
${ feature.source_layer }:
- "ne_10m_ocean"
- "ne_10m_lakes"
- "ne_50m_ocean"
- "ne_50m_lakes"
- "ne_110m_ocean"
- "ne_110m_lakes"
```

zstadler: Thanks!
That seems to be applicable also to `min_zoom` and `max_zoom`

zstadler:
Thanks!

Issue #1184, opened 2025-02-17
[FEATURE] Option to include selected tags other than those permitted by com.onthegomap.planetiler.util.LanguageUtils.VALID_NAME_TAGS?


ejschoen:
**Is your feature request related to a problem? Please describe.**
I really hate to bring this up...  We unfortunately have a requirement to deal with tags from OSM for node labels that are other than the name:* tags.  In particular, official_name:* and nat_name:*.  I don't see an option to specify additional tags.

**Describe the solution you'd like**
Command line option to specify a list of tags to include in the mbtiles output beyond those implied by the --languages option.

**Describe alternatives you've considered**
We're using MapLibre GL JS to render mbtiles, and we're using standard styles for it (OSM Bright, etc).  We've got some logic that can load a style and then substitute more flexible rendering for feature names beyond the "text-field": "{name}" that comes standard with these styles.  However, we need the tags for certain features beyond translation and transliteration tags.

**Additional context**

msbarry:
Hello! I think your best bet would be to fork [planetiler-openmaptiles](https://github.com/openmaptiles/planetiler-openmaptiles) to get full customization over what name tags it passes through, or open up a discussion in [openmaptiles](https://github.com/openmaptiles/openmaptiles) repo if you think there's a solution that would be useful to all consumers and it will eventually get ported over to planetiler-openmaptiles.

That being said, it could make sense to change `LanguageUtils.VALID_NAME_TAGS` regex to just verify a valid language tag suffix, but make `Translations` take a set of prefixes that it passes through.  There are a few issues with that regex that need to be fixed anyway (see #1162). That would let you run with `--languages=en,en_us,de,... --name_tag_prefixes=name,official_name,...` which would pass through a _lot_ of extra tags. Alternatively we could add an explicit argument `--extra-name-tags=official_name:en-us,short_name:de` to allow passing through any name tag from input to output?   Minh - any thoughts on either of these options?

ejschoen:
[...]

PR #1187, opened 2025-02-21
Extra name tags

ejschoen:
Addresses Planetiler issue https://github.com/onthegomap/planetiler/issues/1184.  Adds `extraNameTags` to `PlanetilerConfig` record, so that `org.openmaptiles.OpenMapTilesProffile  and`org.openmaptiles.Generate` can call `OmtLanguageUtils.setExtraNameTags`.  See companion PR for those and for changes to OmtLanguageUtils.


msbarry (approved):
Thanks! This looks good.

PR #1190, opened 2025-02-24
Extra name tags

ejschoen:
Logic for copying extra name tags from OSM to output file is now in `org.onthegomap.planetiler.util.Translations.addTranslations`.  I've removed the logger and logging code as suggested. 
 There are no required changed to openmaptiles-planetiler, and the corresponding PR for that repo is now closed.


ejschoen:
Apologies. I think I got a new PR rather than appending onto the old one.  #1190 contains the work to move the extra name tags logic from OmtTranslationUtils to org.onthegomap.planetiler.util.Translations.

ejschoen:
Looking at the timing, maybe you caught my changes in my repo before I made the new commit request.  I didn't see the email I got from Github while I was sleeping...

msbarry:
Closing as it looks like this was fixed in version 0.9.0.

PR #1207, opened 2025-03-19
fix: add the poi layer to the shortbread example

CommanderStorm:
First of all, thanks for the work you are doing. The yaml format works really well ❤️ 

I noticed that the shortbread example for how to create a custom schema does not contain the poi layer.
This PR adds it.

Docs for this are avaliable here:
https://shortbread-tiles.org/schema/1.0/#layer-pois

I have tried to test every distinct failour case that I could come up with (while trying to keep said testcases minimal => not cause undue burden on review)

image

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.yml
@@ -827,27 +827,42 @@ layers:
         amenity: place_of_worship
     - key: recycling:glass_bottles
       type: boolean
-      value: '${ feature.tags["recycling:glass_bottles"] == "yes" }'
+      value:
+        - if: { "recycling:glass_bottles": yes }
+          value: true
+        - else: false
       include_when:
         amenity: recycling
     - key: recycling:paper
       type: boolean
-      value: '${ feature.tags["recycling:paper"] == "yes" }'
+      value:
+        - if: { "recycling:paper": yes }
+          value: true
+        - else: false
       include_when:
         amenity: recycling
     - key: recycling:clothes
       type: boolean
-      value: '${ feature.tags["recycling:clothes"] == "yes" }'
+      value:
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -610,6 +610,252 @@ layers:
...
+        man_made: tower
+    - key: religion
+      include_when:
+        amenity: place_of_worship
+    - key: denomination
+      include_when:
+        amenity: place_of_worship
+    - key: recycling:glass_bottles
+      type: boolean
+      value: '${ feature.tags["recycling:glass_bottles"] == "yes" }'
```
CommanderStorm, line 830: I think this is the best way to express this relationship.
Unsure if there are better ways..

```py
if amenity=recycling:
   value = recycling:glass_bottles == yes
```

msbarry: Alternatively I think you can do:

```yaml
value:
  - if: { amenity: recycling, "recycling:glass_bottles": yes }
    value: true
```

dropping down to the CEL expressions is a slight performance hit compared to yaml-based expressions - although it's probably very minor when you limit evaluation with the include_when expression

msbarry (commented):
Looks great! One minor suggestion on the recycling tags. Thanks for helping with these improvements!

planetiler-custommap/src/main/resources/samples/shortbread.yml
CommanderStorm: good point, did not think of combinding this in this way.
=> https://github.com/onthegomap/planetiler/pull/1207#discussion_r2003342519

We cannot remove the `include_when`, as above would only be
```py
if amenity == recycling and recycling:glass_bottles == yes
    return True
None

PR #1208, opened 2025-03-19
fix: add bridges layer to shortbread example

CommanderStorm:
Similar to #1207, this adds shortbreads bridges to the example layer

Docs on this layer:
https://shortbread-tiles.org/schema/1.0/#layer-bridges

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
@@ -558,6 +558,19 @@ examples:
       name: 'Name'
       ref: 'ref'
 
+- name: 'bridge as polygon'
+  input:
+    source: osm
+    geometry: polygon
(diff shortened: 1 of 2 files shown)


msbarry (approved):

hyperknot:
Hi @CommanderStorm you seem to be an expert on Shortbread in Planetiler, can you join our discussion here?
https://github.com/versatiles-org/versatiles-generator/issues/10

CommanderStorm:
I don't think I can add anything to said discussion.

In therms of full compatability, there is a bit in terms of `min_size`, but nothing blocking.
All layers are there, all major featurs too. Feel free to review the code (or write testcases), it is fairly readable and nice to work with tbh.

I only noticed these things as I was migrating to shortbread and this cleanup were the things that I had to fix (in an later optimised form) in my implementation.

PR #1214, opened 2025-03-23
fix: add way_area to water pologons/labels

CommanderStorm:
I have unexpected a lot of time during a train ride. Expect a few PRs verifying the spec

This adds the `water_polygons*` `way_area` property.

For this, I have not found a way to test. The trestcases also don't fail 🤷🏻
Advice appreciated

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
@@ -39,7 +39,7 @@ examples:
 - name: waterway=dock
   input:
     source: osm
-    geometry: POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))
+    geometry: POLYGON ((-0.5 -0.5, 0.5 -0.5, 0.5 0.5, -0.5 0.5, -0.5 -0.5))
     tags:
       waterway: dock
   output:
@@ -48,7 +48,11 @@ examples:
     min_zoom: 10
     tags:
       kind: dock
-      way_area: 100
+      # verified with a naive "projection" and sizes from https://www.thoughtco.com/degree-of-latitude-and-longitude-distance-4070616
+      # 1° lat ~ 110.567km at the equator
+      # 1° lon ~ 111.321km at the equator
+      # => ~12308.429km² = ~12308.429e^6 m² = ~1.2308429e^10 m²
+      way_area: 1.236418893560872E10
 
 - name: waterway=canal linestring
   input:
--- a/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
@@ -39,7 +39,7 @@ examples:
 - name: waterway=dock
   input:
     source: osm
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -63,6 +63,8 @@ layers:
     attributes:
     - key: kind
       type: match_value
+    - key: way_area # area in square meter (Mercator Projection)
+      value: "${ feature.area('km2') * 1000000.0 }"
```
CommanderStorm, line 67: The docs only mention km2. If m2 is also avalible, the following would maybe be more performant.

Due to this not being testable, I am going to default to the save way ^^

```suggestion
      value: "${ feature.area('m2') }"
```

CommanderStorm: > (see [tests](https://github.com/onthegomap/planetiler/blob/9a0aa4fa6eea54cc85db3be8356b6cb42e3f7e4d/planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredFeatureTest.java#L1595-L1611))

They don't cover `m2`, that is why I was asking.
Will try the input geometry in the evening, thanks ^^

msbarry: Ah good call, I double checked and this does work, added a test to verify in #1226

planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
```diff
@@ -39,7 +39,7 @@ examples:
 - name: waterway=dock
   input:
     source: osm
-    geometry: polygon
+    geometry: POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))
```
CommanderStorm, line 42: Accoridng to my logic, this would be a 100m2 area.
Planetiler somewhat disagrees, no idea why

image
image

CommanderStorm: Also unsure in which coordinate system the input is, it is likely not this one
image

msbarry: the WKT geometries should be using lon, lat coordinates. 1.23e12 square meters is about 0.2% of earth's surface which sounds about right for 10x10 degrees at the equator.

planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -80,6 +82,8 @@ layers:
     - *name
     - *name_en
     - *name_de
+    - key: way_area # area in square meter (Mercator Projection)
+      value: "${ feature.area('km2') * 1000000.0 }"
```
msbarry, line 86: switch to `"${ feature.area('m2') }"` too?

CommanderStorm:
```suggestion
      value: "${ feature.area('m2') }"
```

PR #1215, opened 2025-03-23
[breaking] fix: typo in `water_lines_labels` for shortbread tiles spec

CommanderStorm:
there is a typo in this layer id.
Unsure if this a breaking change, your call

https://shortbread-tiles.org/schema/1.0/#layer-water_lines_labels


msbarry:
Looks good! I added a `[breaking]` prefix as a reminder to call this out in the release notes

PR #1216, opened 2025-03-23
fix: implemented dams and piers in shortbread spec

CommanderStorm:
in the current version of the shortbread spec `{dam,pier}_{polygons,lines}` is present.
This PR implements this.

Docs: https://shortbread-tiles.org/schema/1.0/#layer-dam_lines

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.yml
@@ -174,12 +174,12 @@ layers:
   - source: osm
     geometry: line
     min_zoom: 12
-    include_when:
+    include_when: &pier_filter
       man_made: 
         - pier
         - breakwater
         - groyne
-    attributes:
+    attributes: &pier_attrs
     - key: kind
       type: match_value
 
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -149,6 +149,54 @@ layers:
...
+    geometry: polygon
+    min_zoom: 12
+    include_when:
+      man_made: 
+        - pier
+        - breakwater
+        - groyne
+    attributes:
+    - key: kind
+      type: match_value
```
CommanderStorm, line 198: not done as below because of the 50 alias limit:
image


```suggestion
    include_when: &pier_filter
      man_made: 
        - pier
        - breakwater
        - groyne
    attributes: &pier_attrs
    - key: kind
      type: match_value

- id: pier_polygons
  features:
  - source: osm
    geometry: polygon
    min_zoom: 12
    include_when: *pier_filter
    attributes: *pier_attrs
```

CommanderStorm: Yes, if https://github.com/onthegomap/planetiler/pull/1224 is merged, this (and other similar comments) can be committed without breaking the verification ^^

PR #1218, opened 2025-03-23
fix: add missing tags for sites and land in planetiler spec

CommanderStorm:
sites and land are mostly correct.
There a few items which likely have been added afterwards => this PR adds them to planetiler as well

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.yml
@@ -284,6 +284,7 @@ layers:
       - farmland
       - farmyard
       - forest
+      - garages
       - grass
       - greenfield
       - greenhouse_horticulture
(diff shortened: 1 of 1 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -360,11 +361,12 @@ layers:
...
       amenity:
-      - university
+      - bicycle_parking
+      - college
       - hospital
-      - prison
       - parking
-      - bicycle_parking
+      - prison
+      - university
```
CommanderStorm, line 369: this is sorting and adding `college`

planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -344,7 +346,6 @@ layers:
           - sand
         13:
           amenity: grave_yard
-          natural: wood
```
CommanderStorm, line 347: Would be interesting to throw an error for this.
I don't think overrides should be able to overlap.

PR #1220, opened 2025-03-23
fix layer-order difference for shortbread

CommanderStorm:
I don't know how important this is, but the sorting is slightly different for our implementation vs the shortbread website


msbarry:
Looks good, definitely makes sense to match the spec.  We could also break each layer out into a separate file if that were easier to manage?

CommanderStorm:
> We could also break each layer out into a separate file

I don't think you should do that. Currently, this is in one piece and as a config file, splitting is always akward, considering you then need to resolve paths..

PR #1221, opened 2025-03-23
[breaking] fix: streets being differently slightly different to shortbread

CommanderStorm:
The todo below is a bit insane, but that is the spec 🤷🏻

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.yml
@@ -100,15 +100,17 @@ layers:
     attributes:
     - key: kind
       type: match_value
-    - &water_lines_attr_tunnel
+    - &attr_tunnel
       key: tunnel
+      min_zoom: 11 # to reuse for streets, not relevant for water as min_zoom is lower
       value: true
       include_when:
         tunnel: [ yes, building_passage ]
         covered: yes
       else: false
-    - &water_lines_attr_bridge
+    - &attr_bridge
       key: bridge
+      min_zoom: 11 # to reuse for streets, not relevant for water as min_zoom is lower
       value: true
       include_when:
         bridge:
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -571,43 +615,59 @@ layers:
...
+        11:
+          area:aeroway: runway
+        13:
+          area:aeroway: taxiway
     include_when:
-      __all__:
-      - highway: [ pedestrian, service ]
-      - area: yes
+      # TODO this is a really strange difference to street_polygons.
+      # https://github.com/shortbread-tiles/shortbread-docs/issues/80
```
msbarry, line 638: interesting... what does the tilemaker implementation do?

CommanderStorm: they don't care about `aeroway` and only consider `area:aeroway`

image

CommanderStorm: schould I change it as such?

msbarry: Seems like a mistake... I see plenty of runway polygons tagged with `area=yes` `aeroway=runway` (not `area:aeroway=runway` ([for example](https://www.openstreetmap.org/way/997816571)).  Think it would make sense to support both in both layers?

CommanderStorm: Not sure.


@pnorman do you have an opinion how plane tiler should handle this?

pnorman: If there's an question with the shortbread spec, it should be raised there.

msbarry:
Sorry for the delay! This looks good for now, we can revisit the runway tagging once we get more clarity on the shortbread repo issue.

PR #1224, opened 2025-03-23
feat: remove alias limit for custom maps

CommanderStorm:
Currently, this alias limit is 50.
For large custom maps, like shortbread, this is a bit to few and pretty annoying.

The opton configured does:
> Restrict the number of aliases for collection nodes to prevent Billion laughs attack.

https://javadoc.io/static/org.snakeyaml/snakeyaml-engine/2.3/org/snakeyaml/engine/v2/api/LoadSettingsBuilder.html#setMaxAliasesForCollections(int)

I don't think that this attack vector is a valid one for this tool.


msbarry:
Looks good! Thanks!

PR #1232, opened 2025-03-31
test(shortbread): apply `min_zoom` and `allow_extra_tags: false` to the shortbread spec

CommanderStorm:
during https://github.com/onthegomap/planetiler/pull/1221, I noticed that we could ommit testing some things by accident.

Currently these options are not set, but I think this helps finding bugs => lets enable it

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml
@@ -998,6 +998,7 @@ examples:
     layer: boundary_labels
     geometry: point
     min_zoom: 5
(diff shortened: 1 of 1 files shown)


CommanderStorm:
Cannot reproduce the CI-fail locally..
image

the schema also does never set this. Unsure where it is coming from.

msbarry:
Makes sense, what if we set the default value to be allow_extra_tags: false in the code and you can override with allow_extra_tags: true if you want partial match?

CommanderStorm:
Changing said default would make sense, but would be quite a breaking change for people who test their styles..

Would be better though imo..

PR #1244, opened 2025-04-22
Fix geopackage transform

msbarry:
According to https://www.geopackage.org/spec/:

> The axis order in WKB stored in a GeoPackage follows the de facto standard for axis order in WKB and is therefore always (x,y{,z}{,m}) where x is easting or longitude, y is northing or latitude, z is optional elevation, and m is optional measure. This ordering explicitly overrides the axis order as specified in the SRS metadata, applying Case 4 from OGC 08-038r7, Revision to Axis Order Policy and Recommendations[[K11]](https://www.geopackage.org/spec/#K11). This was done to maintain consistency with previous implementations of WKB that predated the OGC policy.

So this PR updates geopackage reader to force axis order to be longitude first, even if the CRS embedded in it says it's latitude first (which many do)


bdon:
This fixes the reading of the water polygons in geographic (EPSG:4326) GPKG created by GDAL.

bdon:
(Actually, there seems to be a bug related to EPSG:4326, likely something I introduced, where certain latitudes do not appear at z0. will investigate...)

msbarry:
This does the coordinate swap by swapping the x/y coordinates at read-time so it is possible it is not handling high latitudes/longitudes correctly ... I'll add another test but let me know what you find out

msbarry:
The new tests seem to pass so I think it should handle the full lat/lon extents...

bdon:
trying to track down why all the geometries from -60 to +60 latitude are missing:

CleanShot 2025-04-26 at 23 17 04@2x

```java
package com.onthegomap.planetiler.examples;

import com.onthegomap.planetiler.FeatureCollector;
import com.onthegomap.planetiler.Planetiler;
import com.onthegomap.planetiler.Profile;
import com.onthegomap.planetiler.config.Arguments;
import com.onthegomap.planetiler.reader.SourceFeature;
import java.nio.file.Path;

public class GpkgDemo implements Profile {

  @Override
  public void processFeature(SourceFeature sourceFeature, FeatureCollector features) {
    if (sourceFeature.canBePolygon()) {
      features.polygon("water")
        .setZoomRange(0, 10);
    }
  }

  @Override
  public String name() {
    return "Geopackage Demo";
  }

  @Override
  public String description() {
    return "An example";
  }

  @Override
  public boolean isOverlay() {
    return true;
  }

  @Override
  public String attribution() {
    return """
      <a href="https://www.openstreetmap.org/copyright" target="_blank">&copy; OpenStreetMap contributors</a>
      """.trim();
  }

  public static void main(String[] args) throws Exception {
    run(Arguments.fromArgsOrConfigFile(args));
  }

  static void run(Arguments args) throws Exception {
    Planetiler.create(args)
      .setProfile(new GpkgDemo())
      .addGeoPackageSource("water", Path.of("data", "sources", "water-polygons-split-4326.gpkg"), "http://example.com")
      .overwriteOutput(Path.of("data", "gpkg.pmtiles"))
      .run();
  }
}
```
input: https://r2-public.protomaps.com/datasets/water-polygons-split-4326.gpkg
output: https://pmtiles.io/#url=https%3A%2F%2Fr2-public.protomaps.com%2Ftmp%2Fgpkg.pmtiles&map=0.42/0/0

bdon:
Nevermind, this isn't a bug. The unprojected geometries just happen to fail the default min pixel size threshold at those latitudes. adding `.setMinPixelSize(0);` makes it appear correct:
CleanShot 2025-04-27 at 10 33 27@2x

msbarry:
Nice! No issues besides that? If so I can get this merged.

bdon:
LGTM

Issue #1273, opened 2025-06-24
Question about `mergeLineStrings`


zstadler:
I'm trying to exclude waterways shorter than 30px in zoom levels lower than `maxzoom`. I thought it will be easy:

```java
  @Override
  public List<VectorTile.Feature> postProcessLayerFeatures(String layer, int zoom,
    List<VectorTile.Feature> items) {
    if (zoom == config.maxzoom()) {
      return items;
    }
    // FeatureMerge has several utilities for merging geometries in a layer that share the same tags.
    // `mergeLineStrings` combines lines with the same tags where the endpoints touch.
    // Tiles are 256x256 pixels and all FeatureMerge operations work in tile pixel coordinates.
    return FeatureMerge.mergeLineStrings(items,
      30,	// after merging, remove lines that are still less than 30px long
      0.25,	// simplify output linestrings using a 0.2px tolerance
      -1 	// remove any detail outside the tile boundary
    );
  ```

However, it seems like the linestrings are cut to the tile boundaries **before** filtering-out short lines.

![Image](https://github.com/user-attachments/assets/5efbbac8-9506-4d4e-9c91-ec617f8a529d)

The line segments that are removed in zoom=13 are part of the same OSM way that was kept on the other side of the tile boundary.
 
Is there a way to:
1. Merge all linestrings that exist in the tile, according to their properties
2. Optionally simplify, according to `tolerance`
3. Filter-out short linestrings , according to `minLength`
4. Optionally remove details at a certain distance from the tile boundaries, according to `buffer`

zstadler:
To reproduce:
1. Download [`waterway.java.txt`](https://github.com/user-attachments/files/20887896/waterway.java.txt)
2. Rename it to `waterways.java`
3. Run it with Planetiler or using docker
   ```
   docker run --rm -v .:/w --workdir /w --entrypoint java ghcr.io/onthegomap/planetiler -cp @/app/jib-classpath-file waterway.java --download
   ```
3. Open [PMTiles viewer](https://pmtiles.io/#map=13.99/29.99363/35.06786&showTileBoundaries=true) in a browser
4. Browse and open `data/waterway.pmtiles`

msbarry:
You also need to call `.setBufferPixels(30)` in `process()` method. And in postProcessLayerFeatures, the last argument to mergeLineStrings should probably be 4 or something so that the extra detail outside the tile is removed after merging.

[...]

PR #1277, opened 2025-06-28
Use `min_length` to `setBufferPixels()`

zstadler:
Integrate the [solution proposed there](https://github.com/onthegomap/planetiler/issues/1273#issuecomment-3004486389) into the custom map flow.

Resolve https://github.com/zstadler/planetiler/issues/2
Resolve https://github.com/onthegomap/planetiler/issues/1273

Related to https://github.com/zstadler/planetiler/issues/1

### DIFF
--- a/planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
+++ b/planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
@@ -21,6 +21,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A map feature, configured from a YML configuration file.
@@ -29,6 +31,8 @@ import java.util.function.Function;
  * and {@link #processFeature(Contexts.FeaturePostMatch, FeatureCollector)} processes matching elements.
  */
 public class ConfiguredFeature {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ConfiguredFeature.class);
+
   private static final double LOG4 = Math.log(4);
   private final Expression geometryTest;
   private final Function<FeatureCollector, Feature> geometryFactory;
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/ConfiguredFeature.java
```diff
@@ -91,6 +92,15 @@ public ConfiguredFeature(String layer, TagValueProducer tagValueProducer, Featur
     processors.add(makeFeatureProcessor(feature.minZoom(), Integer.class, Feature::setMinZoom));
     processors.add(makeFeatureProcessor(feature.maxZoom(), Integer.class, Feature::setMaxZoom));
     processors.add(makeFeatureProcessor(feature.minSize(), Double.class, Feature::setMinPixelSize));
+    if (layer.postProcess() != null) {
+      var merge = layer.postProcess().mergeLineStrings();
+      if (merge != null) {
+        var minLength = merge.minLength();
+          if (minLength > 4) {
+            processors.add(makeFeatureProcessor(minLength, Double.class, Feature::setBufferPixels));
```
msbarry, line 110: What do you think about setting min length to 0 too when mergeLineStrings is present?

zstadler: I was working on that as a separate PR. I'll add it here.

zstadler: I hope avoiding the execution of
```
processors.add(makeFeatureProcessor(feature.minSize(), Double.class, Feature::setMinPixelSize));
```
is good enough

msbarry: There's a default min size so you'd need to explicitly set it to 0. Feel free to do in a separate pr if you want

zstadler: That brings the topic of allowing custom map users to set different values for `*_at_max_zoom`.
This is also related to differentiating between having `tolerance: 0`, and omitting it. IMO, when there is no `tolerance` at the layer level, the `simplify_tolerance` and `simplify_tolerance_at_max_zoom` should be used.

I suggest that min size will be set to zero in this PR. Perhaps a separate issue would be appropriate for discussing `*_at_max_zoom` before starting the implementation.

msbarry (approved):

PR #1339, opened 2025-09-22
planetiler-openmaptiles adjusted and updated + excluded tests now re-…

phanecak-maptiler:
Follow-up on https://github.com/openmaptiles/planetiler-openmaptiles/pull/264 , e.g:

1. `planetiler-openmaptiles` adjusted and updated
2.  Excluded tests now re-enabled


msbarry:
Thanks!!

PR #1341, opened 2025-09-28
Improve handling of super relation

### LINKED ISSUE #857: [FEATURE] OSM multiline relation handling
Is there a general-purpose way to handle relations that define long connected lines (ie. [route](https://wiki.openstreetmap.org/wiki/Relation:route)/[waterway](https://wiki.openstreetmap.org/wiki/Relation:waterway)) in a similar way to how planetiler handles [multipolygon](https://wiki.openstreetmap.org/wiki/Relation:multipolygon) relations today?

Polygons/multipolygons are currently handled transparently to profiles whether they came from a single way or a complex [multipolygon relation](https://wiki.openstreetmap.org/wiki/Relation:multipolygon). Relations are also used to group ways into long linestrings/multilinestrings, for example: [hiking/biking/bus routes](https://wiki.openstreetmap.org/wiki/Relation:route), [waterways](https://wiki.openstreetmap.org/wiki/Relation:waterway), [country boundaries](https://wiki.openstreetmap.org/wiki/Relation:boundary), [public transit](https://wiki.openstreetmap.org/wiki/Relation:public_transport). Today, profiles need to process each way in the relation and emit a line, then merge connected lines when post-processing finished tiles. This doesn't let you process an entire connected linestring as a single geometry, for example to compute its overall length. It's also not supported by yaml schemas.

For example to process every national cycling route, a profile could do:

```yaml
include_when:
  - network: ncn
geometry: linestring
```

Is something like that sufficient to cover most cases? If not how much more flexibility does it need? There are also [super-relations](https://wiki.openstreetmap.org/wiki/Super-relation) (boundary and routes), cases where a way and relation it's contained in both match, and relation member roles that may convey importance. 

The goal would be to make it possible to reconstruct multilines in a java profile (probably by extending multipolygon reconstruction logic) and expose a simple API for it through yaml configs.

HarelM:
- Resolves #857 (?)

This does the following:
It stores a map between a relation to its parent relation in pass1.
When passing on ways in pass2 it traverses the tree up to get all the relationinfo stored for each way.
It makes sure not to hang on circular relations (I've encountered those unfortunately in the past).

I believe that relation hierarchy is not too deep so this should not become a performance bottleneck.
Let me know what you think.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
@@ -85,6 +85,10 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
   private LongObjectHashMap<OsmRelationInfo> relationInfo = Hppc.newLongObjectHashMap();
   // ~800mb, ~1.6GB when sorting
   private LongLongMultimap.Appendable wayToRelations = LongLongMultimap.newAppendableMultimap();
+
+  // HM TODO: size info is probably needed
+  private LongLongMultimap.Appendable relationToParentRelations = LongLongMultimap.newAppendableMultimap();
+
   private final Object wayToRelationsLock = new Object();
   // for multipolygons need to store way info (20m ways, 800m nodes) to use when processing relations (4.5m)
   // ~300mb
@@ -265,12 +269,14 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
                   for (OsmRelationInfo info : infos) {
                     relationInfo.put(relation.id(), info);
                     relationInfoSizes.addAndGet(info.estimateMemoryUsageBytes());
-                    for (var member : relation.members()) {
-                      var type = member.type();
-                      // TODO handle nodes in relations and super-relations
-                      if (type == OsmElement.Type.WAY) {
-                        wayToRelations.put(member.ref(), encodeRelationMembership(member.role(), relation.id()));
-                      }
+                  }
+                  for (var member : relation.members()) {
+                    var type = member.type();
+                    // TODO handle nodes in relations
+                    if (type == OsmElement.Type.WAY) {
+                      wayToRelations.put(member.ref(), encodeRelationMembership(member.role(), relation.id()));
+                    } else if (type == OsmElement.Type.RELATION) {
+                      relationToParentRelations.put(member.ref(), relation.id());
                     }
                   }
                 }
@@ -537,7 +543,28 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
         if (rel != null) {
           rels.add(new RelationMember<>(parsed.role, rel));
         }
+        rels.addAll(getRelationInfosForRelationId(parsed.relationId, new HashSet<>()));
       }
+    }
+    return rels;
+  }
+
+  private List<RelationMember<OsmRelationInfo>> getRelationInfosForRelationId(long relationId, HashSet<Long> seen) {
+    if (!seen.add(relationId)) {
+      return List.of();
+    }
+    LongArrayList parentRelations = relationToParentRelations.get(relationId);
+    if (parentRelations.isEmpty()) {
+      return List.of();
+    }
+    List<RelationMember<OsmRelationInfo>> rels = new ArrayList<>(parentRelations.size());
+    for (int p = 0; p < parentRelations.size(); p++) {
+      long parentId = parentRelations.get(p);
+      OsmRelationInfo parentRelation = relationInfo.get(parentId);
+      if (parentRelation != null) {
+        rels.add(new RelationMember<>("", parentRelation));
+      }
+      rels.addAll(getRelationInfosForRelationId(parentId, seen));
     }
     return rels;
   }
--- a/planetiler-core/src/test/java/com/onthegomap/planetiler/reader/osm/OsmReaderTest.java
+++ b/planetiler-core/src/test/java/com/onthegomap/planetiler/reader/osm/OsmReaderTest.java
@@ -703,6 +703,92 @@ class OsmReaderTest {
     assertEquals(List.of(), feature.relationInfo(TestRelInfo.class));
   }
 
+  @Test
+  void testSuperRelationOrderedEasily() {
+    record TestRelInfo(long id, String name) implements OsmRelationInfo {}
+    OsmReader reader = new OsmReader("osm", () -> osmSource, nodeMap, multipolygons, new Profile.NullProfile() {
+      @Override
+      public List<OsmRelationInfo> preprocessOsmRelation(OsmElement.Relation relation) {
+        return List.of(new TestRelInfo(1, "name"));
+      }
+    }, stats);
+    var nodeCache = reader.newNodeLocationProvider();
+    long index = 1;
+    var node1 = node(index++, 0, 0);
+    var node2 = node(index++, 1, 1);
+    var node3 = node(index++, 2, 2);
+    var node4 = node(index++, 3, 3);
+    var way1 = new OsmElement.Way(index++);
+    way1.nodes().add(node1.id(), node2.id());
+    way1.setTag("key", "value");
+    var way2 = new OsmElement.Way(index++);
+    way2.nodes().add(node3.id(), node4.id());
+    way2.setTag("key", "value");
+    var relation1 = new OsmElement.Relation(index++);
+    var relation2 = new OsmElement.Relation(index++);
+    var relation3 = new OsmElement.Relation(index++);
+    relation1.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way1.id(), "rolename"));
+
+    relation2.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation1.id(), "rolename"));
+    relation2.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way2.id(), "rolename"));
+
+    relation3.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation1.id(), "rolename"));
+
+    processPass1Block(reader, List.of(node1, node2, node3, node4, way1, way2, relation1, relation2, relation3));
+
+    SourceFeature feature1 = reader.processWayPass2(way1, nodeCache);
+    SourceFeature feature2 = reader.processWayPass2(way2, nodeCache);
+    assertEquals(3, feature1.relationInfo(TestRelInfo.class).size());
+    assertEquals(1, feature2.relationInfo(TestRelInfo.class).size());
+  }
+
+  @Test
+  void testSuperRelationOrderedBadly() {
+    record TestRelInfo(long id, String name) implements OsmRelationInfo {}
+    OsmReader reader = new OsmReader("osm", () -> osmSource, nodeMap, multipolygons, new Profile.NullProfile() {
+      @Override
+      public List<OsmRelationInfo> preprocessOsmRelation(OsmElement.Relation relation) {
+        return List.of(new TestRelInfo(1, "name"));
+      }
+    }, stats);
+    var nodeCache = reader.newNodeLocationProvider();
+    long index = 1;
+    var node1 = node(index++, 0, 0);
+    var node2 = node(index++, 1, 1);
+    var node3 = node(index++, 2, 2);
+    var node4 = node(index++, 3, 3);
+    var node5 = node(index++, 4, 4);
+    var node6 = node(index++, 5, 5);
+    var way1 = new OsmElement.Way(index++);
+    way1.nodes().add(node1.id(), node2.id());
+    way1.setTag("key", "value");
+    var way2 = new OsmElement.Way(index++);
+    way2.nodes().add(node3.id(), node4.id());
+    way2.setTag("key", "value");
+    var way3 = new OsmElement.Way(index++);
+    way3.nodes().add(node5.id(), node6.id());
+    way3.setTag("key", "value");
+    var relation1 = new OsmElement.Relation(index++);
+    var relation2 = new OsmElement.Relation(index++);
+    var relation3 = new OsmElement.Relation(index++);
+    relation1.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation2.id(), "rolename"));
+    relation1.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way1.id(), "rolename"));
+
+    relation2.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation3.id(), "rolename"));
+    relation2.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way2.id(), "rolename"));
+
+    relation3.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way3.id(), "rolename"));
+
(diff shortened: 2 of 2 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -265,12 +269,14 @@ void processPass1Blocks(Iterable<? extends Iterable<? extends OsmElement>> block
                   for (OsmRelationInfo info : infos) {
                     relationInfo.put(relation.id(), info);
                     relationInfoSizes.addAndGet(info.estimateMemoryUsageBytes());
-                    for (var member : relation.members()) {
-                      var type = member.type();
-                      // TODO handle nodes in relations and super-relations
-                      if (type == OsmElement.Type.WAY) {
-                        wayToRelations.put(member.ref(), encodeRelationMembership(member.role(), relation.id()));
-                      }
+                  }
```
HarelM, line 272: I've closed the for loop here as I don't think there's a need to do the code related to ref management for each OsmRelationInfo (In most cases this list is probably a list of one item, but still).

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -85,6 +85,10 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
   private LongObjectHashMap<OsmRelationInfo> relationInfo = Hppc.newLongObjectHashMap();
   // ~800mb, ~1.6GB when sorting
   private LongLongMultimap.Appendable wayToRelations = LongLongMultimap.newAppendableMultimap();
+
+  // HM TODO: size info is probably needed
```
HarelM: I wasn't sure what to write here...

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -537,17 +543,40 @@ private List<RelationMember<OsmRelationInfo>> getRelationMembershipForWay(long w
         if (rel != null) {
           rels.add(new RelationMember<>(parsed.role, rel));
         }
+        LongArrayList parentRelations = relationToParentRelations.get(parsed.relationId);
```
msbarry, line 546: Minor: since this will be empty most of the time, could we skip the rest of the work/making the new hashset if `parentRelations.isEmpty()` ?

And also update `SparseUnorderedBinarySearchMultimap.get` to avoid allocating a new LongArrayList it's going to be empty.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -537,17 +543,40 @@ private List<RelationMember<OsmRelationInfo>> getRelationMembershipForWay(long w
...
+        var hashSet = new HashSet<Long>();
```
msbarry: Minor instead of hashSet let's call it "visited" or something to indicate it prevents cycles.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
msbarry: The only issue I can see is that if `infos` is an empty list it will now store metadata about it. From what I can see most of the time preprocessOsmRelation returns null not empty list but probably safer to make the check above be `if (infos != null && !infos.isEmpty()) {`

But also this means that profiles need to emit a relation info for all intermediate relations for any parent to be accessible. I'm not sure if that's a good or a bad thing 🤔

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -537,17 +543,40 @@ private List<RelationMember<OsmRelationInfo>> getRelationMembershipForWay(long w
...
 
+  private List<RelationMember<OsmRelationInfo>> getRelationInfosForRelationId(long relationId, HashSet<Long> seen) {
+    if (!seen.add(relationId)) {
+      return List.of();
+    }
+    LongArrayList parentRelations = relationToParentRelations.get(relationId);
+    List<RelationMember<OsmRelationInfo>> rels = new ArrayList<>(parentRelations.size());
+    OsmRelationInfo parentRelation = relationInfo.get(relationId);
+    if (parentRelation != null) {
+      rels.add(new RelationMember<>("", parentRelation));
```
msbarry: This means that profiles won't have access to a way's role in a relation, or the relation's role in a super relation. It seems like since `feature.relationInfo` will return all parents then we'd want role populated in each and profiles could figure out which one they'd want to use.  We might also want "parent relation" optional attribute on relation members to be able to make sense out of the results when processing?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
msbarry: Looks like there are ~200k superrelations and ~1m relations in those superrelations, so worst-case this would be a negligible ~20mb or ~40mb while sorting.

https://github.com/onthegomap/planetiler/issues/857#issuecomment-2029626660

msbarry (commented):
Thanks for adding this! Added a few comments thinking through how consumers will use this...

Here are some comments with example superrelations:

- https://github.com/onthegomap/planetiler/issues/857#issuecomment-2105285592 Parisian subway line groupings that define color for the line [3328695](https://www.openstreetmap.org/relation/3328695#map=17/48.86859/2.31292&layers=T)
- https://github.com/onthegomap/planetiler/issues/857#issuecomment-2443679810 Get name from top superrelation:
  - EuroVelo 10 top super relation [OSM](https://www.openstreetmap.org/relation/2689634) [waymarkedtrails](https://cycling.waymarkedtrails.org/#route?id=2689634&type=relation&map=5.0/60.4928/18.8055)
  - EuroVelo 10 part Poland super relation [OSM](https://www.openstreetmap.org/relation/9566115)
  - EuroVelo 10 part Poland 3 [OSM](https://www.openstreetmap.org/relation/9422510#map=10/54.7333/18.0989)
- Also [Israel Bike Trail](https://www.openstreetmap.org/relation/19598237#map=13/31.29521/35.28915)
- And [appalachian trail](https://www.openstreetmap.org/relation/156553)

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
msbarry: Most roles for superrelation members seem to be "" but here's one that does use a "connection" role for a child relation: https://www.openstreetmap.org/relation/2171555.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: Updated in ca5309f5097fffd4d30ea514311abf9b71888523

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: Updated in ca5309f5097fffd4d30ea514311abf9b71888523
Not sure about the second part, but doesn't SparseUnorderedBinarySearchMultimap.get returns an `EMPTY_LIST` so no allocation is being made?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: Fixed in ca5309f5097fffd4d30ea514311abf9b71888523

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: Fixed in ca5309f5097fffd4d30ea514311abf9b71888523
About the intermediate relations, yeah, I'm not sure either what's the right thing to do here...

HarelM:
I think I've addressed most of the comments, let me know if you want me to change anything else.
Thanks for the quick review!

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: How do you suggest solving this? encoding the role like it is done for ways?

HarelM: I've added role support, not sure how to add the parent id though...

HarelM:
In theory, the parent map can run outside the info loop (similar to multipolygon?) to collect the parent information and store it in the map, this will avoid the need to check if an info was provided for a relation.
Let me know which way you'd like to have it, I'm not sure I have strong feelings either way...

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
msbarry: Given the storage required for the parent relations is minimal I think I'm inclined to just store all parent relationships so we always traverse the intermediate relations but we only include the relation info for relations that your profile emitted when processing a way.

HarelM: I've tried to add that code using the test below, the problem with this approach is that you'd need to store all the ways that belong to a relation to start traversing up the tree, and this will increase memory usage of the stored parent map.
I'll see if I can think of a better way so solve this, but I'm not sure it will be easy.
Feel free to use the following test to experiment with the code change on this branch:
```java
@Test
  void testSuperRelationSetInfoOnlyForTopmost() {
    record TestRelInfo(long id, String name) implements OsmRelationInfo {}
    OsmReader reader = new OsmReader("osm", () -> osmSource, nodeMap, multipolygons, new Profile.NullProfile() {
      @Override
      public List<OsmRelationInfo> preprocessOsmRelation(OsmElement.Relation relation) {
        if (relation.id() == 6) {
          return List.of(new TestRelInfo(6, "topmostinfo"));
        }
        return null;
      }
    }, stats);
    var nodeCache = reader.newNodeLocationProvider();
    long index = 1;
    var node1 = node(index++, 0, 0);
    var node2 = node(index++, 1, 1);
    var way1 = new OsmElement.Way(index++);
    way1.nodes().add(node1.id(), node2.id());
    way1.setTag("key", "value");
    var relation1 = new OsmElement.Relation(index++);
    var relation2 = new OsmElement.Relation(index++);
    var relation3 = new OsmElement.Relation(index);
    relation1.members().add(new OsmElement.Relation.Member(OsmElement.Type.WAY, way1.id(), "leafway"));

    relation2.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation1.id(), "midrel"));

    relation3.members().add(new OsmElement.Relation.Member(OsmElement.Type.RELATION, relation2.id(), "topmosrel"));

    processPass1Block(reader, List.of(node1, node2, way1, relation1, relation2, relation3));

    SourceFeature feature1 = reader.processWayPass2(way1, nodeCache);
    assertEquals(1, feature1.relationInfo(TestRelInfo.class).size());
    assertEquals("topmostinfo", feature1.relationInfo(TestRelInfo.class).get(0).relation().name());
  }
```

HarelM: The following code solves this issue, let me know if you want me to check it in or not - basically store all the links between ways to their relations in the `wayToRelations` variable without storing the info, so that a traverse up can start. 
image

msbarry: Hmm good point... relation to parent relation is 20-40mb but way to parent relation is at least 2-4GB which is no longer negligible.  Probably the right way to do this in the future is to store all of the element to parent element relations on disk so they can be looked up if needed but not increase RAM requirements. That a bigger project though. I think it's fine to leave the restriction for now that you need to declare all intermediate relations for now.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
msbarry: I was thinking something like this:

```diff
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongLongMultimap.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongLongMultimap.java
@@ -208,8 +208,10 @@ public interface LongLongMultimap extends MemoryEstimator.HasEstimate, DiskBacke
       }
       int size = keys.size();
       int index = Arrays.binarySearch(keys.buffer, 0, size, key);
-      LongArrayList result = new LongArrayList();
-      if (index >= 0) {
+      if (index < 0) {
+        return EMPTY_LIST;
+      } else {
+        LongArrayList result = new LongArrayList();
         // binary search might drop us in the middle of repeated values, so look forwards...
         for (int i = index; i < size && keys.get(i) == key; i++) {
           result.add(values.get(i));
@@ -218,8 +220,8 @@ public interface LongLongMultimap extends MemoryEstimator.HasEstimate, DiskBacke
         for (int i = index - 1; i >= 0 && keys.get(i) == key; i--) {
           result.add(values.get(i));
         }
+        return result;
       }
-      return result;
     }
```

HarelM: Ohh, I've changed that now.
I think that in most cases the map is simply empty and there's a condition for that one line before your change, but sure, I don't mind the micro optimization.

HarelM:
No worries, I just did.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
HarelM: Commit: d749160

HarelM:
Let me know when a new version is available with this fix as I would like to see if I can support super relation with it and see the ergonomics around it (I probably could've tested it without a release, but I think the code changes here are good anyway).

zstadler:
What changes, if any, are needed in the [BikeRouteOverlay.java](https://github.com/onthegomap/planetiler/blob/main/planetiler-examples/src/main/java/com/onthegomap/planetiler/examples/BikeRouteOverlay.java) example to handle super-relations?

HarelM:
I'm not sure it needs any changes, but it depends on how bike route super relations are tagged in OSM.
If they are tagged with:
`type = route` and `route = mtb | bicycle` they will be picked up in the `preprocessOsmRelation` section of the profile to be later on used in the `processFeature` method.
In this method a new line feature will be created so that later on it will be added to the tiles.
For each relation that a way belongs to there will be a line with the name of the relataion (and super relation) and its `ref`.

Let me know if this answers your question.
If the super relations are tagged differently than "regular" relations then there might be a need to change the profile code as well.

msbarry:
Thanks for making this change @HarelM ! Here's an example bike route super relation: https://www.openstreetmap.org/relation/19598237  - the child and parent relations are both tagged with type=route route=mtb so I think it should pick them all up automatically. Let me merge this though and you can test it out

Issue #1349, opened 2025-10-01
[FEATURE] Maplibre Vector Tile support


msbarry:
The MLT spec is getting close to the initial release. Planetiler should add support for writing MLT format when it's ready.

jatech1:
Can someone point me to what command line param needs to be set to build MLT pmtiles output?  I went through the issue and PR and am probably just missing it... thanks!

millionpoundhat:
`--tile-format=mlt`

jatech1:
Tried it with 0.9.3 and it does not do anything.  Looks like this code was merged after 0.9.3 release commit.  I tried it with the latest snapshot from last week and it seems to build but only about 1% reduction in size.  Is this actually building with mlt format?  Is there any info on the overall status of mlt support and when the 0.9.4 release is planned?  Thanks!

louwers:
MLT Java 0.0.9 release is out: https://github.com/maplibre/maplibre-tile-spec/releases/tag/java-v0.0.9

louwers:
Note to self: update https://maplibre.org/maplibre-tile-spec/implementation-status/ when this is closed.

msbarry:
Just released [v0.10.0](https://github.com/onthegomap/planetiler/releases/tag/v0.10.0) with MLT suport.

PR #1355, opened 2025-10-09
Initial support for maplibre vector tiles

msbarry:
Support writing tiles in the new maplibre vector tile format when you use the `--tile-format=mlt` command-line arg.

TODO:

- [x] naming? MLT/MLTv1?
- [x] better exception handling
- [x] type coercion?
- [x] ensure deterministic output
- [x] verify tiles look correct from maplibre-gl-js draft branch https://github.com/maplibre/maplibre-gl-js/pull/6570

later:
- [ ] reduce impact on planetiler.jar size https://github.com/maplibre/maplibre-tile-spec/issues/592
- [ ] any low hanging fruit on performance

image

### DIFF
--- a/planetiler-core/pom.xml
+++ b/planetiler-core/pom.xml
@@ -192,6 +192,11 @@
       <version>1.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-text</artifactId>
+      <version>1.14.0</version>
+    </dependency>
+    <dependency>
       <groupId>org.maplibre</groupId>
       <artifactId>mlt</artifactId>
       <version>0.0.3</version>


planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
```diff
@@ -124,8 +125,9 @@ public TileArchiveMetadata metadata() {
       case UNKNOWN -> TileCompression.UNKNOWN;
     };
 
-    String format = switch (header.tileType()) {
-      case MVT -> TileArchiveMetadata.MVT_FORMAT;
+    TileFormat format = switch (header.tileType()) {
+      case MVT -> TileFormat.MVT;
+      case MLT -> TileFormat.MLT;
```
nyurik, line 130: you probably want to use something like `MLTx01` or similar - because we plan to have many more :)

msbarry: Please elaborate?

msbarry:
Resolves #1349

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
TimSylvester: For the most part, those "tags" can be treated as version information and they're for the individual tables/layers, not for the file as a whole.

So I'd suggest assigning a value for MLT as a tile type, and using the JSON if any additional information is needed before getting to actually parsing tiles, but we should try to avoid that.

nyurik: sounds good, if these are tile-wide, than yes, MLT is enough. I for some reason thought this was more of a per-layer setting.

msbarry: Is the plan for decoders to handle all future column types, ie. will https://github.com/maplibre/maplibre-tile-spec/blob/main/spec/schema/mlt_tileset_metadata.proto just grow as you add more capabilities or will you make a completely new proto definition for each tag version?  If it's the former then it seems like as new capabilities are added and we figure out a way for users to opt into using them through planetiler that we could just make use of those new features without choosing an mlt version?

msbarry:
I generated a planet mlt tileset and viewed it using https://github.com/maplibre/maplibre-gl-js/pull/6570 but ran into some rendering artifacts https://github.com/maplibre/maplibre-tile-spec/issues/640. It seems they are issues with the ts decoder so I think should not block merging this PR.

Followup tasks are mostly in the mlt repo:
- remove ORC dependency to bring planetiler jar back down close to where it was before mlt
- improve encoding performance
- investigate why mlt tile size reduction isn't as big as we'd expect

planetiler-core/src/main/java/com/onthegomap/planetiler/pmtiles/ReadablePmtiles.java
TimSylvester: It's still under discussion.  At the moment, I would argue that if we're going to do anything very different it should be considered a different format, e.g., MRT for raster, even if it's based on the same top-level container format and even shares a single range of tag values.  So clients would generally be expected to implement all the features, unless they are specialty clients that are only using their own private data sources.

In the case of new features in future tags, it would be necessary to have multiple sources for legacy version support, or version negotiation with a tile server.

nyurik: @msbarry I think "per-column" extendability is less stable/certain than "per-layer". In other words, anyone who supports `layer tag=1` must support all features we define, i.e. all data types, all compression methods, etc - tag1 defines a package of support. Same will happen for tag2. Each layer can be easily skipped if unrecognized.

Issue #1344, opened 2025-10-01
[FEATURE] [Custom] Allow setting the overlay metadata


zstadler:
**Is your feature request related to a problem? Please describe.**
Add the ability to declare the tile-set as an overlay.

**Describe the solution you'd like**
Add a `is_overlay` root attribute that sets the `Profile::isOverlay` value. The default would be `false`.

PR #1363, opened 2025-10-16
Allow setting the overlay metadata

zstadler:
Add an `is_overlay` schema root attribute.

Resolve https://github.com/onthegomap/planetiler/issues/1344


msbarry:
Thanks!

PR #1380, opened 2025-10-30
Add schema validation to Maven

zstadler:
Also, add simplified WKT geometry pattern to schema.

Schema validation limitations:
- The plugin does not support validation of schema violations. See https://github.com/dataliquid/json-yaml-validator-maven-plugin/issues/51
- The plugin does not support YAML aliases. See https://github.com/dataliquid/json-yaml-validator-maven-plugin/issues/52

### DIFF
--- a/planetiler-custommap/planetilerspec.schema.json
+++ b/planetiler-custommap/planetilerspec.schema.json
@@ -25,11 +25,19 @@
               },
               "geometry": {
                 "description": "Geometry type of the input feature",
-                "type": "string",
-                "enum": [
-                  "polygon",
-                  "line",
-                  "point"
+                "oneOf": [
+                  {
+                    "type": "string",
+                    "enum": [
+                      "polygon",
+                      "line",
+                      "point"
+                    ]
+                  },
+                  {
+                    "type": "string",
+                    "pattern": "^(POLYGON|LINESTRING|POINT) "
+                  }
                 ]
               },
               "tags": {
--- a/planetiler-custommap/pom.xml
+++ b/planetiler-custommap/pom.xml
@@ -63,6 +63,73 @@
           <skipPublishing>true</skipPublishing>
         </configuration>
       </plugin>
+
+      <plugin>
+        <groupId>com.dataliquid</groupId>
+        <artifactId>json-yaml-validator-maven-plugin</artifactId>
+        <version>1.0.0</version>
+        <executions>
+
+          <execution>
+            <id>validate-positive-yaml</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>validate</goal>
+            </goals>
+            <configuration>
+              <sourceDirectory>.</sourceDirectory>
+              <includes>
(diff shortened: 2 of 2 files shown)


msbarry:
Instead of doing this through maven what if we add a json schema validator as a dependency. Then we could have the unit tests that validate these files also check they match the json schema. It would also let us add a `planetiler.jar validate` command that users could use to validate one of their files, or even just do it automatically before running a yaml file.

zstadler:
I have no strong feelings about this approach. Having a `validate` and a `validate-spec` commands would be useful, and the test could potentially use them also for testing with invalid inputs.
Could you take the lead on this?

basmussen:
Hi @zstadler,

Great news! We've just released https://github.com/dataliquid/json-yaml-validator-maven-plugin/releases/tag/1.1.0 of the json-yaml-validator-maven-plugin which addresses both issues you reported:

- **Issue 51** (Check the schema identifies invalid inputs) - ✅ Implemented with `expectedResult`, `expectedErrors` and `strictErrorMatching` parameters
- **Issue 52** (YAML aliases treated as strings) - ✅ Fixed, YAML aliases now properly resolve to their referenced values

Thank you for reporting these issues - they've made the plugin much more robust!

### About Our JSON Validation Approach

When we researched JSON validators a few months ago, we found that while many validators exist, only the networknt/json-schema-validator truly adheres correctly to the JSON Schema specification. Many other validators we tested had issues with spec compliance, particularly with complex features like conditionals, $ref resolution, and proper error reporting.

That's why we built this Maven plugin around networknt/json-schema-validator. Our experience has shown it handles even complex JSON validations excellently, including:
- Conditional validations (`if`/`then`/`else`)
- Complex schema composition (`allOf`, `oneOf`, `anyOf`)
- Full YAML support with proper alias resolution
- All JSON Schema versions (Draft 4 through 2020-12)

### Example with v1.1.0's New Features

```xml
<plugin>
    <groupId>com.dataliquid.maven</groupId>
    <artifactId>json-yaml-validator-maven-plugin</artifactId>
    <version>1.1.0</version>
    <executions>
        <!-- Validate that invalid configs are properly rejected -->
        <execution>
            <id>validate-invalid-configs</id>
            <goals><goal>validate</goal></goals>
            <configuration>
                <schemaFile>src/main/resources/planetiler.schema.json</schemaFile>
                <sourceDirectory>src/test/resources/invalid-configs</sourceDirectory>
                <expectedResult>ERROR</expectedResult>
                <expectedErrors>
                    <expectedError>.*storage.*required.*</expectedError>
                    <expectedError>.*threads.*minimum.*</expectedError>
                </expectedErrors>
            </configuration>
        </execution>
    </executions>
</plugin>
```

I see @msbarry suggested an alternative approach with direct dependency integration. Both approaches could actually complement each other well - our plugin for build-time validation and the direct dependency for runtime validation and CLI commands.

Feel free to try v1.1.0 if you'd like to continue with the Maven plugin approach. We're happy to help if you encounter any issues!

PR #1397, opened 2025-11-17
When merging features, assign IDs ending in 0 to the result

### LINKED ISSUE #1393: [FEATURE] FeatureMerge should (optionally?) assign new IDs to merged features instead of reusing one of the input feature's IDs

When tiling OSM data, Planetiler currently assigns IDs to tile features that encode the type and ID of the input OSM element. The formula is `osm_id * 10 + (1 for nodes, 2 for ways, 3 for relations)`. This reversible encoding is useful for applications where you'd like to hyperlink from a map feature back to the corresponding OSM element. One example is the tileset preview tool on tiles.openstreetmap.us ([example](https://tiles.openstreetmap.us/vector/openmaptiles/); click a map feature to see a popup that includes a link back to OSM). But this is also useful in maps meant for end users where you wish to connect things on the map back to the source of the information in OSM.

When merging features together with FeatureMerge.java, multiple OSM elements may end up combined in the tiles into a single tile feature. But because the merge logic reuses an ID from one of the input elements, a map that decodes this ID will incorrectly connect the composite feature back to a single OSM element, which can lead to confusing behavior like this: https://github.com/osmus/tileservice/issues/142

Feature merging will inevitably break the 1:1 correspondence between tile features and OSM elements. However, it would be helpful if the merged features had IDs that made it clear that they did not correspond directly to a single OSM element, so that data consumers could avoid mistakenly linking them. IDs that end in `0` appear to already be reserved for a similar purpose (features that do not come from OSM). So I was thinking that a solution to this might be to have FeatureMerge.java assign merged features an ID from the input set (as it does today), but modified to end in 0, which should preserve ID uniqueness but also encode the fact that this feature does not correspond to any single OSM element.

If this solution sounds good to you, I'm happy to work on a pull request.

**Describe alternatives you've considered**
Removing IDs entirely from modified elements would also work, but might cause other problems. For example, in MapLibre, `feature-state` can only be used on features that have an ID, which I think makes it impossible to style an individual feature e.g. highlight it on hover or when a user clicks on it. So I think preserving unique IDs on merged features is probably desired for most use cases.

**Additional context**
- https://github.com/osmus/tileservice/issues/142

jake-low:
In FeatureMerge.java, merged features are now given a new ID ending in 0 instead of reusing one of the input feature IDs. This indicates the tile feature doesn't correspond to a single source element.

For OSM features where IDs encode as osm_id * 10 + element_type, IDs ending in 0 are reserved for non-OSM or composite features. This helps prevent data consumers from incorrectly linking merged features back to a particular OSM element.

Fixes #1393.

I added new tests for this change and updated some existing tests which were expecting the old behavior in their assertions. Note that there are also a couple of tests in the planetiler-openmaptiles submodule that will need to be updated if this change is merged. Happy to open a PR for that too, but wasn't sure if that should be done in tandem with this one or afterwards.

I also tested this by installing my modified version of Planetiler and using it to build [Sourdough tiles](https://github.com/jake-low/sourdough), then viewed the resulting PMTiles archive and verified that merged features have IDs ending in 0, and unmerged features have their original IDs (encoding the OSM type and ID of the source feature).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           # don't need to check planetiler-openmaptiles
           submodules: false
--- a/.github/workflows/docs_mlc_config.json
+++ b/.github/workflows/docs_mlc_config.json
@@ -2,6 +2,9 @@
   "ignorePatterns": [
     {
       "pattern": "^http://localhost.*$"
+    },
+    {
+      "pattern": "^.*medium.com.*$"
     }
   ],
   "retryCount": 5,
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         submodules: true
     - name: Set up JDK 21
@@ -38,7 +38,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         submodules: true
     - name: Set up JDK ${{ matrix.jdk }}
@@ -62,7 +62,7 @@ jobs:
     timeout-minutes: 15
     continue-on-error: true
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         submodules: true
     - name: Set up JDK 21
@@ -88,7 +88,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         submodules: true
     - name: Cache data/sources
@@ -125,7 +125,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         submodules: true
     - name: Cache data/sources
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -27,12 +27,12 @@ jobs:
       with:
         access_token: ${{ github.token }}
     - name: 'Checkout branch'
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
       with:
         path: branch
         submodules: true
     - name: 'Checkout base'
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
       with:
         path: base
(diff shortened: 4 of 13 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
```diff
@@ -68,6 +68,20 @@ public class FeatureMerge {
...
+   * <p>
+   * For OSM features, IDs are encoded as {@code osm_id * 10 + element_type} where element_type is 1 for nodes, 2 for
+   * ways, and 3 for relations. IDs ending in 0 are reserved for features that don't map to a single OSM element.
+   *
+   * @param id the original feature ID
+   * @return the modified ID ending in 0
+   */
+  private static long makeMergedId(long id) {
+    return (id / 10) * 10;
+  }
```
msbarry, line 83: Thanks for adding this! I'm realizing this will update IDs on all elements, not just OSM elements. I'm not sure that is too important as the IDs are less meaningful in other sources.

I just double checked and confirmed that all features (even non-osm ones) get multiplied by `config.featureSourceIdMultiplier()` so it should be safe to do this for all features without causing unwanted collisions. Could we use that instead of a hardcoded 10 constant though?

planetiler-core/src/main/java/com/onthegomap/planetiler/FeatureMerge.java
```diff
@@ -141,7 +155,13 @@ private static List<VectorTile.Feature> mergeGeometries(
...
           .map(d -> d.feature.geometry())
           .forEachOrdered(combined);
-        result.add(feature1.copyWithNewGeometry(combined.finish()));
+        result.add(new VectorTile.Feature(
+          feature1.layer(),
+          makeMergedId(feature1.id()),
+          combined.finish(),
+          feature1.tags(),
+          feature1.group()
+        ));
```
msbarry: Can we extract this into a helper on VectorTile.Feature - like `.copyWithIdAndGeometry(id, geometry)` ?

jake-low:
Thanks for the review! I implemented the suggested `VectorTile.Feature.copyWithIdAndGeometry()` helper function, and I added an `exclude.txt` list based on the PR you linked in order to temporarily disable the failing OMT tests.

I wasn't sure how to use `config.featureSourceIdMultiplier()` instead of the hardcoded constant. I don't think FeatureMerge has access to `config` itself, so I think the caller would have to pass either the config or the multiplier in as an argument when calling a merge function such as `mergeLineStrings()`? This would be a breaking change to the API though, unless we provide a version that uses a default value like 10.

There are still two checks that are failing. One is the standalone example, which I think is failing since the tests have been changed to expect the new behavior but it's fetching the current 0.9.4 build to test with, which has the old behavior. The other is the link checker, which looks like it's failing to fetch some links from Medium, so probably unrelated to these changes? FWIW I can load those links in my browser just fine, so maybe Medium is blocking requests from GitHub Actions as part of an anti-scraping effort.

msbarry:
Looks good! Thanks for making the change. I fixed that doc issue on main and just merged it in, but the examples test failure is an issue any time we make a breaking change.  It's a bit tedious to wire-through the config, I can try to do that in a followup.

PR #1403, opened 2025-11-22
fix(shortbread): Move shortbread off the admin-points shapefile by geofabrik

CommanderStorm:
Geofabrik does not seem to supply this anymore.

@MichaelKreil Seems to have been smart and downloaded that to git.

### DIFF
--- a/planetiler-custommap/src/main/resources/samples/shortbread.yml
+++ b/planetiler-custommap/src/main/resources/samples/shortbread.yml
@@ -13,9 +13,6 @@ sources:
   ocean:
     type: shapefile
     url: https://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip
-  admin_points:
-    type: shapefile
-    url: https://github.com/shortbread-tiles/shortbread-tilemaker/raw/refs/heads/main/data/admin-points-4326/admin_points.shp
   osm:
     type: osm
     url: '${ args.osm_url }'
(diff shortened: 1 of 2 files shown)


planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -15,7 +15,7 @@ sources:
     url: https://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip
   admin_points:
     type: shapefile
-    url: https://shortbread.geofabrik.de/shapefiles/admin-points-4326.zip
+    url: https://github.com/shortbread-tiles/shortbread-tilemaker/raw/refs/heads/main/data/admin-points-4326/admin_points.shp
```
CommanderStorm: okay that won't work a zip'ed folder != shapefile

Any chance you might accept a .zip in this repo? @msbarry

msbarry: I think this is probably fine, but what exactly does that shapefile contain? Does tilekiln use it for osm.org or does it generate those points a different way? I'm wondering if it's strictly necessary.

On a side note it would probably be good to add shortbread to CI here to make sure you can always run it from scratch.

PartTimeDataScientist: > I think this is probably fine, but what exactly does that shapefile contain? Does tilekiln use it for osm.org or does it generate those points a different way? I'm wondering if it's strictly necessary.
> 
> On a side note it would probably be good to add shortbread to CI here to make sure you can always run it from scratch.

Description of the file contents is here: https://github.com/shortbread-tiles/shortbread-tilemaker/blob/main/boundary_labels_shape_file.md

Could be that it's not even needed for Planetiler. For me it was the easy fix to add the small .zip to Github and download from there (#1403) in order to get it working again. Other stuff is out of my capabilites.

Depending on how much work it is to check that dependency and/or work around it this could also be a temporary fix and another ticket for getting rid of the dependency...

CommanderStorm: Good point.

Lets solve it properly. Planetiler does not need this workaround since `polygon_point_on_surface` exists

planetiler-custommap/src/main/resources/samples/shortbread.yml
```diff
@@ -378,29 +375,27 @@ layers:
...
-        2: '${ feature.tags.has("ADMIN_LEVEL", "2") && double(feature.tags.WAY_AREA) >= 2e8 }'
-        3: '${ double(feature.tags.WAY_AREA) >= 7e7 }'
-        4: '${ double(feature.tags.WAY_AREA) >= 1e7 }'
-    # TODO sort by WAY_AREA descending
+        2: '${ feature.tags["admin_level"] == "2" && feature.area("km2") >= 2e6 }'
+        3: '${ feature.area("km2") >= 7e5 }'
+        4: '${ feature.area("km2") >= 1e5 }'
+    include_when:
+      __all__:
+      - boundary: administrative
```
CommanderStorm, line 388: implicit/undocumented assumption upstream.

Map looks a bit weird without this as demonstrated in the DMV testcase ^^

msbarry:
Looks good to me! Thanks for making the change. I'll merge this now to unblock people but there is also an "innermost point" utility you can use instead of point on surface that generally produces nicer looking results if you want.

PR #1404, opened 2025-11-24
Improve name subkey validation

### LINKED ISSUE #1162: Accept Unicode extensions in localized name subkeys
The code for validating `name:*=*` keys rejects extension syntaxes that are allowed under BCP&nbsp;47.

For example, [this peak in Hong Kong](https://www.openstreetmap.org/node/7655049003) has `name:en-t-zh=*` set to an English translation of a Chinese name. While it’s unlikely that a user will specifically ask for a map labeled in names based on mode of derivation, they might be interested in labels that take one side or another in a geopolitical dispute. [This lake](https://www.openstreetmap.org/relation/223999) whose name is disputed between two U.S. states. I’ve tagged each state’s recognized name in `name:en-u-sd-usnc=*` and `name:en-u-sd-usva=*`, using [Unicode subdivision identifiers](https://unicode.org/reports/tr35/#UnicodeSubdivisionIdentifier). This syntax is also being suggested for [a more prominent dispute](https://community.openstreetmap.org/t/gulf-of-america-gulf-of-mexico/124571/127) in current events.

These keys fail the test because it doesn’t account for the one-letter `-t-` or `-u-` extension marker:

https://github.com/onthegomap/planetiler/blob/4ecb02d136fca611b920abb17e8d60e66f21d8d2/planetiler-core/src/main/java/com/onthegomap/planetiler/util/LanguageUtils.java#L13

Instead of continuing to patch this homegrown regular expression, we could look around for more comprehensive ones such as [this JavaScript implementation](https://github.com/sebinsua/ietf-language-tag-regex/).

1ec5:
Replaced the regular expression test with one based on preflighting a `Locale` object.

Fixes #1162.


planetiler-core/src/test/java/com/onthegomap/planetiler/util/LanguageUtilsTest.java
```diff
@@ -30,37 +30,67 @@ void testRemoveNonLatin(String in, String out) {
...
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "nombre",
+    "",
+    "xxxxx",
+    "TEC",
+    "ja_rm",
+    "ja_kana",
+    "zh_pinyin",
+    "zh_zhuyin",
```
1ec5, line 73: I don’t think Planetiler should support `name:*=*` subkeys with underscores. They’ve been [deprecated](https://wiki.openstreetmap.org/wiki/Special:Diff/2430035/2451776) in favor of valid BCP&nbsp;47 equivalents [for over a decade](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/1235934). Some of these subkeys are still in use, but steadily declining in favor of the standard subkeys, while others are virtually nonexistent at this point.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/LanguageUtils.java
```diff
@@ -1,13 +1,19 @@
...
-  // Name tags that should be eligible for finding a latin name.
-  // See https://wiki.openstreetmap.org/wiki/Multilingual_names
+  /**
+   * Name tags that should be eligible for finding a latin name.
+   * 
+   * @see https://wiki.openstreetmap.org/wiki/Multilingual_names
+   * @deprecated Use {@code isValidLanguageTag(tag))}
+   */
+  @Deprecated(forRemoval = true)
   public static final Predicate<String> VALID_NAME_TAGS =
```
1ec5, line 18: openmaptiles/planetiler-openmaptiles#285 migrates away from this constant to take advantage of the improvements in this PR. The two PRs are loosely coupled, but both will need to land in order to actually fix #1162.

planetiler-core/src/main/java/com/onthegomap/planetiler/util/LanguageUtils.java
```diff
@@ -76,8 +82,40 @@ public static String removeLatinCharacters(String name) {
...
+      return false;
+    }
+
+    // Check whether the code is well-formed according to BCP 47.
+    Locale locale;
+    try {
+      locale = Locale.forLanguageTag(tag);
+    } catch (NullPointerException e) {
+      return false;
+    }
```
msbarry: Minor: I think forLanguageTag only throws an NPE if tag is null, we could just check `if (tag == null || tag.isEmpty()) {` at the very beginning and not have to catch this here

planetiler-core/src/test/java/com/onthegomap/planetiler/util/LanguageUtilsTest.java
msbarry: Got it, sounds good to me

planetiler-core/src/main/java/com/onthegomap/planetiler/util/LanguageUtils.java
1ec5: Oh, darn, I was hoping it would also throw on a malformed language tag, but it does seem to be more lenient than that.

1ec5: Looks like [the equivalent builder method](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.Builder.html#setLanguageTag-java.lang.String-) is stricter. I’ll add some tests to confirm that we’re able to detect malformed language tags, not just ones that result in an undefined language according to the parser.

planetiler-core/src/test/java/com/onthegomap/planetiler/util/LanguageUtilsTest.java
```diff
@@ -30,37 +30,69 @@ void testRemoveNonLatin(String in, String out) {
...
-    "name:be-tarask",
-    "name:ja_rm",
-    "name:ja_kana",
-    "name:vls",
-    "name:zh-hant-CN",
-    "name:zh_pinyin",
-    "name:zh_zhuyin",
-    "name:zh-Latn-tongyong",
-    "name:zh-Latn-pinyin",
-    "name:zh-Latn-wadegiles",
```
1ec5, line 47: This is actually invalid. It should be `zh-Latn-wadegile`, since a subtag can only be eight characters long at most.

msbarry:
Looks good! Thanks!

PR #1410, opened 2025-11-25
Update planetiler-openmaptiles submodule

1ec5:
Updated the planetiler-openmaptiles submodule to incorporate openmaptiles/planetiler-openmaptiles#285, which goes with #1404.


1ec5:
Not sure why `testLatinFallbacks()` is [failing](https://github.com/onthegomap/planetiler/actions/runs/19687741648/job/56396907268?pr=1410).

msbarry:
Looks like planetiler-openmaptiles is pinned to the last release not the latest snapshot version of planetiler so your pr in that repo didn't pick up that change.  I think the latest change must have made that test fail. 

We probably want planetiler-openmaptiles to use the latest snapshot. But we could also disable the omt tests temporarily using this suggestion I made in Jake's PR https://github.com/onthegomap/planetiler/pull/1397#pullrequestreview-3487336784

PR #1427, opened 2026-01-01
[FEATURE] Add relation_members geometry type for processing OSM relation members

### LINKED ISSUE #1426: [FEATURE] Add `relation_members` geometry type for processing OSM relation members
## Summary

Add a new `geometry: relation_members` option to YAML custom maps that allows processing individual members of OSM relations as separate features. Instead of emitting one feature per relation, this emits one feature per qualifying member, using each member's geometry and tags.

**Key Features:**
- Select relations using existing `include_when`/`exclude_when` filters
- Emit one feature per qualifying member (way/node)
- Filter members by type, role, and tags
- Add member-level attributes from member tags
- Support inline script expressions for member filtering and attributes

## Problem

Currently, when working with OSM relations in YAML custom maps, you can only emit a single feature for the entire relation. However, many use cases require processing individual members of a relation as separate features:

- **Route relations** (`type=route`): Extract individual road/rail segments that make up a route
- **Boundary relations** (`type=boundary`): Extract individual boundary segments
- **Public transport routes**: Extract individual stops or route segments
- **Waterway relations**: Extract individual river segments

For example, a `type=route` relation with `route=railway` might contain 50 way members representing different track segments. Currently, you'd get one feature for the relation, but you need 50 features - one for each segment - with each segment's own geometry and tags.

## Proposed Solution

Add a new `geometry: relation_members` option that:

1. **Selects relations** using the existing `include_when`/`exclude_when` filters (applied to relation tags)
2. **Emits one feature per qualifying member** instead of one feature per relation
3. **Uses each member's geometry** (way → line/polygon, node → point)
4. **Supports member-level filtering** by type, role, and tags
5. **Supports member-level attributes** from member tags

### 1. Railway Route Segments
Extract individual track segments from railway route relations:

```yaml
layers:
  - id: railway_segments
    features:
      - source: osm
        geometry: relation_members
        include_when:
          type: route
          route: railway
        member_types: [way]
        member_include_when:
          railway: rail
        attributes:
          - key: route_name
            tag_value: name
        member_attributes:
          - key: segment_ref
            tag_value: ref
```

### 2. Administrative Boundary Segments
Extract individual boundary segments from administrative boundary relations:

```yaml
layers:
  - id: boundary_segments
    features:
      - source: osm
        geometry: relation_members
        include_when:
          type: boundary
          admin_level: "6"
        member_types: [way]
        member_roles: ["outer"]
        member_attributes:
          - key: boundary_type
            tag_value: boundary
```

### 3. Bus Route Stops
Extract individual stops from public transport route relations:

```yaml
layers:
  - id: bus_stops
    features:
      - source: osm
        geometry: relation_members
        include_when:
          type: route
          route: bus
        member_types: [node]
        member_roles: ["stop"]
        member_include_when:
          public_transport: stop_position
        attributes:
          - key: route_ref
            tag_value: ref
        member_attributes:
          - key: stop_name
            tag_value: name
```

### New Geometry Type

```yaml
geometry: relation_members
```

When used, the feature selection (`include_when`/`exclude_when`) applies to **OSM relations**, but features are emitted for **each qualifying member** using the member's geometry.

### New Configuration Fields

All fields are optional and only valid with `geometry: relation_members`:

#### `member_types`
Filter members by OSM element type.

- **Type**: Array of strings
- **Values**: `"node"`, `"way"`, `"relation"` (or subset)
- **Default**: All types (if not specified)

```yaml
member_types: [way]           # Only process way members
member_types: [node, way]     # Process nodes and ways
```

#### `member_roles`
Filter members by their role in the relation.

- **Type**: Array of strings
- **Default**: All roles (if not specified)
- **Note**: Empty string `""` matches members with no role

```yaml
member_roles: ["outer", "inner"]  # Only process outer/inner members
member_roles: [""]                # Only process members with no role
member_roles: ["stop", "platform"] # Only process stop/platform members
```

#### `member_include_when`
Filter members by their tags. Supports both structured expressions and inline script expressions.

- **Type**: Boolean expression (structured or inline script)
- **Default**: Include all members (if not specified)

**Structured example:**
```yaml
member_include_when:
  railway: rail
  highway: [primary, secondary]
```

**Inline script example:**
```yaml
member_include_when: '${ member.tags.has("highway") && member.tags.highway == "primary" }'
```

#### `member_exclude_when`
Exclude members by their tags. Applied after `member_include_when`.

- **Type**: Boolean expression (structured or inline script)
- **Default**: No exclusions (if not specified)

```yaml
member_exclude_when:
  service: __any__
```

#### `member_attributes`
Add attributes to each member feature from the member's tags.

- **Type**: Array of attribute definitions
- **Supports**: All attribute features including inline script expressions

```yaml
member_attributes:
  - key: segment_ref
    tag_value: ref
  - key: segment_name
    tag_value: name
  - key: combined_name
    value: '${ member.tags.name + " (" + feature.tags.name + ")" }'
```

### Member Context Variables

When using inline script expressions in `member_include_when`, `member_exclude_when`, or `member_attributes`, the following variables are available:

- `member.tags` - Map of tags from the member element (way/node)
- `member.role` - Role of the member in the relation (string, may be empty)
- `member.type` - OSM element type: `"node"`, `"way"`, or `"relation"` (string)
- `member.ref` - OSM ID of the member element (long)
- `member.id` - Same as `member.ref` (long)
- `feature.tags` - Map of tags from the parent relation (inherited from parent context)
- All other variables from parent contexts (`args`, `feature.id`, etc.)

## Complete Example

```yaml
layers:
  - id: route_segments
    features:
      - source: osm
        geometry: relation_members
        include_when:
          type: route
          route: railway
        member_types: [way]
        member_roles: [""]  # empty role
        member_include_when:
          railway: rail
        member_exclude_when:
          service: __any__
        attributes:
          - key: route_name
            tag_value: name
          - key: route_ref
            tag_value: ref
        member_attributes:
          - key: segment_ref
            tag_value: ref
          - key: segment_name
            tag_value: name
          - key: combined_name
            value: '${ member.tags.name + " (" + feature.tags.name + ")" }'
```

## Behavior Notes

1. **Feature Selection**: `include_when`/`exclude_when` filters apply to the **relation**, not individual members
2. **Member Filtering Order**: Type → Role → `member_include_when` → `member_exclude_when`
3. **Feature IDs**: Each member feature gets a unique ID generated from relation ID and member reference
4. **Duplicate Members**: If a relation contains the same member multiple times, only the first occurrence is processed
5. **Missing Members**: Missing ways/nodes (common in extracts) are skipped with a logged error
6. **Nested Relations**: Nested relation members are currently skipped (not recursively processed)
7. **Empty Relations**: If all members are filtered out, no features are emitted
8. **Geometry Types**: 
   - Nodes → point geometry
   - Ways → line or polygon geometry (determined by whether way is closed and `area` tag)
   - Relations → skipped (not supported)
9. **Attributes**: Both relation-level attributes (`attributes`) and member-level attributes (`member_attributes`) are applied to each member feature
10. **Post-processing**: Post-processing operations like `merge_linestrings` can be used with `relation_members` geometry

## Validation

- `member_types`, `member_roles`, `member_include_when`, `member_exclude_when`, and `member_attributes` can **only** be used with `geometry: relation_members`
- `member_types` values must be `"node"`, `"way"`, or `"relation"`
- Clear error messages if these fields are used with other geometry types

## Related

This feature follows the same pattern as multipolygon processing, where relation members are processed in two phases:
- **Phase 1**: Identify relations matching `relation_members` features
- **Phase 2**: Store member geometries/tags, then process relations to emit member features

zstadler:
Add a new `geometry: relation_members` option to YAML custom maps that allows processing individual members of OSM relations as separate features. Instead of emitting one feature per relation, this emits one feature per qualifying member, using each member's geometry and tags.

Key features:
- Select relations using existing include_when/exclude_when filters
- Emit one feature per qualifying member (way/node)
- Filter members by type, role, and tags (with inline script support)
- Add member-level attributes from member tags
- Support inline script expressions for member filtering and attributes

New configuration fields:
- member_types: Filter members by OSM element type (node, way, relation)
- member_roles: Filter members by their role in the relation
- member_include_when: Filter members by their tags (structured or inline script)
- member_exclude_when: Exclude members by their tags (structured or inline script)
- member_attributes: Add attributes from member tags (with inline script support)

Implementation follows the multipolygon processing pattern:
- Phase 1: Identify relations matching relation_members features
- Phase 2: Store member geometries/tags, then process relations to emit member features

Includes comprehensive validation:
- Member fields can only be used with relation_members geometry
- member_types values must be valid OSM element types
- Clear error messages for invalid configurations

Resolves #1426

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
@@ -3,6 +3,7 @@ package com.onthegomap.planetiler.reader.osm;
 import static com.onthegomap.planetiler.util.MemoryEstimator.estimateSize;
 import static com.onthegomap.planetiler.worker.Worker.joinFutures;
 
+import com.carrotsearch.hppc.IntArrayList;
 import com.carrotsearch.hppc.IntObjectHashMap;
 import com.carrotsearch.hppc.LongArrayList;
 import com.carrotsearch.hppc.LongObjectHashMap;
@@ -116,6 +117,8 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
   private final IntObjectHashMap<String> roleIdsReverse = new IntObjectHashMap<>();
   private final AtomicLong roleSizes = new AtomicLong(0);
   private final OsmPhaser pass1Phaser = new OsmPhaser(0);
+  private final OsmWaySplitter waySplitter = OsmWaySplitter.roaringBitmapSplitter();
+  private final AtomicLong maxWayId = new AtomicLong(Long.MIN_VALUE);
 
   /**
    * Constructs a new {@code OsmReader} from an {@code osmSourceProvider} that will use {@code nodeLocationDb} as a
@@ -247,14 +250,18 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
     LOGGER.debug("Processed " + FORMAT.integer(PASS1_BLOCKS.get()) + " blocks:");
     pass1Phaser.printSummary();
     timer.stop();
+
+    waySplitter.finish();
   }
 
   void processPass1Blocks(Iterable<? extends Iterable<? extends OsmElement>> blocks) {
     // may be called by multiple threads so need to synchronize access to any shared data structures
+    long threadLocalMaxWayId = Long.MIN_VALUE;
     try (
       var nodeWriter = nodeLocationDb.newWriter();
       var phases = pass1Phaser.forWorker()
-        .whenWorkerFinishes(OsmPhaser.Phase.NODES, nodeWriter::close)
+        .whenWorkerFinishes(OsmPhaser.Phase.NODES, nodeWriter::close);
+      var waySplitWriter = waySplitter.writerForThread();
     ) {
       for (var block : blocks) {
         for (OsmElement element : block) {
@@ -266,7 +273,7 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
             try {
               profile.preprocessOsmNode(node);
             } catch (Exception e) {
-              LOGGER.error("Error preprocessing OSM node " + node.id(), e);
+              LOGGER.error("Error preprocessing OSM node {}", node.id(), e);
             }
             // TODO allow limiting node storage to only ones that profile cares about
             nodeWriter.put(node.id(), node.encodedLocation());
@@ -274,9 +281,13 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
             phases.arriveAndWaitForOthers(OsmPhaser.Phase.WAYS);
             try {
               profile.preprocessOsmWay(way);
+              if (profile.splitOsmWayAtIntersections(way)) {
+                waySplitWriter.addWay(way.nodes());
+              }
             } catch (Exception e) {
-              LOGGER.error("Error preprocessing OSM way " + way.id(), e);
+              LOGGER.error("Error preprocessing OSM way {}", way.id(), e);
             }
+            threadLocalMaxWayId = Math.max(threadLocalMaxWayId, way.id());
           } else if (element instanceof OsmElement.Relation relation) {
             phases.arrive(OsmPhaser.Phase.RELATIONS);
             try {
@@ -309,7 +320,7 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
                 }
               }
             } catch (Exception e) {
-              LOGGER.error("Error preprocessing OSM relation " + relation.id(), e);
+              LOGGER.error("Error preprocessing OSM relation {}", relation.id(), e);
             }
             // TODO allow limiting multipolygon storage to only ones that profile cares about
             if (isMultipolygon(relation)) {
@@ -326,6 +337,7 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
         PASS1_BLOCKS.inc();
       }
     }
+    maxWayId.accumulateAndGet(threadLocalMaxWayId, Long::max);
   }
 
   private static boolean isMultipolygon(OsmElement.Relation relation) {
@@ -343,6 +355,7 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
     var timer = stats.startStage("osm_pass2");
     int writeThreads = config.featureWriteThreads();
     int processThreads = config.featureProcessThreads();
+    long splitWayMultiplier = getSplitWayMultiplier(maxWayId.get());
     Counter.MultiThreadCounter blocksProcessed = Counter.newMultiThreadCounter();
     // track relation count separately because they get enqueued onto the distributor near the end
     Counter.MultiThreadCounter relationsProcessed = Counter.newMultiThreadCounter();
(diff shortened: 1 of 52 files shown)


CommanderStorm (commented):
(not a maintainer)

🤔 Reealy stupid question:
Is the following possible with this PR, or is this another issue?

https://github.com/onthegomap/planetiler/blob/6eb3ab5706342c7baef1dd589c9d4ed6dfb1d604/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml#L1087-L1123

Currently, the US-Canada and russia-ukraine border are relations, so cannot be incuded via the custommap previously (?).

In general, could you add this to the custommap' Readme (aka the docs for this feature)

zstadler:
> Is the following possible with this PR, or is this another issue?

This PR would allow the boundary segments to be taken from the members of boundary relations, while taking attributes such as `maritime` and `disputed` from the relation member's tags.

If I understand it correctly, this PR could address this TODO in the `boundaries` layer.

https://github.com/onthegomap/planetiler/blob/6eb3ab5706342c7baef1dd589c9d4ed6dfb1d604/planetiler-custommap/src/main/resources/samples/shortbread.yml#L372-L373

This PR does **not** address the other `boundaries` TODO, as each applicable member creates a separate feature for each applicable relation it is a member of

https://github.com/onthegomap/planetiler/blob/6eb3ab5706342c7baef1dd589c9d4ed6dfb1d604/planetiler-custommap/src/main/resources/samples/shortbread.yml#L347

It would be a valuable to be able issue a single geometry for each member and be able to aggregate information for tags of the parent relations.

Note that [shortbread.spec.yml](https://github.com/onthegomap/planetiler/blob/main/planetiler-custommap/src/main/resources/samples/shortbread.spec.yml) is a test for [shortbread.yml](https://github.com/onthegomap/planetiler/blob/main/planetiler-custommap/src/main/resources/samples/shortbread.yml) were each name represents an example where the given input should produce the given output. This PR does not affect the schema testing mechanism. The quoted comment should probably be placed in the schema itself.

data/tmp/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
msbarry: The root .gitignore already excludes data directory, so I don't think we need this

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -296,6 +315,32 @@ void processPass1Blocks(Iterable<? extends Iterable<? extends OsmElement>> block
...
+                        } else if (member.type() == OsmElement.Type.NODE) {
+                          nodesInRelationMembers.add(member.ref());
+                        }
+                      }
+                    }
+                  }
+                  break;
+                }
+              }
+            }
```
msbarry: This seems very similar to the above processing in the `synchronized (wayToRelationsLock) { ...` block except that one processes ways in a relation when the way gets processed in pass 2, and this one processes ways and nodes in the relation in pass 2 when the relation gets processed, requiring more intermediate storage.

Can we merge this processing into the block above ? It seems like the only extra functionality that we need  would be to also track nodes that are in relations where preprocessOsmRelation returned something.  Then we can add relationInfo to NodeSourceFeature as well so that pass 2 node processing also has access to what way a node is a member of.  It seems like that should be sufficient for custommap to implement the same functionality?

data/tmp/.gitignore
zstadler: The purpose of the `.gitignore` file with the specific contents is to ensure the existence of `tmpdir=data/tmp` (temp directory).
If the `data/tmp` directory does not exist, and planetiler is run with docker, the directory will be created from within the container and most likely it will be owned by `root`.
I should have made it clear in the comment...

zstadler: Commit dcefcb61be2e7dd0c996296760c89656eeef4976 replaces this with a modification of `scripts/test-release.sh`.

zstadler:
@msbarry,
Most appearances of `multipolygonWayGeometries` in `OsmReader.java` have `relationMembersWayGeometries` counterparts.

Would you be able to look at the missing ones?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
zstadler: Please see [2fb0a10](https://github.com/onthegomap/planetiler/pull/1427/commits/2fb0a10b4fc6c0b4f7630991989c97adef755f69). Is that what you had in mind?

PR #1431, opened 2026-01-05
Fix #1390: Suppress stack traces for incomplete OSM relations

### LINKED ISSUE #1390: [FEATURE] Unnecessary stack trace for incomplete OSM relations
When using Planetiler with a country/region extract, there are usually some relations that are only partially included in the extract. For example, the Geofabrik extract for Ukraine does dot fully include the way members of [relation 6449148](https://www.openstreetmap.org/relation/6449148#map=7/49.138/24.467)  - the "Poloniny National Park" in neighboring Slovenia.

For each such relation, Planetiler log not only states the issue with the OSM data, such as
```
Error processing OSM Relation 6449148 ... no rings to process after fixing
```
but it also adds the stack trace, as if the Planetiler itself had a fault:

Image

1. Replace the `ERR` severity with the de-facto `INF` severity.
2. Replace the elaborated message with a single line. Perhaps
   ```
   0:05:50 INF [osm_pass2:process] - Error processing OSM Relation 6449148: no rings to process after fixing
   ```

zstadler:
When processing regional extracts, incomplete OSM relations (missing ways or nodes) are expected and common. Previously, these exceptions were logged with full stack traces, cluttering logs and obscuring more critical issues.

This change:
- Adds a helper method to detect incomplete relation exceptions
- Logs incomplete relation exceptions as warnings without stack traces
- Keeps full error logging with stack traces for unexpected errors

The fix handles exceptions related to:
- Missing nodes ("Missing location for node")
- Incomplete multipolygons (empty rings, not closed, missing ways)
- GeometryException with stats indicating incomplete relations

Includes a test that verifies incomplete relations are logged as warnings without stack traces, and fails if the fix is reverted.

Fixes #1390

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
@@ -259,8 +259,8 @@ public class OsmReader implements Closeable, MemoryEstimator.HasEstimate {
         return true;
       }
       // Check for GeometryException related to incomplete relations
-      if (e instanceof GeometryException) {
-        String stat = ((GeometryException) e).stat();
+      if (e instanceof GeometryException geometryException) {
+        String stat = geometryException.stat();
         if (stat != null && (stat.contains("invalid_multipolygon") || stat.contains("missing"))) {
           return true;
         }
--- a/planetiler-core/src/test/java/com/onthegomap/planetiler/reader/osm/OsmReaderTest.java
+++ b/planetiler-core/src/test/java/com/onthegomap/planetiler/reader/osm/OsmReaderTest.java
@@ -660,7 +660,7 @@ class OsmReaderTest {
   /**
    * Custom appender to capture log events for testing.
    */
-  private static class TestAppender extends AbstractAppender {
+  private static class TestAppender extends AbstractAppender implements AutoCloseable {
     private final List<LogEvent> logEvents = new ArrayList<>();
 
     protected TestAppender(String name) {
@@ -675,6 +675,11 @@ class OsmReaderTest {
     public List<LogEvent> getLogEvents() {
       return new ArrayList<>(logEvents);
     }
+
+    @Override
+    public void close() {
+      stop();
+    }
   }
 
   @Test
@@ -688,7 +693,7 @@ class OsmReaderTest {
     loggerConfig.addAppender(testAppender, Level.ALL, null);
     context.updateLoggers();
 
-    try {
+    try (testAppender) {
       // Create a profile that throws an incomplete relation exception during preprocessing
       Profile testProfile = new Profile.NullProfile() {
         @Override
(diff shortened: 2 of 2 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/reader/osm/OsmReader.java
```diff
@@ -284,7 +321,11 @@ void processPass1Blocks(Iterable<? extends Iterable<? extends OsmElement>> block
                 }
               }
             } catch (Exception e) {
-              LOGGER.error("Error preprocessing OSM relation " + relation.id(), e);
+              if (isIncompleteRelationException(e)) {
+                LOGGER.warn("Incomplete OSM relation {}: {}", relation.id(), e.getMessage());
+              } else {
+                LOGGER.error("Error preprocessing OSM relation " + relation.id(), e);
+              }
```
msbarry, line 328: Thanks for adding this! I already have some similar handling along these lines where these "verbose" exceptions are instances of "GeometryException.Verbose", and GeometryException has a built in `geometryException.log("Error processing relation")` or `geometryException.log("Error processing relation", stats, "osm_relation_error")` if we want to include these in the summary of geometry errors that gets printed at the end.

It seems like we might be able to replace this whole PR with just changing these 2 "catch (Exception e)" blocks to something like:

```java
} catch (GeometryException e) {
  e.log("Error processing relation " + relation.id());
} catch (Exception e) {
  LOGGER.error("Error preprocessing OSM relation {}", relation.id(), e);
}
```

zstadler: Frankly, I don't have a strong opinion about the appearance in the log, other than avoiding a scary ERROR and a useless stack trace.

Please feel free to change the PR code in any way you like.

msbarry: Actually could you share one or a few full stack traces that you're seeing? I tried adding that line but found that GeometryException is actually caught before that point and logged using that method that suppresses verbose logs already.

zstadler: There is a stack trace in https://github.com/onthegomap/planetiler/issues/1390#issue-3597723973

msbarry: Ahh I see, these geometry exceptions are normally caught and logged quietly (or not logged at all) but your profile asks for the area of a feature, and [`Unit.java`](https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/geo/Unit.java) calls `applyAndWrapException` which propagates any geometry exception that would have been suppressed very loudly.

What do you think the fallback should be when area/length fail to compute? Should it return 0? null? or omit the entire feature

Or more broadly, if any of the properties in a yaml file fail with an exception do you think we should just omit that property? or the entire feature?

zstadler: There are actually two questions here:

1. Should an incomplete polygon be processed at all? For example, should its attributes be computed?
IMO, if a potential feature does not have a valid geometry it should be dropped as soon as can be.

2. How to handle (other) exceptions while computing attribute values?
IMO, it would be best if there is a clear error message with the OSM feature id and type, the layer.feature.name and the attribute name.
That should allow users to fix their definition/code. 
Silent drop is generally not a good option. On the other hand, keeping the current stack trace is not very helpful for the user, and an unnecessary support burden for you.

msbarry: OK got it, what about this approach: https://github.com/onthegomap/planetiler/pull/1454 - it avoids calling attribute processing functions if we already failed to construct the geometry, and also logs any exceptions less verbosely if they do occur?

zstadler: If I understand correctly, #1454 should address question 1 in a more general way than this PR. 
You may want to use the tests added in this PR.

Question 2 is not addressed by this PR #1431 or the issue it would resolve - #1390.

msbarry: I think that [the ConfiguredProfile change](https://github.com/onthegomap/planetiler/pull/1454/changes#diff-aa63ec6d9d5a5d4b517be0ba95b883c34eb57d5ba3b7311c74f4b2bc3fb35554) would address question 2? This would end up logging at the same level of detail as if there is an error creating a map feature for an OSM input feature but it should be much less likely to happen because we'd skip that processing if the feature was invalid in the first place.

Do you have an example osm.pbf and yaml I could try that on to see what output looks like?

zstadler: > Do you have an example osm.pbf and yaml I could try that on to see what output looks like?

Please save [issue1390.yml](https://github.com/user-attachments/files/25069373/issue1390.yml) in an empty directory and run
```
docker run --rm -v ./:/w -w /w ghcr.io/onthegomap/planetiler issue1390.yml --download --output=issue1390.pmtiles
```

msbarry: OK Got it! I just ran that and got 0 warning or error messages with my PR. I'll merge that and close this one. Feel free to open another if you find more examples of bad logging.

PR #1447, opened 2026-01-19
handle GEOMETRYCOLLECTION

phanecak-maptiler:
Handle simple(-r) cases of GEOMETRYCOLLECTION with either only points, only lines or only polygons in the collection.

The GPKG spec (see for example chapter "Features consisting of multiple components" https://www.geopackage.org/guidance/modeling.html) seems to allow also more complicated cases (points, lines and polygons mixed in one collection). This for now results in "empty geometry" even with code from this PR.

### DIFF
--- a/planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
+++ b/planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
@@ -1645,7 +1645,6 @@ class PlanetilerTests {
     double lng6 = GeoUtils.getWorldLon(0.625);
     var polygon = rectangle(lng5, lat5, lng6, lat6);
 
-    // TODO collection of those
     var input = List.of(newReaderFeature(GeoUtils.JTS_FACTORY.createGeometryCollection(Set.of(
       point,
       line,


phanecak-maptiler:
For now there is one quirk:

- several polygons are generated for collection of polygons
- several points are generated for collection of points
- but multi-line is generated for collection of lines

See TODO note in `FeatureRenderedTest.java`. I've tried to avoid changing the "core" code too much hence the change works but unit test is omitted.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeometryType.java
```diff
@@ -37,6 +38,19 @@ public Expression featureTest() {
   }
 
   public static GeometryType typeOf(Geometry geom) {
+    if (geom instanceof GeometryCollection collection && collection.getNumGeometries() >= 1) {
+      var result = typeOfPrimitive(collection.getGeometryN(0));
+      for (int i = 1; i < collection.getNumGeometries(); i++) {
+        if (!result.equals(typeOfPrimitive(collection.getGeometryN(1)))) {
```
msbarry: Did you mean this?

```suggestion
        if (!result.equals(typeOfPrimitive(collection.getGeometryN(i)))) {
```

msbarry: Also is it possible we can have a geometry collection nested in a geometry collection?

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/SimpleFeature.java
```diff
@@ -191,19 +192,73 @@ public Map<String, Object> tags() {
...
+  }
+
+  private static boolean isPoint(Geometry geom) {
+    return geom instanceof Puntal;
+  }
+
   @Override
   public boolean isPoint() {
-    return latLonGeometry instanceof Puntal || worldGeometry instanceof Puntal;
+    return isPoint(latLonGeometry != null ? latLonGeometry : worldGeometry) || isCollectionOfPoints();
```
msbarry: Could you re-use GeometryType.typeOf for these?

planetiler-core/src/test/java/com/onthegomap/planetiler/reader/SimpleFeatureTest.java
```diff
@@ -48,10 +53,20 @@ void testFromWorldGeom() {
...
-  @Test
-  void testIsLine() {
-    var world = SimpleFeature.fromWorldGeometry(newLineString(0, 0, 1, 1));
-    var latLon = SimpleFeature.fromLatLonGeometry(newLineString(0, 0, 1, 1));
+  static List<Geometry> lines() {
+    var line = newLineString(0, 0, 1, 1);
+    var lines = List.of(line);
+    return List.of(
+      line,
+      GeoUtils.JTS_FACTORY.createGeometryCollection(lines.toArray(new Geometry[0]))
```
msbarry: Minor: I usually do `.toArray(Geometry[]::new)` since it just instantiates thew new array directly instead of needing to use the temp array, although this is just test code so I don't  think it makes any difference.

planetiler-core/src/test/java/com/onthegomap/planetiler/render/FeatureRendererTest.java
```diff
@@ -587,16 +620,24 @@ private FeatureCollector.Feature polygonFeature(Geometry geom) {
...
+    return List.of(
+      polygon,
+      GeoUtils.JTS_FACTORY.createGeometryCollection(polygons.toArray(new Geometry[0]))
+    );
+  }
+
+  @ParameterizedTest
+  @MethodSource("polygons")
+  void testSimpleTriangleCCW(Geometry polygon) {
+    var feature = polygonFeature(polygon)
```
msbarry, line 640: Could we also add an end to end test in PlanetilerTests that we the vector tiles end up containing the expected output when we process a GeometryCollection from with same or mixed input types?

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeometryType.java
phanecak-maptiler: > result.equals(typeOfPrimitive(collection.getGeometryN(1)

🤦‍♂️, adjusted. Thanks.

phanecak-maptiler: > Also is it possible we can have a geometry collection nested in a geometry collection?

As per spec, collection is geometry and collection contains geometries, hence I presume yes.

And I was also create such in unit tests ... so adjusted here to handle also that with recursion.

planetiler-core/src/main/java/com/onthegomap/planetiler/reader/SimpleFeature.java
phanecak-maptiler: Less code, nice. Adjusted. Thanks.

planetiler-core/src/test/java/com/onthegomap/planetiler/reader/SimpleFeatureTest.java
phanecak-maptiler: Adjusted, thanks.

planetiler-core/src/test/java/com/onthegomap/planetiler/render/FeatureRendererTest.java
phanecak-maptiler: Added.

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1441,6 +1539,93 @@ record TestRelationInfo(long id, String name) implements OsmRelationInfo {}
...
+            rectangleCoordList(0.25 * 256, 0.75 * 256)
+          )
+        ), Map.of(
+          "attr", "value",
+          "name", "name value"
+        ))
+      )
+    ), results.tiles);
+  }
+
```
msbarry, line 1629: Actually would you mind just adding a single mixed-type geometry collection test? I'm not 100% sure the renderer will handle that, but if it does that should cover all of the cases.

phanecak-maptiler: Added. (Sorry, overlooked the earlier comment.)

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1626,6 +1627,74 @@ void testPolygonCollection(List<SimpleFeature> polygons) throws Exception {
...
+          if (geom instanceof GeometryCollection collection) {
+            for (int i = 0; i < collection.getNumGeometries(); i++) {
+              features.geometry("layer", collection.getGeometryN(i))
+                .setZoomRange(0, 0)
+                .setAttr("name", "name value")
+                .inheritAttrFromSource("attr");
+            }
+          } else {
+            throw new RuntimeException("only collections expected in this test");
+          }
```
msbarry, line 1673: ah this is what I was concerned about... it seems like if you have a geometry collection with points in it you should be able to do `features.point` and it would pick up the points in that collection, `features.polygon` would pick up the polygons, and feature.anyGeometry would pick them all up.... 

That might be a more complicated change, or perhaps a simpler approach would be when the reader gets to a geometry collection, it traverses the nested features and calls the profile separately with each of them?

I'm not too clear on the use-case. Would profiles need to be aware that these are one large feature? Or would it be sufficient for that to be extracted away?

phanecak-maptiler: Collections with only one type of feature are quite clear use-case and easy to do. And that's what I'm aiming for here since 1) we (MapTiler) are able to do a Java profiler and 2) that's what we've run into with customer data.

> That might be a more complicated change, or perhaps a simpler approach would be when the reader gets to a geometry collection, it traverses the nested features and calls the profile separately with each of them?

I was thinking in similar fashion, but given that I lack a well defined use-cases (plus time constraints), I've opted not to address that.

PR as is should do the trick for those having a Java profile. But it is not very relevant to those with a YAML custom profile.

The change might not be that drastic and would benefit custom profiles. But I've mentioned https://www.geopackage.org/guidance/modeling.html and there they mention for example "_Features consisting of multiple components_". And in that case the actual model/schema may or may not be relevant to a profile, hence we may also run into stuff similar to what is addressed in #1427 , e.g. have some more context, which feature relates to other feature, etc. For now I do not have picture clear enough to address that without overengineering.

So TL;DR: Could be addresses, but later, when I run into one or several input files with mixed collections and a person able to explain why it is like that and what is desired outcome, to make sure I'm not doing wrong assumptions.

msbarry: OK got it, that makes sense. Looks like the geometry collection would have a single ID and set of properties so it seems like it might cause issues to process them as separate features.  The only "safe" thing we might be able to do is have "features.line(...)" automatically extract the lines within the geometry collection but maybe better to wait for a concrete use-case to make sure we understand what this case actually means better

PR #1469, opened 2026-02-13
expose `setId()` for custommap/YAML profiles

phanecak-maptiler:
This PR exposes `setId()` for custommap/YAML profiles so that ID for generated feature can be set as follows:

- Primitive example: Just set the ID

```
sources:
        osm:
          type: osm
          url: geofabrik:rhode-island
          local_path: data/rhode-island.osm.pbf
      layers:
      - id: testLayer
        features:
        - source: osm
          geometry: polygon
          id: 1
```

- More practical example: pass the ID from a source feature

```
sources:
        osm:
          type: osm
          url: geofabrik:rhode-island
          local_path: data/rhode-island.osm.pbf
      layers:
      - id: testLayer
        features:
        - source: osm
          geometry: polygon
          id: ${ int(feature.id) }
```

Typical use-case: User has some input data with some IDs already there and wants them in the tileset as feature ID, not as attribute with name "id".

### DIFF
--- a/planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredFeatureTest.java
+++ b/planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredFeatureTest.java
@@ -1938,7 +1938,7 @@ class ConfiguredFeatureTest {
   @ParameterizedTest
   @CsvSource({
     "100, 100",
-    "${ int(feature.id) }, 1",
+    "${ feature.id }, 1",
   })
   void testId(String input, Long output) {
     String schema = """


planetiler-custommap/src/test/java/com/onthegomap/planetiler/custommap/ConfiguredFeatureTest.java
```diff
@@ -1934,4 +1934,28 @@ void testBuffer() {
       assertEquals(5, feature.getBufferPixelsAtZoom(14));
     }, 1);
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    "100, 100",
+    "${ int(feature.id) }, 1",
```
msbarry: Minor: Since you cast the result to `Long.class` in ConfiguredFeature you can omit the explicit `int( ... )` cast in your CEL expression:

```suggestion
    "${ feature.id }, 1",
```

phanecak-maptiler: Adjusted, thank you.

msbarry:
Thanks!

PR #1543, opened 2026-04-09
#1541 Fix z16 tile bugs

### LINKED ISSUE #1541: [BUG] Zoom 16 tile encoding: signed overflow in TiledGeometry and FeatureGroup sort key corruption
**Describe the bug**

The zoom 16 support commit (`1fc95a26`) introduces two related issues in tile ID encoding and sort key handling. A third related issue (sort key range reduction with assert-only validation) was already reported in #398 and is now exacerbated by the SORT_KEY_BITS change from 23 to 22.

**Bug 1 (Critical): `TiledGeometry.encode()` int-to-long sign extension causes silent tile loss**

`encode()` at `TiledGeometry.java:808` returns `int`. At zoom 16, `maxTilesAtThisZoom = 65536`. For any tile column `x >= 32768`, the expression `65536 * x` exceeds `Integer.MAX_VALUE` and wraps to a negative `int`.

At line 302, the result is assigned to `long colStart`, which sign-extends the negative int (e.g., `x=32768` → int `-2147483648` → long `-2147483648L`). `RoaringBitmap.addOffset(yRanges.bitmap(), colStart)` then applies a negative offset, causing all y-range values to underflow and be silently dropped.

**Impact:** All interior (filled) tiles for polygon columns with `x >= 32768` at z16 are lost from the `CoveredTiles` bitmap. When used with `--bounds` or `--polygon`, tiles in roughly the eastern half of the world at z16 are silently excluded.

**Fix:** `long colStart = Integer.toUnsignedLong(encode(...))` — matching the pattern already used on lines 784 and 802.

**Bug 2 (High): `FeatureGroup.encodeKey()` sets the sign bit, breaking sort order**

`encodeKey()` computes `tile << (64 - TILE_ID_BITS)` which is `tile << 31`. At z16, tile IDs can exceed `2^32`, which sets bit 63 (the sign bit) of the packed key. `SortableFeature.compareTo()` uses signed `long` comparison (`key < o.key`), so approximately one-third of z16 tiles sort before all z0–z15 tiles.

**Impact:** Tile iteration order is wrong for high z16 tiles. PMTiles output loses clustered ordering, and any order-sensitive writer/reader logic sees tiles out of sequence.

**Note on sort key range:** The SORT_KEY_BITS reduction from 23 to 22 further exacerbates the existing issue reported in #398 — sort keys that were valid before are now silently truncated at runtime (assert-only validation).

**To Reproduce**

For Bug 1:
1. Create a polygon geometry that spans tiles with `x >= 32768` at zoom 16
2. Run with `--bounds` or `--polygon` set
3. Observe that interior tiles in the eastern half are missing from output

For Bug 2:
1. Generate tiles at zoom 16 with tile IDs exceeding `2^32`
2. Check sort order of output features — high z16 tiles appear before z0

**Expected behavior**

- Bug 1: All z16 tiles should be present regardless of x coordinate
- Bug 2: Features should be sorted in correct tile order across all zoom levels

**Environment**

- Code review of commit `1fc95a26` (support zoom 16)
- Found via static analysis with multiple AI code review tools (Claude, GPT-5.4, Gemini)
- Not yet verified at runtime

smellman:
Prevent int overflow in TiledGeometry.encode and use unsigned long comparison for correct sort order.

# Summary

To fix #1541

~Bug 1 — TiledGeometry encode overflow (TiledGeometry.java):~
~- encode() returns int, which overflows for x >= 32768 at z16 (65536 * 32768 = 2^31)~
~- Fix: Integer.toUnsignedLong(encode(...)) prevents sign extension to negative long~

Bug 2 — Sort order with sign bit (multiple files):
  - SortableFeature.compareTo(): Changed Long.compare → Long.compareUnsigned
  - LongMerger.lessThan(): Changed ak < bk → Long.compareUnsigned(ak, bk) < 0, sentinel Long.MAX_VALUE → -1L (unsigned max)
  - ArrayLongMinHeap: Added configurable unsigned mode (boolean flag) so existing signed-comparison users aren't affected
  - LongMinHeap.newUnsignedArrayHeap(): New factory method used by KWayMerge

Regression tests added:
~- TiledGeometryTest.testPolygonZoom16HighXCoordinate — verifies Bug 1 fix~
  - FeatureGroupTest.testZ16HighTileIdsSortAfterLowerZooms — verifies Bug 2 fix

# co-author

Claude Code

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/ArrayLongMinHeap.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/ArrayLongMinHeap.java
@@ -45,12 +45,17 @@ class ArrayLongMinHeap implements LongMinHeap {
   protected final int max;
   protected int size;
   private final IntBinaryOperator tieBreaker;
+  private final boolean unsigned;
 
   /**
    * @param elements the number of elements that can be stored in this heap. Currently the heap cannot be resized or
    *                 shrunk/trimmed after initial creation. elements-1 is the maximum id that can be stored in this heap
    */
   ArrayLongMinHeap(int elements, IntBinaryOperator tieBreaker) {
+    this(elements, tieBreaker, false);
+  }
+
+  ArrayLongMinHeap(int elements, IntBinaryOperator tieBreaker, boolean unsigned) {
     // we use an offset of one to make the arithmetic a bit simpler/more efficient, the 0th elements are not used!
     posToId = new int[elements + 1];
     idToPos = new int[elements + 1];
@@ -59,6 +64,7 @@ class ArrayLongMinHeap implements LongMinHeap {
     posToValue[0] = Long.MIN_VALUE;
     this.max = elements;
     this.tieBreaker = tieBreaker;
+    this.unsigned = unsigned;
   }
 
   private static int firstChild(int index) {
@@ -163,10 +169,9 @@ class ArrayLongMinHeap implements LongMinHeap {
     }
     final int id = posToId[pos];
     final long val = posToValue[pos];
-    // the finish condition (index==0) is covered here automatically because we set vals[0]=-inf
     int parent;
     long parentValue;
-    while (compareIdPos(val, parentValue = posToValue[parent = parent(pos)], id, parent) < 0) {
+    while (pos > 1 && compareIdPos(val, parentValue = posToValue[parent = parent(pos)], id, parent) < 0) {
       posToValue[pos] = parentValue;
       idToPos[posToId[pos] = posToId[parent]] = pos;
       pos = parent;
@@ -226,21 +231,19 @@ class ArrayLongMinHeap implements LongMinHeap {
   }
 
   private int comparePosPos(long val1, long val2, int pos1, int pos2) {
-    if (val1 < val2) {
-      return -1;
-    } else if (val1 == val2 && val1 != Long.MIN_VALUE) {
-      return tieBreaker.applyAsInt(posToId[pos1], posToId[pos2]);
+    int cmp = unsigned ? Long.compareUnsigned(val1, val2) : Long.compare(val1, val2);
+    if (cmp != 0) {
+      return cmp;
     }
-    return 1;
+    return tieBreaker.applyAsInt(posToId[pos1], posToId[pos2]);
   }
 
   private int compareIdPos(long val1, long val2, int id1, int pos2) {
-    if (val1 < val2) {
-      return -1;
-    } else if (val1 == val2 && val1 != Long.MIN_VALUE) {
-      return tieBreaker.applyAsInt(id1, posToId[pos2]);
+    int cmp = unsigned ? Long.compareUnsigned(val1, val2) : Long.compare(val1, val2);
+    if (cmp != 0) {
+      return cmp;
     }
-    return 1;
+    return tieBreaker.applyAsInt(id1, posToId[pos2]);
   }
 
 }
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
@@ -14,6 +14,9 @@ public class LongMerger {
   // Has a general-purpose KWayMerge implementation using a min heap and specialized (faster)
   // TwoWayMerge/ThreeWayMerge implementations when a small number of lists are being merged.
 
+  /** Unsigned max: used as sentinel for exhausted iterators so they sort last in unsigned comparison. */
+  private static final long MAX_UNSIGNED_LONG = -1L;
+
   private LongMerger() {}
 
   /** Merges sorted items from {@link Supplier Suppliers} that return {@code null} when there are no items left. */
@@ -38,7 +41,7 @@ public class LongMerger {
 
     private final Comparator<T> tieBreaker;
     T a, b;
-    long ak = -1L, bk = -1L;
+    long ak = MAX_UNSIGNED_LONG, bk = MAX_UNSIGNED_LONG;
     final Iterator<T> inputA, inputB;
 
     TwoWayMerge(Iterator<T> inputA, Iterator<T> inputB, Comparator<T> tieBreaker) {
@@ -70,7 +73,7 @@ public class LongMerger {
           ak = a.key();
         } else {
           a = null;
-          ak = -1L;
+          ak = MAX_UNSIGNED_LONG;
         }
       } else if (b == null) {
         throw new NoSuchElementException();
@@ -81,7 +84,7 @@ public class LongMerger {
           bk = b.key();
         } else {
           b = null;
-          bk = -1L;
+          bk = MAX_UNSIGNED_LONG;
         }
       }
       return result;
@@ -93,7 +96,7 @@ public class LongMerger {
 
     private final Comparator<T> tieBreaker;
     T a, b, c;
-    long ak = -1L, bk = -1L, ck = -1L;
+    long ak = MAX_UNSIGNED_LONG, bk = MAX_UNSIGNED_LONG, ck = MAX_UNSIGNED_LONG;
     final Iterator<T> inputA, inputB, inputC;
 
     ThreeWayMerge(Iterator<T> inputA, Iterator<T> inputB, Iterator<T> inputC, Comparator<T> tieBreaker) {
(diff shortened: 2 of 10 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/render/TiledGeometry.java
```diff
@@ -299,7 +299,7 @@ public CoveredTiles getCoveredTiles() {
     }
     if (filledRanges != null) {
       for (var entry : filledRanges.entrySet()) {
-        long colStart = encode(maxTilesAtThisZoom, entry.getKey(), 0);
+        long colStart = Integer.toUnsignedLong(encode(maxTilesAtThisZoom, entry.getKey(), 0));
```
msbarry: Verify this is actually needed for the test to pass

smellman: @msbarry Thank you, you are right. I verified that RoaringBitmap.addOffset wraps correctly with signed/unsigned offsets (they are congruent mod 2^32), so Integer.toUnsignedLong is not needed here. Reverting.

msbarry: Thanks! If anything a comment might help so no one thinks it's a bug in the future.

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
```diff
@@ -438,11 +438,42 @@ void testEncodeLongKeyOrdering(
...
-      FeatureGroup.encodeKey(tileA, layerA, sortKeyA, hasGroupA) < FeatureGroup.encodeKey(tileB, layerB, sortKeyB,
-        hasGroupB)
+      Long.compareUnsigned(
+        FeatureGroup.encodeKey(tileA, layerA, sortKeyA, hasGroupA),
+        FeatureGroup.encodeKey(tileB, layerB, sortKeyB, hasGroupB)
+      ) < 0
     );
   }
 
+  @Test
```
msbarry, line 448: Thanks for updated this, can you also update MinHeapTest, LongMergerTest and add a PlanetilerTest case that verifies this end to end?

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
```diff
@@ -38,7 +38,7 @@ private static class TwoWayMerge<T extends HasLongSortKey> implements Iterator<T
 
     private final Comparator<T> tieBreaker;
     T a, b;
-    long ak = Long.MAX_VALUE, bk = Long.MAX_VALUE;
+    long ak = -1L, bk = -1L;
```
msbarry: Can you replace all -1L with reference to a constant MAX_UNSIGNED_LONG?

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/ArrayLongMinHeap.java
```diff
@@ -163,10 +169,9 @@ private void percolateUp(int pos) {
     }
     final int id = posToId[pos];
     final long val = posToValue[pos];
-    // the finish condition (index==0) is covered here automatically because we set vals[0]=-inf
     int parent;
     long parentValue;
-    while (compareIdPos(val, parentValue = posToValue[parent = parent(pos)], id, parent) < 0) {
+    while (pos > 1 && compareIdPos(val, parentValue = posToValue[parent = parent(pos)], id, parent) < 0) {
```
msbarry, line 174: why is this needed now?

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
smellman: I add tests for MinHeapTest, LongMergerTest, and PlanetilerTests.

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
smellman: I used MAX_UNSIGNED_LONG since the sentinel is a long value (-1L = 0xFFFFFFFFFFFFFFFFL). Did you mean MAX_UNSIGNED_LONG, or do you prefer a different name?

smellman:
@msbarry CI failed in download https://osmdata.openstreetmap.de/download/water-polygons-split-3857.zip , can you re-run this job?

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
```diff
@@ -63,16 +66,16 @@ public boolean hasNext() {
...
         if (inputA.hasNext()) {
           a = inputA.next();
           ak = a.key();
         } else {
           a = null;
-          ak = Long.MAX_VALUE;
+          ak = MAX_UNSIGNED_LONG;
         }
-      } else if (bk == Long.MAX_VALUE) {
+      } else if (b == null) {
```
msbarry, line 78: It seems like this should be `bk == MAX_UNSIGNED_LONG` in case the input interators contain null values (same below)

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/FeatureGroupTest.java
msbarry: Thanks!

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/MinHeapTest.java
```diff
@@ -59,6 +59,54 @@ void create(int capacity, IntBinaryOperator tieBreaker) {
     }
   }
 
+  static class UnsignedLongMinHeapTest {
```
msbarry: Can we make this extend MinHeapTest so that it runs all of the standard tests against unsigned min heap implementation?

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -1066,6 +1066,32 @@ void testZ16Fill() throws Exception {
...
+      (in, features) -> features.point("layer")
+        .setZoomRange(16, 16)
+        .inheritAttrFromSource("attr")
+    );
+
+    assertSubmap(Map.of(
+      TileCoord.ofXYZ(50_000, 500, 16), List.of(
+        feature(newPoint(128, 128), Map.of("attr", "value"))
+      )
+    ), results.tiles);
```
msbarry: This ensures that we output it correctly, but I'm wondering if there's a way to assert that the z16 tile appears in the output archive before a lower zoom tile?

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/ArrayLongMinHeap.java
smellman: The old compareIdPos had a special val1 != Long.MIN_VALUE guard that made the sentinel at position 0 always terminate the loop. The new comparison methods use plain Long.compare/Long.compareUnsigned without that special case, so the explicit pos > 1 bounds check is needed to replace it.

planetiler-core/src/main/java/com/onthegomap/planetiler/collection/LongMerger.java
smellman: b == null is intentional — MAX_UNSIGNED_LONG is a valid key value (when all bits of encodeKey are set), so bk == MAX_UNSIGNED_LONG would incorrectly throw NoSuchElementException for a real item. The test testMergeUnsignedKeyOrder covers this case with key=-1L. Null values from iterators are not a concern here since they would NPE at b.key() during advancement before reaching this check.

planetiler-core/src/test/java/com/onthegomap/planetiler/collection/MinHeapTest.java
smellman: I update test code.

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
smellman: Added a testZ16HighTileIteratesAfterLowerZoomTiles test in FeatureGroupTest that puts features at both z0 and z16 (x=50000) into a  FeatureGroup, then iterates to verify z0 appears before the z16 high tile. Also updated the PlanetilerTests case to output both z0 and z16 tiles and assert their relative order in the output.

msbarry (approved):

msbarry:
Looks great, thanks for the fixes @smellman !

Issue #1286, opened 2025-07-08
[FEATURE] Configurable tile extent per zoom in the output


mourner:
**Is your feature request related to a problem?**
Currently, if I understand the code correctly, the tile extent encoded in the output tiles is hardcoded to 4096:

https://github.com/onthegomap/planetiler/blob/edbe7f6bc01ef45dc7e2f7db71230e9fae09962a/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java#L92-L93

It does scale to higher precision for intermediary representations, but then gets scaled back to 4096.

My use case is pretty common — I want to generate a very small tileset, so I'm limiting maxzoom to a lower one (z11) to remove as much size overhead as possible, but this obviously leads to artifacts on tile boundaries because 4096 on max zoom tiles isn't enough precision when you zoom in on the map close enough.

**Describe the solution you'd like**
I'd like a way to configure tile extent so that e.g. I can leave it 4096 on most zooms, but bump it to 16384 on the maximum zoom level so that it packs as much precision as possible for overzooming.

**Describe alternatives you've considered**
An alternative is raising maxzoom, but that exponentially grows the overall size because of tile buffers & clipping overhead. Being able to increase tile precision at maxzoom for overzooming would be much preferable.

**Additional context**
Example of an overzooming artifact when maxzoom is low:
<img width="1308" height="313" alt="Image" src="https://github.com/user-attachments/assets/c3e7c0a8-dc9a-4c58-8c1f-7b185b0ab794" />

mourner:
Looks like the mismatch happens even if I set the tile buffer to 0, so the root cause is probably somewhere earlier, in the intermediary processing... So I'm wondering — should this code determine scale based on `maxzoomForRendering()` and not `maxzoom()`? It's not documented, but I assumed `maxzoom` is the max zoom of actual tiles in the output, and `render_maxzoom` is the maxzoom that influences tile precision.

https://github.com/onthegomap/planetiler/blob/edbe7f6bc01ef45dc7e2f7db71230e9fae09962a/planetiler-core/src/main/java/com/onthegomap/planetiler/render/FeatureRenderer.java#L262-L266

msbarry:
> It's not documented, but I assumed maxzoom is the max zoom of actual tiles in the output, and render_maxzoom is the maxzoom that influences tile precision.

[...]

PR #1565, opened 2026-05-14
Make tile extent runtime configurable with default 4096

tasiotas:
Hi

I added runtime arg for tile extent. Defaults to 4096, I tested it with 8192 and 16384. My use case required higher precision because I'm rendering maps with tileserver-gl at very high dpi for print.

Passed all tests. I used Codex 5.3

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
@@ -93,8 +93,9 @@ public class VectorTile {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(VectorTile.class);
 
-  // TODO make these configurable
-  private static final int EXTENT = 4096;
+  public static final int DEFAULT_EXTENT = 4096;
+  // Global vector tile extent configured at runtime through PlanetilerConfig.
+  private static volatile int EXTENT = DEFAULT_EXTENT;
   private static final double SIZE = 256d;
   // use a treemap to ensure that layers are encoded in a consistent order
   private final Map<String, Layer> layers = new TreeMap<>();
@@ -104,6 +105,17 @@ public class VectorTile {
     var encoder = new CommandEncoder(scale);
     encoder.accept(input);
     return encoder.result.toArray();
+  }
+
+  public static int extent() {
+    return EXTENT;
+  }
+
+  public static void setExtent(int extent) {
+    if (extent <= 0) {
+      throw new IllegalArgumentException("tile_extent must be > 0, got " + extent);
+    }
+    EXTENT = extent;
   }
 
   /**
@@ -368,7 +380,11 @@ public class VectorTile {
       List<Feature> features = new ArrayList<>();
       for (VectorTileProto.Tile.Layer layer : tile.getLayersList()) {
         String layerName = layer.getName();
-        assert layer.getExtent() == 4096;
+        if (layer.getExtent() != EXTENT) {
+          throw new IllegalStateException(
+            "Unsupported vector tile extent: " + layer.getExtent() + " (expected " + EXTENT + ")"
+          );
+        }
         List<String> keys = layer.getKeysList();
         List<Object> values = new ArrayList<>();
 
@@ -447,8 +463,8 @@ public class VectorTile {
    */
   public static int hilbertIndex(Geometry geometry) {
     Coordinate coord = geometry.getCoordinate();
-    int x = zigZagEncode((int) Math.round(coord.x * 4096 / 256));
-    int y = zigZagEncode((int) Math.round(coord.y * 4096 / 256));
+    int x = zigZagEncode((int) Math.round(coord.x * EXTENT / SIZE));
+    int y = zigZagEncode((int) Math.round(coord.y * EXTENT / SIZE));
     return (int) Hilbert.hilbertXYToIndex(15, x, y);
   }
 
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
@@ -96,6 +96,8 @@ public class VectorTile {
   public static final int DEFAULT_EXTENT = 4096;
   // Global vector tile extent configured at runtime through PlanetilerConfig.
   private static volatile int tileExtent = DEFAULT_EXTENT;
+  private static final int HILBERT_LEVEL = 16;
+  private static final long HILBERT_MAX_COORD = (1L << HILBERT_LEVEL) - 1;
   private static final double SIZE = 256d;
   // use a treemap to ensure that layers are encoded in a consistent order
   private final Map<String, Layer> layers = new TreeMap<>();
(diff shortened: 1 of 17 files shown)


planetiler-core/src/main/java/com/onthegomap/planetiler/config/PlanetilerConfig.java
```diff
@@ -90,6 +91,9 @@ public record PlanetilerConfig(
     if (maxzoom > MAX_MAXZOOM) {
       throw new IllegalArgumentException("Max zoom must be <= " + MAX_MAXZOOM + ", was " + maxzoom);
     }
+    if (tileExtent <= 0) {
+      throw new IllegalArgumentException("tile_extent must be > 0, was " + tileExtent);
+    }
```
msbarry: I think that extent needs to be a power of 2?

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
```diff
@@ -2088,7 +2088,7 @@ void testMergeLineStringsIgnoresRoundingIntersections() throws Exception {
         feature(newMultiLineString(
           newLineString(32, 64.3125, 37, 64.0625, 42, 64.3125),
           newLineString(32, 64, 37, 64.0625, 42, 64)
-        ), Map.of())
+        ), "layer", Map.of(), 0)
       )
     )), sortListValues(results.tiles));
   }
```
msbarry, line 2134: Can we add a test for end to end point/line/polygon for a few different values of tile extent? At least 8192 and 16384, possibly a lower value like 512 or 1024 as well? I think we should probably test a worst case for each of those with coordinates at (-255, -255) (513, -255) (513, 513) (-255, 513) and a few points in the middle

planetiler-core/src/main/java/com/onthegomap/planetiler/render/FeatureRenderer.java
```diff
@@ -264,9 +266,11 @@ private void writeTileFeatures(int zoom, long id, FeatureCollector.Feature featu
           // post-processing.  Features need to be "unscaled" in FeatureGroup after line merging,
           // and before emitting to the output archive.
           scale = Math.max(config.maxzoom(), 14) - zoom;
-          // need 14 bits to represent tile coordinates (4096 * 2 for buffer * 2 for zigzag encoding)
+          // need enough bits to represent tile coordinates (extent * 2 for buffer * 2 for zigzag encoding)
           // so cap the scale factor to avoid overflowing 32-bit integer space
-          scale = Math.min(31 - 14, scale);
+          long maxCoordinate = config.tileExtent() * 4L;
+          int bits = 64 - Long.numberOfLeadingZeros(maxCoordinate - 1);
+          scale = Math.clamp(scale, 0, Math.max(0, 31 - bits));
```
msbarry, line 273: There are a few more subtle places that the 4096 assumption has snuck in that might not explicitly reference 4096 - this is one of them, thanks for fixing! I'm trying to think if there might be any others...

planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java
```diff
@@ -447,8 +463,8 @@ public static VectorGeometryMerger newMerger(GeometryType geometryType) {
    */
   public static int hilbertIndex(Geometry geometry) {
     Coordinate coord = geometry.getCoordinate();
-    int x = zigZagEncode((int) Math.round(coord.x * 4096 / 256));
-    int y = zigZagEncode((int) Math.round(coord.y * 4096 / 256));
+    int x = zigZagEncode((int) Math.round(coord.x * tileExtent / SIZE));
+    int y = zigZagEncode((int) Math.round(coord.y * tileExtent / SIZE));
     return (int) Hilbert.hilbertXYToIndex(15, x, y);
```
msbarry: Zig-zag encoding maps coordinates in a range from -2^14 -> 2^14 to 0 -> 2^15. That assumption was based off 4096x4096 tiles with a 4096 buffer, so -4096 -> 8192 that gets mapped to 0 -> 16384 (2^14 plus 1 for safety).  I think we could probably bump 15 up to 16, but then if tile extent / size exceeds a certain threshold then we might need to shift x and y right by a certain amount to keep it in the right range.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
```diff
@@ -48,10 +49,24 @@
  */
 public class GeoUtils {
 
-  /** Rounding precision for 256x256px tiles encoded using 4096 values. */
-  public static final PrecisionModel TILE_PRECISION = new PrecisionModel(4096d / 256d);
+  /**
+   * Rounding precision for 256x256px tiles encoded using {@code tile_extent} values (default 4096).
+   */
+  private static final AtomicReference<PrecisionModel> tilePrecision =
+    new AtomicReference<>(new PrecisionModel(4096d / 256d));
```
msbarry, line 56: Ideally we move all references to 4096 out of static variables and either pass them as args to functions that need them or extract them from geoutils to a class that you instantiate with a tile extent.  That might make this PR very big though, let me know what you think - if it's too much I could do in a followup PR.

msbarry (commented):
Thanks for adding this! I'm a big nervous about other places that have a 4096 assumption built into them without explicitly referring to 4096. I think the end to end tests in PlanetilerTests will help flush out any of those.

planetiler-core/src/main/java/com/onthegomap/planetiler/geo/GeoUtils.java
msbarry: This issue also asks for the extent to be configurable per-zoom #1286 so we might even want to make it a global setting 🤔

tasiotas: how would you implement per zoom tile extent configuration? per zoom args?
--tileExtentZ12, --tileExtentZ13, etc... ?

planetiler-core/src/main/java/com/onthegomap/planetiler/config/PlanetilerConfig.java
tasiotas: yep, added power of 2 check

planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java
tasiotas: added

candux:
Thanks for this PR! Tested it and it works well to generate OSM indoor data at 16384 units. I think the pictures speak for themself :)
| v0.10.2 | This PR with 16384-unit extent |
| --- | --- |
| Planetiler v0.10.2 | This PR with 16384-unit extent |

tasiotas:
oh no, that last commit shouldnt be here, pls ignore if possible.

[ab69c4a](https://github.com/onthegomap/planetiler/pull/1565/commits/ab69c4ad682c4844ddedae7298b79de6f1d69249)

PR #1568, opened 2026-05-29
Add `--source_parallelism` flag to run multiple input sources concurrently

yuiseki:
## Summary

When a custom schema declares many small input sources whose per-feature work is light, `Planetiler.run` processes them sequentially and cannot saturate `process_threads`. On one such build, the pool runs at avg 2.6 of 31 threads per source (~8% CPU utilization).

This PR adds an opt-in `--source_parallelism=N` flag (default `1`). With `N>1`, up to `N` source stages run concurrently against a shared executor.

## Impact

Purely additive. Default preserves the existing sequential behavior bit-for-bit. Output `.pmtiles` is MD5-identical across `N=1`, `N=4`, and `N=8` on a real multi-source build I use locally.

## Performance

NVMe, 32-core host, JDK 21:

| `--source_parallelism` | wall-clock | speedup |
| --- | ---: | ---: |
| 1 (default) | 10m 43s | 1.00x |
| 4 | 4m 01s | 2.67x |
| 8 | 3m 52s | 2.77x |

HDD, same workload: `N=4` is 2.04x (14m 04s vs 6m 54s); `N=8` regresses to 9m 15s under disk contention.

## Notes

Per-stage thread CPU breakdown in `stats.json` gets mixed when stages overlap, since `Timers.currentStage` assumes LIFO nesting. Wall-clock per stage, output archive, and progress logs are unaffected. Happy to follow up with a fix in a separate PR if maintainers want it bundled.

Defaulting `N>1` and auto-tuning are out of scope here; open to either as follow-ups.

## AI assistance

Per [CONTRIBUTING.md#ai-assisted-contributions](https://github.com/onthegomap/planetiler/blob/main/CONTRIBUTING.md#ai-assisted-contributions): drafted with Claude Code. I reviewed every line, ran `spotless:check` and `planetiler-core` tests on JDK 21, and confirmed byte-identical output on a real build.

### DIFF
--- a/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
+++ b/planetiler-core/src/main/java/com/onthegomap/planetiler/Planetiler.java
@@ -1,5 +1,7 @@
 package com.onthegomap.planetiler;
 
+import static com.onthegomap.planetiler.worker.Worker.joinFutures;
+
 import com.onthegomap.planetiler.archive.TileArchiveConfig;
 import com.onthegomap.planetiler.archive.TileArchiveMetadata;
 import com.onthegomap.planetiler.archive.TileArchiveWriter;
@@ -39,6 +41,7 @@ import com.onthegomap.planetiler.util.Translations;
 import com.onthegomap.planetiler.util.Wikidata;
 import com.onthegomap.planetiler.validator.JavaProfileValidator;
 import com.onthegomap.planetiler.worker.RunnableThatThrows;
+import com.onthegomap.planetiler.worker.Worker.NamedThreadFactory;
 import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Files;
@@ -49,6 +52,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


msbarry:
Thanks for taking the time to make this change! Would it be possible to share full logs from the planetiler run before and after this change with your input data? In general planetiler tries to keep parallelism minimal so that num threads spreads works across all available cores instead of just loading it up with threads, so I want to see exactly what is limiting the parallelism before deviating from that goal.

yuiseki:
Thanks for the thoughtful question, and the point about keeping parallelism minimal makes sense to me. I put together a fully reproducible benchmark on open data so you can see exactly what limits the parallelism without needing my private input.

## What limits the parallelism

Each source stage is fed by a single shapefile reader thread (shapefiles are not splittable, so reading a source is inherently single-threaded). When a schema has many small sources whose per-feature work is light, that single reader cannot produce features fast enough to keep `process_threads` busy, so most of the pool sits idle for that stage. Because sources run one after another, the entire source-reading phase stays at low utilization no matter how high `process_threads` is.

In the run below, the 140 source stages average **3.5 active threads out of 31**, and the source-reading phase as a whole runs at about **4.7 threads (15% of 31 cores)**. The total CPU time is essentially identical across `N=1/4/8` (~12m), so this is not extra work, it is the same work spread over idle cores. `--source_parallelism` overlaps several single-threaded readers so the pool fills up.

## Reproducible open-data benchmark

7 Geofabrik free shapefile extracts, 20 layers each = 140 shapefile sources, 57,206,689 features, z11-14. 32-core host, JDK, NVMe, warm page cache.

Regions: ireland-and-northern-ireland, connecticut, iceland, new-hampshire, rhode-island, luxembourg, vermont (all from https://download.geofabrik.de). Geometries normalized with `ogr2ogr` so geotools accepts a handful of degenerate OSM polygons. Schema, download script, and the full before/after logs are in this gist: https://gist.github.com/yuiseki/8679a16d3dc946a13d13408687cec900

```
java -jar planetiler.jar generate-custom --schema=schema.yml --output=out.pmtiles [--source_parallelism=N]
```

| `--source_parallelism` | total wall | source-read phase | overall avg threads (of 31) | speedup |
| --- | ---: | ---: | ---: | ---: |
| 1 (default) | 1m45s | 94s | 6.9 | 1.00x |
| 4 | 34s | 23s | 21.4 | 3.06x |
| 8 | 42s | 31s | 18.1 | 2.49x |

`sort` (~4s) and `archive` (~6s) are unchanged between runs since they are already parallel, so the whole difference is in the source-reading phase. `N=4` is the sweet spot here and `N=8` regresses, which is why the flag is opt-in with a default of `1` rather than auto-tuned.

Output is byte-identical across all three (`md5 27c5acd626bcd30e90676c0838fe298e`, same 57,206,689 features), so nothing about the result changes, only how long it takes to produce.

## On the cosmetic caveat

As noted in the PR description, per-stage CPU numbers in the logs get mixed once stages overlap, because `Timers.currentStage` assumes LIFO nesting. The per-stage `cpu:`/`avg:` fields under `N>1` are therefore inflated and should be ignored; total wall, the overall summary line, output archive, and progress logs are all correct. Happy to fix that in a separate PR if you want it bundled.

Full before/after logs, schema, and the download script are all in the gist: https://gist.github.com/yuiseki/8679a16d3dc946a13d13408687cec900

PR #1570, opened 2026-06-07
fix(testLegacyWikidataNamesJson): close var in tes

tejas-ae:
Was reading through planetiler-core/src/test/java/com/onthegomap/planetiler/util/WikidataTest.java and the resource opened there can leak if testLegacyWikidataNamesJson exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Not sure if this is intentional, so feel free to ignore.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


msbarry:
Thanks!

Issue #1573, opened 2026-06-08
[FEATURE] Enable changing outlineFeatureTableNames for MLT


manuel-hegner:
**Is your feature request related to a problem? Please describe.**
For many usecases MLT pretessalation is very useful. Often enough generated outlines are required though. Right now there does not seem to be a way to change this.

**Describe the solution you'd like**
A dedicated option (and potentially other MLT settings) for this would be nice.

**Describe alternatives you've considered**
It would be even better if a Java Profile could modify the ConversionConfig directly.

PR #1575, opened 2026-06-08
Add support for MLT polygon outline

manuel-hegner:
This PR solves #1573 in a very simple way by just providing a toggle to turn the MLT polygon outline feature completely on or off.


msbarry:
Sorry for the delay getting back to you on this! This looks good overall, I just added a test to make sure we exercise this functionality in CI with no errors and will merge it in a few.

PR #1477, opened 2026-02-19
Expose label grid functions for custommap/YAML profiles

phanecak-maptiler:
This PR exposes "label grid" related functions for custommap/YAML profiles:

- `pointLabelGridPixelSize()`
- `pointLabelGridLimit()`
- `sortKey()`
- `sortKeyDescending()`

Those are needed for layers with many points for "label grid" point density limiting.

Usage example:

```
sources:
  osm:
    type: osm
    url: geofabrik:rhode-island
    local_path: data/rhode-island.osm.pbf
layers:
- id: testLayer
  buffer: 64
  features:
  - source: osm
    geometry: point
    point_label_grid:
      pixel_size:
        maxzoom: 8
        value: 32
      limit:
        maxzoom: 8
        value: 64
    sort_key: ${ feature.id }
```

PR #1576, opened 2026-06-14
Expose label grid functions for custommap/YAML profiles (2)

phanecak-maptiler:
This PR exposes "label grid" related functions for custommap/YAML profiles:

- `pointLabelGridPixelSize()`
- `pointLabelGridLimit()`
- `sortKey()`
- `sortKeyDescending()`

Those are needed for layers with many points for "label grid" point density limiting.

Usage example:

```
sources:
  osm:
    type: osm
    url: geofabrik:rhode-island
    local_path: data/rhode-island.osm.pbf
layers:
- id: testLayer
  buffer: 64
  features:
  - source: osm
    geometry: point
    point_label_grid:
      pixel_size:
        maxzoom: 8
        value: 32
      limit:
        maxzoom: 8
        value: 64
    sort_key: ${ feature.id }
```

Replacement for #1477


planetiler-custommap/src/main/java/com/onthegomap/planetiler/custommap/configschema/PointLabelGrid.java
```diff
@@ -0,0 +1,8 @@
+package com.onthegomap.planetiler.custommap.configschema;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record PointLabelGrid(
+  @JsonProperty("pixel_size") PointLabelGridPixelSize pixelSize,
+  @JsonProperty("limit") PointLabelGridLimit limit
+) {}
```
msbarry, line 8: I think eventually we want both of these to just be a shared "zoom function" expression that can be reused between them and elsewhere, and access the full API of [`ZoomFunction`](https://github.com/onthegomap/planetiler/blob/main/planetiler-core/src/main/java/com/onthegomap/planetiler/util/ZoomFunction.java#L4). This should be fine for an initial version as long as when we make the common zoom function expression, this "maxzoom/value" syntax is still valid in it.

